PR #6, opened 2018-01-27
removed mailroom directory and file

etwum:


natasha-aleksandrova:
Looks good! 
Note that python doc strings are usually enclosed in triple quotes `"""Function doc"""` or could use a comment line `# simple comment`

etwum:
thanks for the feedback. not really sure why i used the quotes instead of the hash symbol which i normally do for comments

natasha-aleksandrova:
Either will work of course just style tips 🙂

PR #12, opened 2018-01-27
added, then updated, break_me.py (lesson01)

johnpharmd:
simple Python functions that demonstrate, respectively, NameError, TypeError, SyntaxError, and AttributeError


students/johnpharmd/lesson01/break_me.py
```diff
@@ -0,0 +1,23 @@
...
+    b = '5'
+    print(5 + b)
+
+
+def funct3():
+	print('hello, world!)
+
+
+def funct4():
+    string = 'a, b, c, d'
```
natasha-aleksandrova: not important for this exercise but something to keep in mind

`string` is a built in module and need to be careful about overwriting built-in modules or reserved keywords. github is nice enough to syntax highlight them for you : ) (its in orange color)

here is example of what happens:
```
>>> import string
>>> string.digits
'0123456789'
>>> string = 'my string!'
>>> string.digits
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute 'digits'
>>>```

PR #25, opened 2018-01-29
Lesson 2

harvey-herela:
Solution for lesson 2

### DIFF
--- a/students/HarveyHerela/lesson02/part1.py
+++ b/students/HarveyHerela/lesson02/part1.py
@@ -1,0 +1,10 @@
+horizontalEdge = "+ - - - - + - - - - +"
+verticalEdge = "|         |         |"
+
+print(horizontalEdge)
+for x in range(0,3):
+    print(verticalEdge)
+print(horizontalEdge)
+for x in range(0,3):
+    print(verticalEdge)
+print(horizontalEdge)
--- a/students/HarveyHerela/lesson02/part2.py
+++ b/students/HarveyHerela/lesson02/part2.py
@@ -1,0 +1,59 @@
+import sys
+
+# Prints a specific row
+# Use 'edge' as the corner pieces and middle
+# and 'filler' as the filler pieces
+def print_row( edge, filler, side ):
+    # Special case, side <= 1
+    iter = 2
+    if side <= 1:
+        iter = 1
+    # print the left side, then the right side,
+    # with an edge in-between
+    print(edge, end = ' ')
+    for t in range(0, iter):
+        half = side // 2
+        for x in range(0,half):
+            print(filler, end=' ')
+        print(edge, end = ' ')
+    
+    print()
+
+# print_grid
+# parameters: 1 parameter for length of side
+def print_grid( side ):
+    # Print a top row
+    print_row('+','-',side)
+    
+    # Now print an open box, twice
+    half = side // 2
+    
+    # Special case for side <= 1
+    iter = 2
+    if side <= 1:
+        iter = 1
+    
+    for t in range(0, iter):
+        # First, print out the appropriate number
+        # of vertical pieces
+        for y in range(0,half):
+            print_row('|', ' ', side)
+        # Now print out the bottom row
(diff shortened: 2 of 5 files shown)


natasha-aleksandrova:
hello @harvey-herela !

please make sure to submit your github PR link to EdX so you can get a grade for each lesson.

students/HarveyHerela/lesson02/part1.py
```diff
@@ -0,0 +1,10 @@
+horizontalEdge = "+ - - - - + - - - - +"
```
robalford: Python naming convention is to use snake case (horizontal_edge) for all variable and function names

students/HarveyHerela/lesson02/part1.py
```diff
@@ -0,0 +1,10 @@
+horizontalEdge = "+ - - - - + - - - - +"
+verticalEdge = "|         |         |"
+
+print(horizontalEdge)
+for x in range(0,3):
```
robalford: Python/PEP8 style tip: add space between function args

students/HarveyHerela/lesson02/part2.py
```diff
@@ -0,0 +1,59 @@
+import sys
+
+# Prints a specific row
+# Use 'edge' as the corner pieces and middle
+# and 'filler' as the filler pieces
+def print_row( edge, filler, side ):
```
robalford: Python/PEP8 style for function declarations is one space between parameters, no space between parantheses and parameters, e.g. def print_row(edge, filler, side):

students/HarveyHerela/lesson02/part2.py
```diff
@@ -0,0 +1,59 @@
...
+    
+    for t in range(0, iter):
+        # First, print out the appropriate number
+        # of vertical pieces
+        for y in range(0,half):
+            print_row('|', ' ', side)
+        # Now print out the bottom row
+        print_row('+','-',side)
+
+#get the parameter
```
robalford: Add space between '#' and block comment

students/HarveyHerela/lesson02/part2.py
```diff
@@ -0,0 +1,59 @@
...
+if len(sys.argv) < 2:
+    print("Missing arguments!")
+    exit()
+
+try:
```
robalford: Nice use of try/except block and accepting command line arguments for the grid size is a nice touch!

students/HarveyHerela/lesson02/part2.py
```diff
@@ -0,0 +1,59 @@
...
+    num = int(sys.argv[1])
+    
+    if num < 1:
+        raise ValueError("Value should be greater than 0")
+    print_grid(num)
+
+except Exception as ex:
+    print("Try again with a positive integer")
+    
+print()
```
robalford: Add new line to end of file

students/HarveyHerela/lesson02/part2.py
```diff
@@ -0,0 +1,59 @@
...
+    exit()
+
+try:
+    num = int(sys.argv[1])
+    
+    if num < 1:
+        raise ValueError("Value should be greater than 0")
+    print_grid(num)
+
+except Exception as ex:
```
robalford: It's best to avoid catching the base Exception object when you can hone in on a particular error/exception, ValueError in this case. Don't need to cast Exception as ex here, since you're not referencing the exception object anywhere in the except block. This pattern is often used when you're logging the error, or printing the error itself to the console.

robalford:
Nice work overall! There are some PEP8 style revisions you could make to the third part of the assignment, similar to the ones I referenced in my comments on the first two. I would suggest running your code through a PEP8 linter because that will help you catch any white space errors or style guide violations. These kinds of errors aren't going to prevent your code from running but makes your code more Pythonic and readable to other Python devs. The Anaconda plugin for Sublime Text is a great free option for PEP8 linting as well as the free version of the PyCharm IDE.

PR #30, opened 2018-01-30
completed lesson 3 assignment

rmart300:
let me know if there are any issues with my interpretation of the needed methods

### DIFF
--- a/students/rmart300/lesson03/list_lab.py
+++ b/students/rmart300/lesson03/list_lab.py
@@ -38,5 +38,27 @@ for fruit in fruit_list:
     if fruit.startswith('P'):
         print(fruit)
 
+#series 2
+"""
+Using the list created in series 1 above:
+
+Display the list.
+Remove the last fruit from the list.
+Display the list.
+Ask the user for a fruit to delete, find it and delete it.
+(Bonus: Multiply the list times two. Keep asking until a match is found. Once found, delete all occurrences.)
+"""
+print(fruit_list)
+fruit_list.pop()
+print(fruit_list)
+fruit_to_remove = None
+while fruit_to_remove not in fruit_list:
+    fruit_to_remove = input("enter name of fruit to remove from list > ")
+
+fruit_list = fruit_list*2
+
+while fruit_to_remove in fruit_list:
+    fruit_list.remove(fruit_to_remove)
+print(fruit_list)
 
 
--- a/students/rmart300/lesson03/list_lab.py
+++ b/students/rmart300/lesson03/list_lab.py
@@ -83,13 +83,25 @@ def series3(fruit_list):
     print("Here's the fruit you like: ")
     print(fruit_list)
         
+def series4(fruit_list):
+    """
+    Once more, using the list from series 1:
+
+    Make a copy of the list and reverse the letters in each fruit in the copy.
+    Delete the last item of the original list. Display the original list and the copy.
+    """
+    fruit_list_copy = fruit_list[:]
+    fruit_list_copy.reverse()
+    fruit_list.pop()
+    print(fruit_list)
+    print(fruit_list_copy)
 
 if __name__ == '__main__':
     fruit_list = series1()
     #fruit_list_copy = fruit_list[:]
     #series2(fruit_list_copy)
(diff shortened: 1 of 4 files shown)


students/rmart300/slicing/slicing_methods.py
```diff
@@ -0,0 +1,47 @@
...
+    """exchange first and last elements of a sequence"""
+    return seq[-1:] + seq[1:-1] + seq[:1]
+    
+
+def remove_every_other(seq):
+    """remove every other item from sequence"""
+    return seq[::2]
+
+def remove_first_and_last_four(seq):
+    """remove the first four and the last four of the sequence, return the middle"""
```
natasha-aleksandrova: ... and `and then every other item in between`

your assert should be 
`assert remove_first_and_last_four(a_string) == " sas"`

students/rmart300/slicing/slicing_methods.py
```diff
@@ -0,0 +1,47 @@
...
+    return seq[4:-4]
+
+def reverse_seq(seq):
+    """reverse a sequence"""
+    return seq[::-1]
+
+def switch_thirds(seq=[]):
+    """divide sequence into 3, return second third then last third then first third"""
+    """if thirds not equal in size, the last one will contain the extra columns"""
```
natasha-aleksandrova: you can do
```
"""This is a really long 
multiline function doc
more here
"""
```

students/rmart300/slicing/slicing_methods.py
```diff
@@ -0,0 +1,47 @@
...
+    third = int(len(seq) / 3)
```
natasha-aleksandrova: not necessary to cast to `int` as it will already be an `int`

natasha-aleksandrova:
so you start with `this is a string` 
once you remove first 4 and last 4 you are left with
" is a st"
then you remove every other item (same logic as `remove_every_other`)

let me know if that helps

----
I can see how the wording on EdX can be mis-interpreted, I'll see about updating that to clarify.

rmart300:
ok I updated remove_first_and_last_four in slicing_methods.py, I tried to create a pull request but its not letting me submit it for this change or my change to lesson03 list assignment

natasha-aleksandrova:
@rmart300 all your commits end up part of this PR, while PR is open any commits to your source branch will end up as part of same PR. Typical flow usually involves multiple branches which we can hopefully start doing at some point.

You can always check Commits or Files Changed tabs in your PR and you should see all of your changes.

We will review shortly your updates.

students/rmart300/lesson03/list_lab.py
```diff
@@ -0,0 +1,107 @@
...
+    Again, using the list from series 1:
+
+    Ask the user for input displaying a line like “Do you like apples?” for each fruit in the list (making the fruit all lowercase).
+    For each “no”, delete that fruit from the list.
+    For any answer that is not “yes” or “no”, prompt the user to answer with one of those two values (a while loop is good here)
+    Display the list.
+    """
+    valid_answer = ('Yes','yes','y','Y','No','no','n','N')
+    for fruit in fruit_list:
+        answer = input("Do you like " + fruit.lower() + " > ")
```
natasha-aleksandrova: you can always do `answer.lower()` to avoid having upper/lower in `valid_answer` var

students/rmart300/lesson03/list_lab.py
```diff
@@ -0,0 +1,107 @@
...
+        
+def series4(fruit_list):
+    """
+    Once more, using the list from series 1:
+
+    Make a copy of the list and reverse the letters in each fruit in the copy.
+    Delete the last item of the original list. Display the original list and the copy.
+    """
+    fruit_list_copy = fruit_list[:]
+    fruit_list_copy.reverse()
```
natasha-aleksandrova: this is going to reverse the items in the list, not the letters in each fruit.

natasha-aleksandrova:
@rmart300 everything is looking really good, I'm going to merge it now.

take a look at `series4`  function and the requirements for it.

additionally.. (totally not required) you may want to checkout formatting options for README.md
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet

PR #31, opened 2018-01-30
Fixed series.py & list_lab.py

JerHsi:


natasha-aleksandrova:
@JerHsi thanks! looks much better

PR #33, opened 2018-01-30
added lesson01 folder

tricaonguyen:


robalford:
@tricaonguyen I'm going to make this PR as approved so you can go ahead and move on to the next assignment, but do make sure that you configure your text editor to translate tabs to spaces so that every time you hit the tab key, the single \t character is replaced by four \s characters. There are instructions in the course materials [here](https://uwpce-pythoncert.github.io/PythonCertDevel/supplemental/dev_environment/sublime_as_ide.html#basic-settings) for how to set this up for Sublime Text, but you can do this for whatever editor you're using.

PR #35, opened 2018-01-30
Lesson2 assignment

itsabhister:
Hello,
This is my first assignment and am new to Git. Hope I'm doing this correctly.

### DIFF
--- a/students/itsabhi/grid.py
+++ b/students/itsabhi/grid.py
@@ -1,0 +1,27 @@
+def printnewrow(size):
+    plus = "+"
+    minus = "-"
+    bar = "|"
+    i = size
+    while(i>0):
+        print (bar + " "*size + bar  + " "*size + bar)
+        i -= 1
+
+def draw_grid(x,y):
+    
+    plus = "+"
+    minus = "-"
+    bar = "|"
+    print (plus + minus*y + plus + minus*y + plus)
+    while (x>0):
+        x -= 1
+        printnewrow(y)
+        print (plus + minus*y + plus + minus*y + plus)
+
+
+def main():
+    draw_grid(2,4)
+
+if __name__ == "__main__":
+    main()
+    
--- a/students/itsabhi/README.rst
+++ b/students/itsabhi/README.rst
@@ -1,0 +1,1 @@
+Python code for UWPCE-PythonCert class, written by Abhijit Gupta


robalford:
@itsabhister looks like you've got the basic git workflow down for submitting assignments. 

It's a good idea to create a README file at the root of your repository per the instructions in the course materials [here](https://uwpce-pythoncert.github.io/PythonCertDevel/modules/GitWorkflow.html#adding-a-new-file). Also, please create a new directory in your repo for each lesson in the course and put the files for each of the assignments there.

You're off to a good start with the grid printer assignment but you should go through each of the three parts of the assignment in order, and make your code more generable and dynamic with each part.

For part 1, just try to create a simple 2x2 grid that looks exactly like this:

screen shot 2018-01-30 at 8 45 13 pm

Then for part 2, make it a function that can create a 2x2 grid of varying sizes based on a 'size' input parameter.

Part 3 is a bit trickier in that your function will take two input parameters, one to specify an equal number of grids and columns (3x3, 4x4 etc.) and the other for the 'size' of each cell.

Your current work on the assignment is closest to part 3, but I think there's value in going through each of these steps. Don't forget the spaces between the dashes and columns, and consider how you want to deal with those in your 'size' parameter.

I would suggest becoming familiar with the Python [PEP8](https://www.python.org/dev/peps/pep-0008/) style guide. Following these standards helps a lot with readability and is a best practice for production-ready Python. 

If you use Sublime Text with the Anaconda plugin, this includes a code linter that will show you PEP8 violations in your code. There is a description of how to set up Sublime Text for Python programming in the course materials [here](https://uwpce-pythoncert.github.io/PythonCertDevel/supplemental/dev_environment/sublime_as_ide.html#sublime-as-ide). Or, alternatively you can use the free version of the excellent Python IDE PyCharm which gives you PEP8 linting as well as lots of other powerful features.

I'll included a couple of other specific comments in your code to consider as you continue to work on the assignment.

students/itsabhi/grid.py
```diff
@@ -0,0 +1,27 @@
+def printnewrow(size):
+    plus = "+"
+    minus = "-"
```
robalford: The variables 'plus' and 'minus' are not used in the printnewrow function so you don't need to assign them here.

students/itsabhi/grid.py
```diff
@@ -0,0 +1,27 @@
+def printnewrow(size):
```
robalford: Python convention is to use 'snake case' such as print_new_row for naming functions and variables.

students/itsabhi/grid.py
```diff
@@ -0,0 +1,27 @@
...
+    i = size
+    while(i>0):
+        print (bar + " "*size + bar  + " "*size + bar)
+        i -= 1
+
+def draw_grid(x,y):
+    
+    plus = "+"
+    minus = "-"
+    bar = "|"
```
robalford: Variable 'bar' is not used in this function so should not be assigned here

itsabhister:
Thank you Natasha and Rob.

Rob, very valuable feedback. I have submitted a pull request with fixes.
Please let me know if this looks better.

PS. I could not paste the new PR into the "Grid Printer" assignment
section. Looks like this could be because it was marked as completed?

Thanks,
Abhi.

natasha-aleksandrova:
@itsabhister probably because it was already submitted. 

we will review your new PR even without submission just indicate as part of PR that this is update to <exercise_name>

PR #36, opened 2018-01-31
assignment: string_format.py

JerHsi:


students/JerryH/Lesson03/string_format.py
```diff
@@ -0,0 +1,52 @@
...
+    if t[0] < 10:
+        t0 = "file_00{}".format(str(t[0]))
+    elif t[0] < 100:
+        t0 = "file_0{}".format(str(t[0]))
+    else:
+        t0 = "file_{}".format(str(t[0]))
+
+    t1 = '{0:.2f}'.format(t[1])
+    t2 = '%.2E' % Decimal(t[2])
+    t3 = '%.2E' % Decimal(t[3])
```
natasha-aleksandrova: you shouldn't need the `Decimal` module here, without it also works:
```
>>> '%.2e' % 10000
'1.00e+04'
```

PR #42, opened 2018-02-01
more lesson02 work

johnpharmd:


students/johnpharmd/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,13 @@
+def fizzbuzz():
+    n = 1
+    while n < 101:
```
robalford: Check out Python's built-in range function for a more concise alternative to the while loop:

`for x in range(1, 101):`

students/johnpharmd/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,13 @@
...
+        if n % 3 == 0 and n % 5 == 0:
+            print('FizzBuzz')
+        elif n % 3 == 0:
+            print('Fizz')
+        elif n % 5 == 0:
+            print('Buzz')
+        else:
+            print(n)
+        n += 1
+
```
robalford: Looks good! Just for fun, can you think of a way to write this where you only evaluate for %3 and %5 once?

students/johnpharmd/lesson02/grid_draw.py
```diff
@@ -0,0 +1,9 @@
+def grid_draw():
+    n = 1
+    while n < 12:
+        if (n - 1) % 5 == 0:
+            print(2 * ('+ ' + 4 * ('- ')) + '+')
+        else:
+            print(2 * ('| ' + 8 * ' ') + '|')
+        n += 1
+
```
robalford: Remember to add new lines at the end of your files

robalford:
Fizz buzz looks good and grid printer is coming along nicely as well. Can you figure out a way to increase the size of each individual cell based on a function input parameter?

PR #43, opened 2018-02-01
Modified mailroom.py - changed list structures for donors & other comments

JerHsi:


students/JerryH/Lesson03/mailroom.py
```diff
@@ -1,63 +1,65 @@
...
-        if op == 1: #Append or list donors
+        if op == THANK_YOU_OPT: #Append or list donors
             #Add donor's name or list the exisiting donors
             d_name = thank_you_prompt()
             while d_name.lower() == "list": #provide a list of donors if "list" is entered
                 #Get each unique donor's name from the list
-                for each_name in (sorted(set(get_all_donor_names(donors)))):
+                for each_name in sorted(get_all_donor_names(donors)):
                     print("{}".format(each_name))
                 d_name = thank_you_prompt()
```
natasha-aleksandrova: one of the suggestions from the previous PR didn't save unfortunately, apologies.
again, it's optional, but I'd rather suggest and have you decide on improvements on your own

above could be written in this style:
```
donor_name = None
while not donor_name:
    donor_name = donor_name_prompt()  # more intuitive name ( was thank_you_prompt )
    if donor_name == "list":
        donor_name = None
        list_donor_names()
```
should be a little more readable and less code. Less code is not _always_ good, but often it is 🙂 

while loops are there to repeat the same blocks of code, with the help of condition variable (`donor_name`)
in the code you have this line is repeated `d_name = thank_you_prompt()` which often indicates that while loop logic can be improved. you want to minimize logic branches.

natasha-aleksandrova:
I think instead of me suggesting each pep8 violation it's better if you run the code through http://pep8online.com/

thanks!

natasha-aleksandrova:
going to merge for now. feel free to do another PR if you'd like.

PR #44, opened 2018-02-01
Lesson 1 merge

prgrover:


robalford:
Looks good! Note that unless you comment out the `test_syntax_error` function the code won't run at all.

PR #45, opened 2018-02-02
Lesson 02 - Fibonacci Exercise - Emmanuel Twum

etwum:


students/etwum/lesson02/series.py
```diff
@@ -0,0 +1,130 @@
...
+        # iterate the count variable
+        count += 1
+
+    # return the nth lucas integer
+    return print(lucas_list[-2])
+
+
+
+
+def sum_series(n ,y = 0, z = 1):
```
natasha-aleksandrova: pep8: extra spaces around keyword arguments, should be

`def sum_series(n, y=0, z=1):`

students/etwum/lesson02/series.py
```diff
@@ -0,0 +1,130 @@
...
+        lucas_list.append(x4)
+
+        # iterate the count variable
+        count += 1
+
+    # return the nth lucas integer
+    return print(lucas_list[-2])
+
+
+
```
natasha-aleksandrova: pep8: should only be 2 lines between functions

students/etwum/lesson02/series.py
```diff
@@ -0,0 +1,130 @@
...
+        series_list.append(x4)
+
+        # iterate the count variable
+        count += 1
+
+    # return the nth integer
+    return print(series_list[-2])
+
+
+while True:
```
natasha-aleksandrova: you probably don't need the while True here since you have `break` at the end

PR #49, opened 2018-02-02
More changes to mailroom.py

JerHsi:


students/JerryH/Lesson03/mailroom.py
```diff
@@ -1,79 +1,93 @@
...
-            #Add donor's name or list the exisiting donors
-            d_name = thank_you_prompt()
-            while d_name.lower() == "list": #provide a list of donors if "list" is entered
-                #Get each unique donor's name from the list
-                for each_name in sorted(get_all_donor_names(donors)):
-                    print("{}".format(each_name))
-                d_name = thank_you_prompt()
+def list_all_donor_names(donor_list):
+    for each_name in sorted(get_all_donor_names(donor_list)):
+        print("{}".format(each_name))
```
natasha-aleksandrova: from what I see here neither `list_all_donor_names` or `get_all_donor_names` need the `donor_list` param, since it's just referring to the global `donors` var

students/JerryH/Lesson03/mailroom.py
```diff
@@ -1,79 +1,93 @@
...
 
-            #for each_name in get_all_donor_names(donors):
+            # for each_name in get_all_donor_names(donors):
             for each_donor in donors:
-                total_donation = 0 #initialize total donation from the same donor
-                #Get total donation from each donor
+                total_donation = 0  # initialize total donation from the same donor
+                # Get total donation from each donor
                 for i in range(len(each_donor[1])):
                     total_donation += each_donor[1][i]
```
natasha-aleksandrova: so here you can do:
`total_donation = sum(each_donor[1])`

https://docs.python.org/2/library/functions.html#sum

students/JerryH/Lesson03/mailroom.py
```diff
@@ -1,79 +1,93 @@
...
-                donors.append([d_name, [donation]])
+            # donors.append([donor_name, donation])
+            if donor_name not in get_all_donor_names(donors):
+                # print(donors)
+                donors.append([donor_name, [donation]])
             else:
-                #print(get_all_donor_names(donors).index(d_name))
-                donors[get_all_donor_names(donors).index(d_name)][1].append(donation)
+                # print(get_all_donor_names(donors).index(donor_name))
+                donors[get_all_donor_names(donors).index(donor_name)][1].append(donation)
```
natasha-aleksandrova: you could have a function called `find_donor_by_name`
which returns an item in the `donors` data structure (our database essentially)
```
find_donor_by_name(name):
    for d in donors:
        if d[0] == name:
            return d
```
so here you could simply do
```
existing_donor = find_donor_by_name(donor_name)
if existing_donor:
    existing_donor[1].append(donation)
else:
    donors.append([donor_name, [donation]])
```

PR #57, opened 2018-02-03
Dustin_L - Lesson03 - List Lab

rdlaner:
List Lab assignment in Lesson03.

### DIFF
--- a/students/Dustin_L/lesson03/list_lab.py
+++ b/students/Dustin_L/lesson03/list_lab.py
@@ -89,8 +89,37 @@ def series_two(lst):
     return l
 
 
+def series_three(lst):
+    """Performs a set of operations on a list of fruit.
+
+    A deep copy is made of the passed list. The user is then prompted to specify
+    which fruits in the list they like. For all that they do not like, those
+    fruits are removed from the list.
+
+    Args:
+        lst (list): List to be operated upon.
+
+    Returns:
+        list: Modified list
+    """
+    l = copy.deepcopy(lst)
+
+    for fruit in lst:
+        usr_in = input(f'Do you like {fruit.lower()} (Yes / No)?' + PROMPT).lower()
+
+        while (usr_in[0] != 'y') and (usr_in[0] != 'n'):
+            usr_in = input('Please enter "Yes" or "No"' + PROMPT).lower()
+
+        if usr_in.startswith('n'):
+            l.remove(fruit)
+
+    print(l)
+    return l
+
+
 if __name__ == '__main__':
     fruit_list = ['Apples', 'Pears', 'Oranges', 'Peaches']
 
     series_one(fruit_list)
     series_two(fruit_list)
+    series_three(fruit_list)
--- a/students/Dustin_L/lesson03/list_lab.py
+++ b/students/Dustin_L/lesson03/list_lab.py
@@ -117,9 +117,35 @@ def series_three(lst):
     return l
 
 
+def series_four(lst):
+    """Perform a set of operations on a list of fruit.
+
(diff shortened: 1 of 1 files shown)


students/Dustin_L/lesson03/list_lab.py
```diff
@@ -0,0 +1,151 @@
...
+    l += ['Grape']
+    print(l)
+
+    # Add fruit using 'insert()' function
+    l.insert(0, 'Mango')
+    print(l)
+
+    # Display all fruit that start with 'P'
+    for fruit in l:
+        if fruit.startswith('P'):
```
robalford: You might want to do something like `if fruit[0].lower() == 'p'` here just in case a user inputs lower case 'papayas'  e.g. in the first prompt

students/Dustin_L/lesson03/list_lab.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # Remove last item
+    l.pop()
+    print(l)
+
+    # Prompt user for fruit and then delete all occurences from list
+    l *= 2
+    while True:
+        usr_in      = input('Enter fruit to remove from list' + PROMPT)
+        removed_lst = [i for i in l if i.lower() != usr_in.lower()]
```
robalford: Nice use of a list comprehension to build a filtered copy of the list. You've been reading ahead :)

students/Dustin_L/lesson03/list_lab.py
```diff
@@ -0,0 +1,151 @@
...
+
+        if removed_lst != l:
+            l = removed_lst
```
robalford: You don't necessarily need to reassign `l` here, you could just print and return `removed_list` at this point

students/Dustin_L/lesson03/list_lab.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Args:
+        lst (list): List to be operated upon.
+
+    Returns:
+        list: Modified list
+    """
+    l = copy.deepcopy(lst)
+
+    for fruit in lst:
```
robalford: An alternative to using `deepcopy` would be to use the slicing operator to loop through a copy of the list like this, so you keep all the original indexes as you loop:

`for fruit in lst[:]:`

students/Dustin_L/lesson03/list_lab.py
rdlaner: I see. I'm used to programming in C, where removed_list would be scoped to the While statement and not accessible outside of it (since it is declared in that scope). Is that not the case with Python? What exactly are the scoping rules for Python?
Thanks!

students/Dustin_L/lesson03/list_lab.py
rdlaner: Ah, nice one, thanks.

students/Dustin_L/lesson03/list_lab.py
robalford: Yeah that makes sense. I learned programming and continue to mostly work in Python (and a little JavaScript). The subtle differences between languages are interesting. 

In Python, a function body is the inner most possible scope, so variables defined on the final iteration of a loop remain accessible within the enclosing function body. 

Here are a couple of pretty in-depth articles about scoping in Python that I just found online if you want to get into the nitty-gritty:

http://sebastianraschka.com/Articles/2014_python_scope_and_namespaces.html

https://eli.thegreenplace.net/2015/the-scope-of-index-variables-in-pythons-for-loops/

rdlaner: That's great info, thanks!

PR #58, opened 2018-02-04
print exercise

ZhuojingXie:
I have so many trouble to using loop and using global/local variables. even though I made it eventually.

Should I start from bottom or should I start from Top? What I am trying to say, in this case, is should I focus on how to print out the first line or I should think about what I need to complete  print_grid2(m,n) function at very beginning?

### DIFF
--- a/students/ZhuojingXie/lesson02/fizzbuzz.py
+++ b/students/ZhuojingXie/lesson02/fizzbuzz.py
@@ -1,0 +1,9 @@
+for i in range(1,101):
+    if i %3 == 0 and i% 5 == 0:
+        print('fizzBuzz')
+    elif i%3 == 0:
+        print('fizz')
+    elif i%5 == 0:
+        print('Buzz')
+    else:
+        print(i)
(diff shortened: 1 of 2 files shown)


ZhuojingXie:
fizzbuzz is much more easier than print grid

natasha-aleksandrova:
@ZhuojingXie don't stress about print grid 🙂 

it can be a tough one to implement as there is so much going on; in further exercise you will see more practical applications of `for` and `while` loops.

natasha-aleksandrova:
@ZhuojingXie I'm going to merge this, but please be sure to submit in EdX to get the grade (something you should always do right after you make the PR).

ZhuojingXie:
Understand. But could you please more specific about how to submit in Edx? is that the upload block under every exercise? btw, what PR stand for?

natasha-aleksandrova:
PR stands for Pull Request

There is "Assignment Submission Area" under each assignment and you should something like this:
screen shot 2018-02-04 at 12 17 50 pm

robalford:
@ZhuojingXie looks like you've included the link to your student directory rather than pull request when you submitted the assignment on the course website. I see how this would be confusing because it looks like your pull request was already merged before you submitted the assignment. I will grade your assignment as 'Good' so you can move on, and next time when you open your pull request for the assignment paste the link to the pull request itself into the field on the course website as @natasha-aleksandrova describes in her comment here.

PR #59, opened 2018-02-04
Andrew Park Mailroom

nDruP:
I finished mailroom.py

### DIFF
--- a/students/nDruP/lesson03/mailroom.py
+++ b/students/nDruP/lesson03/mailroom.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env python3
+"""
+1. Create Data structure that holds Donor, Donation Amount.
+2. Prompt user to Send a Thank You, Create a Report, or quit.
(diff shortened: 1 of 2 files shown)


students/nDruP/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+    Then prompt for donation amount, add amount to donation history
+    Compose and print an email thanking the donor for their donation.
+    Return to main
+    """
+    donor_name = "list"
+    donation_amt = None
+    print("Let's craft a very personal thank you note for our donor!")
+    print("Return to the main menu at any time by entering 'exit'")
+    print("Pull up a list of donor names by entering 'list'")
+    while donor_name == "list":
```
robalford: Another way to construct this loop would be to use `while True`, which will continue the loop until a `break` statement or `return` statement that exits the function. Then you wouldn't need to assign the `donor_name` and `donation_amount` variables before the loops.

PR #60, opened 2018-02-04
The rest of assignment 2: fizzbuzz and series

harvey-herela:


students/HarveyHerela/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,16 @@
...
+        buzz = True
+    
+    if fizz or buzz:
+        if fizz:
+            print("Fizz", end = '')
+        if buzz:
+            print("Buzz", end = '')
+        print()
+    else:
+        print(i)
```
natasha-aleksandrova: this certainly works! but can you think of another way that is more inline and less code?

harvey-herela: Yes. I could have just 1 Boolean instead of 2, and it would only keep track if fizz/buzz/fizzbuzz was printed instead of determining what to print. Should I update my code and do another pull request?

natasha-aleksandrova: You can do it with no booleans at all 😉 

not required, but if you have time I would say give it another try (yes, it is truly given at interviews)!

PR #68, opened 2018-02-05
Lesson 02

MikeS-1:


students/MikeShand/lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,10 @@
+def fizz(n):
+    for i in range(n):
+        if i%3==0 and i%5!=0:
+            print('fizz')
+        elif i%5==0 and i%3!=0:
+            print('buzz')
+        elif i%3==0 and i%5==0:
```
natasha-aleksandrova: if you do this check as part of the first if statement then you wouldn't need additional checks like !=

PR #69, opened 2018-02-05
Adding mailroom2.py

JerHsi:
@natasha-aleksandrova Could you check if I am on the right track so far? Just want to check it with you before I start writing for the second part of the assignment. Thanks

### DIFF
--- a/students/JerryH/Lesson04/mailroom2.py
+++ b/students/JerryH/Lesson04/mailroom2.py
@@ -1,21 +1,15 @@
 #!/usr/bin/env python3
+from collections import defaultdict
 
 # List of sample donors and donations
-donors = {
+donors = defaultdict(list, {
     "Bill Gates": [234.22, 45645.24, 43953.09, 98823],
     'Jeff Bezo': [4564.23],
     'Mike Dell': [299.09, 26273.67],
     'Harry Potter': [8234.09, 83948.04, 7834.23],
     'Ben Williams': [29283.00, 1334.34],
     'Guy James': [93.00, 34.34]
-    }
-
-options = {
-    'THANK_YOU_OPT': 1,
-    'REPORT_OPT': 2,
-    'SEND_LETTER_OPT': 3,
-    'QUIT_OPT': 4
-    }
+    })
 
 menu = {
     'op1': "Send a Thank You",
@@ -24,9 +18,10 @@ menu = {
     'op4': "Quit"
     }
 
+QUIT_OPT = '4'
 
 def prompt():
-    return int(input("Please choose the following options:\n1) {op1}.\n2) {op2}.\n3) {op3}.\n4) {op4}\n".format(**menu)))
+    return input("Please choose the following options:\n1) {op1}.\n2) {op2}.\n3) {op3}.\n4) {op4}\n".format(**menu))
 
 
 def donor_name_prompt():
@@ -35,10 +30,7 @@ def donor_name_prompt():
 
 
 def get_all_donor_names():
-    all_names = []
-    for d in donors.keys():
-        all_names.append(d)
-    return all_names
+    return list(donors)
 
 
 def list_all_donor_names():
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
@JerHsi certainly, I will try tonight.. in the worst case early tomorrow morning

students/JerryH/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,102 @@
...
+    'Ben Williams': [29283.00, 1334.34],
+    'Guy James': [93.00, 34.34]
+    }
+
+options = {
+    'THANK_YOU_OPT': 1,
+    'REPORT_OPT': 2,
+    'SEND_LETTER_OPT': 3,
+    'QUIT_OPT': 4
+    }
```
natasha-aleksandrova: as part of the instructions we have: "See if you can use a dict to switch between the users selections. "

it looks like you tried to incorporate a dict into your options.. but you can take it a step further and do something like:

```
selection_map = {
    "1": send_thank_you,
    "2": create_report,
    "3": send_letters,
    "4": quit_program
}
```
the dict values are function objects, and it will act like a switch based on user input for the option.
you use will use it like this:
```
option_value = input("..")
selection_map[optiona_value]()
```

you can also use `selection_map` for validation:
```
if option_value not in selection_map.values() :
     # continue looping asking for user input
```

above pattern will also encourage you to break up code into more functional blocks (by using functions)

students/JerryH/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,102 @@
...
+def donor_name_prompt():
+    return input("Send a Thank You - Please enter a full name or type \"list\""
+        "to list the current donors:\n")
+
+
+def get_all_donor_names():
+    all_names = []
+    for d in donors.keys():
+        all_names.append(d)
+    return all_names
```
natasha-aleksandrova: dict is a great choice for `donors` data structure.. what's even better is it is so easy now to get a list of donor names 🙂 
```
>>> list(donors)
['Bill Gates', 'Jeff Bezo', 'Mike Dell', 'Harry Potter', 'Ben Williams', 'Guy James']
>>> list(donors.keys())
['Bill Gates', 'Jeff Bezo', 'Mike Dell', 'Harry Potter', 'Ben Williams', 'Guy James']
>>> donors.keys()
dict_keys(['Bill Gates', 'Jeff Bezo', 'Mike Dell', 'Harry Potter', 'Ben Williams', 'Guy James'])
```
so your function could simply be:
```
def get_all_donor_names():
    return list(donors)
```

students/JerryH/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,102 @@
...
+                    list_all_donor_names()
+            # print(donor_name)
+            donation = float(input("Please enter the donation amount:\n"))
+            # If the donnor doesn't exist in the donor dictionary - add his info
+            if donor_name not in get_all_donor_names():
+                # print(donors)
+                donors[donor_name] = [donation]
+            # else append the donation into the existing donors
+            else:
+                donors[donor_name].append(donation)
```
natasha-aleksandrova: I'm getting a little fancy here 😃 

there is an "extension" of a `dict` object called `defaultdict` that I'm really fond of, it makes some things much easier and cleaner.
The way it works is at the time you create your dict you specify default value, so in your case instead of having to check if donor name exists so you can safely use `append` on the existing donor, you just define default value as list and go ahead and use `append` regardless.

you can read the doc here:
https://docs.python.org/2/library/collections.html#collections.defaultdict

and see my example below:

```
>>> donors = defaultdict(list, {
...     "Bill Gates": [234.22, 45645.24, 43953.09, 98823],
...     'Jeff Bezo': [4564.23],
...     'Mike Dell': [299.09, 26273.67],
...     'Harry Potter': [8234.09, 83948.04, 7834.23],
...     'Ben Williams': [29283.00, 1334.34],
...     'Guy James': [93.00, 34.34]
...     })
>>>
>>> donors["James Bond"].append(15000)
>>> donors
defaultdict(<class 'list'>, {'Bill Gates': [234.22, 45645.24, 43953.09, 98823], 'Jeff Bezo': [4564.23], 'Mike Dell': [299.09, 26273.67], 'Harry Potter': [8234.09, 83948.04, 7834.23], 'Ben Williams': [29283.0, 1334.34], 'Guy James': [93.0, 34.34], 'James Bond': [15000]})
```

JerHsi:
@natasha-aleksandrova I made some more changes based on your comments. Please take a look.  Thanks.

students/JerryH/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,113 @@
...
+selection_map = {
+    "1": send_thank_you,
+    "2": create_report,
+    "3": send_letters,
+    "4": quit_program
+	}
+
+
+def main():
+	option_value = 0
```
natasha-aleksandrova: can you take a look at the indentation here and couple other places? it is more than 4 spaces

students/JerryH/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,113 @@
...
+	while option_value != QUIT_OPT:
+		option_value = prompt()
+		selection_map[option_value]()
```
natasha-aleksandrova: be careful you might get an exception error if the user enters anything other than 1-4. 

you could catch KeyError here:
```
try:
    selection_map[option_value]()
except KeyError:
    print("%s is not a valid option! Please try again." % option_value)

```

students/JerryH/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,113 @@
...
+
+    donation = float(input("Please enter the donation amount:\n"))
+
+    # If the donnor doesn't exist in the donor dictionary - add his info
+    if donor_name not in get_all_donor_names():
+        # print(donors)
+        donors[donor_name] = [donation]
+    # else append the donation into the existing donors
+    else:
+        donors[donor_name].append(donation)
```
natasha-aleksandrova: plz review my previous comment on this area.

what you have here is correct and will work, but we can make it better 🙂 

since you are using defaultdict now, you do not need to check if donor exists or not, you can just call append on the new donor name. it makes things much simpler.

I am happy to explain this more extensively if you would like me to, let me know.

students/JerryH/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,113 @@
...
+	        r_sorted[d][2],
+	        r_sorted[d][1]/r_sorted[d][2]))
+	print("\n")
+
+
+def send_letters():
+    #print("This is option3.")
+    for name in donors:
+        file_name = name.lower().replace(' ', '_', 3) + '.txt'
+        f = open(file_name, 'w')
```
natasha-aleksandrova: highly recommend getting used to using `with` statement when opening files. you do not need to know how it works yet internally (unless you do, and I can share some materials)

here is little more about it here:
https://docs.quantifiedcode.com/python-anti-patterns/maintainability/not_using_with_to_open_files.html
and from course docs:
https://uwpce-pythoncert.github.io/PythonCertDevel/modules/Files.html

students/JerryH/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,113 @@
...
+
+	# for each_name in get_all_donor_names(donors):
+	for each_donor in donors.keys():
+	    report.append([each_donor, sum(donors[each_donor]), len(donors[each_donor])])
+
+	# sorting the report based on donations
+	r_sorted = sorted(report, key=lambda r: r[1], reverse=True)
+
+	# Create the report
+	for d in range(len(r_sorted)):
```
natasha-aleksandrova: since you're only using `d` to access list by index, you can just loop through items directly

```
# Create the report
for donor_report in r_sorted:
    # print(d)
    print("{:23}${:12.2f}{:10}   ${:12.2f}".format(donor_report[0], donor_report[1], donor_report[2],
                                                   donor_report[1] / donor_report[2]))

```

PR #72, opened 2018-02-05
Grid Printer Excercise, first commit

JeroenSweerts:
Please review my code for the Grid Printer Exercise. The code is built up in 3 parts:
1. A function with no arguments
2. A function with 1 argument
3. A function with 2 arguments

The code has been tested and seemed to behave like expected.

Thanks.

Jeroen

### DIFF
--- a/students/JeroenSweerts/Lesson2/FizzBuzz.py
+++ b/students/JeroenSweerts/Lesson2/FizzBuzz.py
@@ -1,0 +1,9 @@
+for i in range(1,101):
+    if (i)%3 == 0 and (i)%5 == 0:
+        print('FizzBuzz')
+    elif (i)%3 == 0:
+        print('Fizz')
+    elif (i)%5 == 0:
+        print('Buzz')
+    else:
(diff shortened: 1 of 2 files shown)


students/JeroenSweerts/Lesson2/Grid_Printer.py
```diff
@@ -0,0 +1,60 @@
+################################PART1###################################
```
natasha-aleksandrova: more pythonic to lower case your files: `grid_printer.py`

students/JeroenSweerts/Lesson2/FizzBuzz.py
```diff
@@ -0,0 +1,9 @@
+for i in range(1,101):
+    if (i)%3 == 0 and (i)%5 == 0:
```
natasha-aleksandrova: you don't need to put parens around the `i` var
should just be:
```
if i * 3 == 0 or i  % 5 == 0:
```

PR #73, opened 2018-02-05
Lesson 2 assignment: Fizz Buzz Excercise

itsabhister:


students/itsabhi/Lesson2/fizzbuzz.py
```diff
@@ -0,0 +1,24 @@
...
+                print ("Fizz")
+                i += 1
+                continue
+        if (i%5 == 0):
+            print ("Buzz")
+            i += 1
+            continue
+        else:
+            print (i)
+            i += 1
```
natasha-aleksandrova: see if you can make some adjustments to your solution by utilizing `range` function and by reducing/consolidating if / else statements

what's great about range in this case is it provides the counter for you, so no need to define your own counter (`i`)

example below is the most basic usage of `range`:
```
>>> for i in range(5):
...     print(i)
...
0
1
2
3
4
```

you can also specify your start and end
```
>>> max_size = 5
>>> for i in range(1, max_size + 1):
...     print(i)
...
1
2
3
4
5
```

itsabhister:
Thanks Natasha!

Yes, the range makes it a lot cleaner. Please review.

https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/77.

Thanks,
Abhi.

PR #77, opened 2018-02-06
Updated using range function

itsabhister:
Optimized w/ range and fewer if iterations.


natasha-aleksandrova:
looks much better!

PR #85, opened 2018-02-07
Herela-List lab

harvey-herela:


students/HarveyHerela/lesson03/list_lab.py
```diff
@@ -0,0 +1,103 @@
...
+
+# Start of series 2
+print("Start of series 2")
+# Display the list.
+print(my_fruits)
+
+# Remove the last fruit from the list.
+# Display the list.
+print("Removing the last fruit")
+my_fruits.pop(len(my_fruits)-1)
```
robalford: If you call `pop()` on a list without passing an index, it will remove and return the last item in the list so this can just be `my_fruits.pop()`.

students/HarveyHerela/lesson03/list_lab.py
```diff
@@ -0,0 +1,103 @@
...
+print(my_fruits)
+
+# Ask the user for a fruit to delete, find it and delete it.
+# (Bonus: Multiply the list times two. Keep asking until a match is found. Once found, delete all occurrences.)
+my_fruits = my_fruits * 2
+num = 0
+while num == 0:
```
robalford: See if you can use the `in` keyword here to make this more concise e.g. `while response in my_fruits:` ...

students/HarveyHerela/lesson03/list_lab.py
```diff
@@ -0,0 +1,103 @@
...
+print(my_fruits)
+
+# Ask the user for input displaying a line like "Do you like apples?" for each fruit in the list (making the fruit all lowercase).
+# For each "no", delete that fruit from the list.
+# For any answer that is not "yes" or "no", prompt the user to answer with one of those two values (a while loop is good here)
+
+# I don't want to invalidate my iterator, so make a copy of the list
+# to use for looping
+my_fruits_loop = list(my_fruits)
+for fruit in my_fruits_loop:
```
robalford: Another way to loop through a copy would be to use the slice operator:

`for fruit in my_fruits[:]`

natasha-aleksandrova:
@harvey-herela for future assignment - if you could, would be more readable if each task/series was in a separate function; for readability and to keep things organized better. thanks!

PR #86, opened 2018-02-07
fizz_buzz assignment

rmart300:


students/rmart300/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,14 @@
+
+
+for i in range(100):
```
natasha-aleksandrova: you can specify start and end for your range like:
`for i in range(1, 101)`

this way you don't need to `i + 1`

PR #87, opened 2018-02-07
grid printer assignment

alexLaws:
Here's the Grid Printer Assignment


robalford:
Looks good. Nice work breaking up the problem into smaller tasks and reusable functions. I would recommend running the code through an online linter to catch a few Python-specific style violations: http://pep8online.com/. Or you can probably configure your text editor to catch pep8 (Python style guide) violations for you.

natasha-aleksandrova:
Great suggestion by Rob to run the code through linter.

Since I know you use Sublime, here is decent guide to getting that installed in sublime
http://cheng.logdown.com/posts/2015/06/14/sublime3-install-python-linters

specifically parts 3 and 6

PR #88, opened 2018-02-07
commiting the fibonacci and lucas series code.

JeroenSweerts:
Please review the code of the series.py file for Fibonacci and lucas series.

Thanks.

Jeroen

### DIFF
--- a/students/JeroenSweerts/Lesson2/series.py
+++ b/students/JeroenSweerts/Lesson2/series.py
@@ -1,0 +1,66 @@
+######################STEP1#########################################
+def fibonacci(n):
+    """The first line (n = n - 1) ensures that we get the nth number in the series.
+    For example fibonacci(1) should return 0. Without this line it would return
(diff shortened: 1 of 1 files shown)


students/JeroenSweerts/Lesson2/series.py
```diff
@@ -0,0 +1,66 @@
...
+            a, b = b, a + b
+            n -= 1
+        return a
+
+print("sum_series(8,2,1): " ,sum_series(8,2,1))
+print("sum_series(8,0,1): " ,sum_series(8,0,1))
+######################Generalizing#########################################
+
+print("testing fibonacci:")
+assert fibonacci(8) == 13, "fibonacci function generates wrong output"
```
natasha-aleksandrova: this should actually be 9 == 34, you want to get nth value starting with zero index

(assignment wasn't super clear on that, I have updated it in EdX)

PR #92, opened 2018-02-08
Trigrams

JerHsi:
@natasha-aleksandrova Not sure if I am doing this right.

### DIFF
--- a/students/JerryH/Lesson04/trigrams.py
+++ b/students/JerryH/Lesson04/trigrams.py
@@ -6,10 +6,9 @@ from collections import defaultdict
 
 
 def get_text():
-    file = open("sherlock.txt", "r")
-    text = file.read()
-    file.close()
-    return text
+    with open("sherlock_small.txt", "r") as fh: # fh: file handle
+        return fh.read()
+
 
 
 def get_word_list():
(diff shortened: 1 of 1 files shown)


students/JerryH/Lesson04/trigrams.py
```diff
@@ -0,0 +1,58 @@
...
+    # Stops the loop if there's no more next word after the chosen word pair
+    while next_word != None:
+        next_word = word_dict[new_wordings[-2] + ' ' + new_wordings[-1]] # Find the next word
+
+        if next_word != []:
+            new_wordings.append(next_word[0])
+            # Delete that word from the list so it doesn't get used again
+            word_dict[new_wordings[-2] + ' ' + new_wordings[-1]].pop(0)
+        else:
+            next_word = None
```
natasha-aleksandrova: I think there is a bug in the above block that is stopping it form moving on to add more words... Please review below logic which should use all the words available:
```
while next_word:  # unless we have specific reason, just check for truthiness 
    next_word = word_dict[" ".join(new_wordings[-2:])]  # get last two new words and join with space
    if next_word:  # again just check for truthiness
        new_wordings.append(random.choice(next_word))  # add new word we found, and use random choice from the list
```

students/JerryH/Lesson04/trigrams.py
```diff
@@ -0,0 +1,58 @@
...
+import random
+from collections import defaultdict
+
+
+def get_text():
+    file = open("sherlock.txt", "r")
+    text = file.read()
+    file.close()
+    return text
+
```
natasha-aleksandrova: use `with` statement, it will look something like this:
```
with open('filename', 'r') as fh:  # fh: file handle
    return fh.read()
```

students/JerryH/Lesson04/trigrams.py
```diff
@@ -0,0 +1,58 @@
...
+import string
+import random
+from collections import defaultdict
+
+
+def get_text():
+    file = open("sherlock.txt", "r")
+    text = file.read()
+    file.close()
```
natasha-aleksandrova: keep an eye on var names that overwrite built-in objects, you can see git highlights them in orange

students/JerryH/Lesson04/trigrams.py
```diff
@@ -0,0 +1,58 @@
...
+# create a default list dictionary
+word_dict = defaultdict(list, {})
+words = get_word_list()
+
+new_wordings = [] # Beginning of the new wordings
+r = random.randint(0, len(words))
+
+# Create trigrams in dictionary 'word_dict'
+for i in range(len(words)-2):
+    word_dict[words[i] + ' ' + words[i+1]].append(words[i+2])
```
natasha-aleksandrova: you should be able to move above code into `main()` 

it's common to either put this code under `__name__ == '__main__'` block or into it's own function. basically you either write it as a script or you write it with function/classes

students/JerryH/Lesson04/trigrams.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def get_word_list():
+    text = get_text().replace('\n', ' ').lower()
+    punc = string.punctuation.replace("-", "") #keep "-" in the punctuation
+    table = str.maketrans({key: None for key in punc})
+    new_text = text.translate(table)
+
+    word = new_text.split()
+    return word
```
natasha-aleksandrova: more pythonic to just `return new_text.split()`

PR #93, opened 2018-02-08
Dustin_L - Lesson03 - Mailroom Lab Part I

rdlaner:
Mailroom Lab I Complete.
Addressed comments for string formatting lab.


natasha-aleksandrova:
@rdlaner general comment before I start reviewing in detail.. please run the code through linter 🙂 

you can use online version: http://pep8online.com/
or configure your editor with linter

I'm seeing quite a bit of pep8 violations. thanks!

students/Dustin_L/lesson03/mailroom.py
```diff
@@ -0,0 +1,179 @@
...
+    while True:
+        choice = get_usr_input()
+
+        if choice == THANK_YOU_OPT:
+            send_thank_you()
+        elif choice == REPORT_OPT:
+            create_report()
+        else:
+            break
+
```
natasha-aleksandrova: nice use of functions and constants here for clarity

students/Dustin_L/lesson03/mailroom.py
```diff
@@ -0,0 +1,179 @@
...
+    Returns:
+        int: Value corresponding to user choice
+    """
+    select_prompt = '\nPlease select from the following options:\n'\
+                    '\t1. Send a Thank You\n'\
+                    '\t2. Create a Report\n'\
+                    '\t3. quit\n'\
+                    ' --> '
+    while True:
+        usr_in = int(input(select_prompt).strip())
```
natasha-aleksandrova: good to keep strip() in mind for user inputs, but since you're expecting an int here `strip()` is a redundant call as python will happily convert `'1 '` to `1`

students/Dustin_L/lesson03/mailroom.py
```diff
@@ -0,0 +1,179 @@
...
+
+        if usr_in == THANK_YOU_OPT or usr_in == REPORT_OPT or usr_in == QUIT_OPT:
+            break
```
natasha-aleksandrova: you could simplify this to 
```
usr_in = None
while usr_in not in (THANK_YOU_OPT, REPORT_OPT, QUIT_OPT):
    print("please enter either... ")
```

students/Dustin_L/lesson03/mailroom.py
```diff
@@ -0,0 +1,179 @@
...
+        if usr_in.startswith('q'):
+            break
+        elif usr_in == 'list':
+            print()
+            for dnr in DONATION_DB:
+                print(dnr[NAME_IDX])
+        else:
+            if usr_in in first_names:
+                for i, _ in enumerate(DONATION_DB):
+                    if usr_in in DONATION_DB[i][NAME_IDX].lower():
```
natasha-aleksandrova: instead of using `DONATION_DB[i]` use value that would go into `_` that is already provided by `enumerate`

rdlaner:
Ok, I had set up pylint, but I configured it to disable some of the white space warnings out of personal preference, the main two being whitespace around operators and extending line length from 80 to 100.  Is that acceptable or do I have to strictly adhere to pep8?

natasha-aleksandrova:
it is highly recommended but of course it is up to you if you adhere or not.

from our point of view we want to set good standards so we will always point out violations 🙂 

when it's just you writing python, it doesn't really matter but normally you work on teams where multiple people contribute to the same source code that's where it becomes really important. 

consider just one of the use cases: we work on a team together and I use pycharm with default pep8 rules, and you use a different (or even same ide) with custom rules. Every time I work on the same file you worked on before, my pycharm will auto-format and make changes to correct pep8. Do you see how that could be a problem?

If you work on a team where everyone agrees to use certain exceptions then that's what you do.

In general, it's just easier to have common standards across developers and makes code much more readable when it uses the same style.

----
I could have made my reply more lengthy so I'll just post this here, recommend to read the whys you do and don't follow pep8:
https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

rdlaner:
That makes sense, thanks for the input. With respect to linters, do you have one that you prefer (pylint, flake8, pep8, etc)?

natasha-aleksandrova:
I've been using pycharm for the past two years which comes with built-in pep8 checks.

If you're install as a plugin probably pylint.

this is a pretty good article comparing different linters:
https://blog.sideci.com/about-style-guide-of-python-and-linter-tool-pep8-pyflakes-flake8-haking-pyling-7fdbe163079d

PR #94, opened 2018-02-08
lesson01 and lesson02 submission

DeFeNdog:


students/carlos_novoa/lesson01/break_me.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+
+"""
+Lesson1, Task 1: Explore Errors
+"""
+
+
+def name_error():
+    print('abc'+1)
```
robalford: This function will cause a `TypeError` rather than a `NameError`. In Python, you don't have to name values before performing operations with them. So the error here is from trying to add/concatenate a string and integer (`TypeError`) rather than using a symbol that has not been assigned a value (`NameError`).

students/carlos_novoa/lesson02/fizzbuzz.py
```diff
@@ -0,0 +1,17 @@
...
+def fizzbuzz():
+    for i in range(1, 101):
+        if (i % 5 == 0 and i % 3 == 0):
+            print('FizzBuzz')
+        elif (i % 3 == 0):
+            print('Fizz')
+        elif (i % 5 == 0):
+            print('Buzz')
+        else:
+            print(i)
```
robalford: FizzBuzz looks good. I don't your assignment submission on the course website for this one, though.

robalford:
Nicely done with well commented code in grid printer which really helps with readability for this particular assignment. I left a note to remember to submit FizzBuzz via the course website as well so we can grade it there, and another comment on `break_me.py` but this is ready to merge.

I would recommend running all your code through an online linter to catch Python-specific style violations: http://pep8online.com/. Or you can probably configure your text editor to catch pep8 (Python style guide) violations for you. Adhering to the Pep8 style guide is an important element of writing clean and maintainable Python, especially when working on a team.

students/carlos_novoa/lesson02/fizzbuzz.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+
+"""
+Lesson2, FizzBuz
+"""
+
+
+def fizzbuzz():
+    for i in range(1, 101):
+        if (i % 5 == 0 and i % 3 == 0):
```
robalford: Forgot to mention in my review earlier that `if` statements don't require parentheses in Python so these can all be rewritten as:

`if i % 5 == 0 and i % 3 == 0:`

students/carlos_novoa/lesson01/break_me.py
DeFeNdog: I don't know how I overlooked that. Revised to:

`def name_error():
    print(error)`

students/carlos_novoa/lesson02/fizzbuzz.py
DeFeNdog: Much more clean and Pythonic, thanks

DeFeNdog:
I installed pep8 Sublime packages but am not seeing where I'm getting style violations. Could you point me to a specific instance that I can look at so that I can adjust my dev environment?

robalford:
@DeFeNdog if you follow the instructions [here](https://uwpce-pythoncert.github.io/PythonCertDevel/supplemental/dev_environment/sublime_as_ide.html#anaconda) for installing the Anaconda package for Sublime and adjusting the default settings, you'll see output like this as you work in your editor:

screen shot 2018-02-08 at 9 47 07 pm

The white bordered areas are flagging Pep8 violations, and if you right click anywhere and select `Anaconda|Show error list` you'll get info on each of the specific violations. Generally these are conventions about white space and formatting that won't keep your code from running, but adhering to Pep8 makes your code a lot more readable to other Python devs.

Alternatively, you can just copy and paste your code into the form for the online linter I linked to above and it will give you similar output.

PR #97, opened 2018-02-08
Lesson3 excercise

ZhuojingXie:
lesson03 exercise finished. 

Please let me know some suggestions or feedback about how to improve my code. I can do the homework by myself. However I am pretty sure my code isn't look "beautiful". In a lot of case I know there is a better way to achieve, however I can only use the most silly way because I don't know how to write it in a more efficient way .


students/ZhuojingXie/lesson03/slicing lab.py
```diff
@@ -0,0 +1,45 @@
+#with the first and last items exchanged.
```
natasha-aleksandrova: don't use spaces in python file names; should be `slicing_lab.py`

students/ZhuojingXie/lesson03/list_lab.py
```diff
@@ -0,0 +1,81 @@
...
+fruit1 = input("Tell me another fruit you like> ")
+list1 = [fruit1] + list1
+print("The list now is : ", list1)
+fruit1 = input("Okay, now the last fruit you would like to choose> ")
+list1.insert(0,fruit1)
+print("The list now is : ", list1)
+print("the fruit stating with letter 'P' is ")
+for fruit in list1:
+    if fruit[0] == 'P' or fruit[0] == 'p':
+        print(fruit)
```
natasha-aleksandrova: as a general comment, would be better to have each task as individual function and have some spaces between common logical blocks of code. makes the code more readable 👍

students/ZhuojingXie/lesson03/list_lab.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+#series 1
+list1 =  ["Apples", "Pears", "Oranges" , "Peaches"]
+print(list1)
+fruit1 = input("Tell me a fruit you like > ")
+list1.append(fruit1)
+print(list1)
+num1 = int(input("Now, tell me a numer you like between 1 to {} > ".format(len(list1))))
+while num1<0 or num1 > len(list1):
```
natasha-aleksandrova: another way it can be represented is:
`while not 1 <= num <= len(list1)`

which is saying: while number is not between 1 and size of list inclusive keep looping

natasha-aleksandrova: you want to try and minimize duplicate or common code, so instead of:
```
num1 = int(input("Now, tell me a numer you like between 1 to {} > ".format(len(list1))))
while num1<0 or num1 > len(list1):
    num1 = int(input("Please pick up a number between 1 to {} ".format(len(list1))))
```
you can have a more concise version:
```
num1 = 0
while not 1 <= num1 <= len(list1):
    num1 = int(input(".... ")
```

students/ZhuojingXie/lesson03/list_lab.py
```diff
@@ -0,0 +1,81 @@
...
+print("Okay, seems the number you choose are {}, and the fruit corresponding to that number is {}".format(num1, list1[num1-1]))
+fruit1 = input("Tell me another fruit you like> ")
+list1 = [fruit1] + list1
+print("The list now is : ", list1)
+fruit1 = input("Okay, now the last fruit you would like to choose> ")
+list1.insert(0,fruit1)
+print("The list now is : ", list1)
+print("the fruit stating with letter 'P' is ")
+for fruit in list1:
+    if fruit[0] == 'P' or fruit[0] == 'p':
```
natasha-aleksandrova: you can use 

`fruit[0].lower() == 'p'`

or

`fruit.lower().startswith('p')`

students/ZhuojingXie/lesson03/list_lab.py
```diff
@@ -0,0 +1,81 @@
...
+        print(fruit)
+
+# series 2
+# Multiply the list times two. Keep asking until a match is found.
+# User can type both the names of the fruit of the corresponding number
+# Once found, delete all occurrences
+# series 2
+
+list2 =  list1*2
```
natasha-aleksandrova: ideally you want to use original list (["Apples", "Pears", "Oranges" , "Peaches"]) in each series section.

at the top of the file you could have 
`FRUITS = ["Apples", "Pears", "Oranges" , "Peaches"]` 

then each series function, for example
```
def series1():
    fruits = FRUITS[:]
    # or 
    fruits = deepcopy(fruits)   # from copy import deepcopy
```

this allows you to copy the original fruits into new fruits and do modifications as needed per series

students/ZhuojingXie/lesson03/list_lab.py
```diff
@@ -0,0 +1,81 @@
...
+    for fruitinlist in list2:
+        if fruitinlist == num2:
+            list2.remove(num2)
+            retry = False
+    if num2.isdigit():
+        if int(num2)>=0 and int(num2)<=len(list2):
+            rmfruit = list2[int(num2)-1]
+            for fruitinlist in list2:
+                if fruitinlist == rmfruit:
+                    list2.remove(rmfruit)
```
natasha-aleksandrova: note that this could have been made into a function as it is the same for / if stmts as a little bit above

students/ZhuojingXie/lesson03/list_lab.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+
+# series 3
+nolikelist =[]
+likelist=[]
+
+for fruitinlist in list1:
+    answer = input("Do you like {}, Please type 'Yes' or 'No' > ".format(fruitinlist.lower()))
+    while answer.lower() != "yes" and answer.lower() != "no":
```
natasha-aleksandrova: another more concise way:
`while answer.lower() not in ('yes', 'no')`

students/ZhuojingXie/lesson03/list_lab.py
```diff
@@ -0,0 +1,81 @@
...
+        answer = input("Please try again, type in 'Yes' or 'No' > ")
+    if answer.lower() == 'yes':
+        likelist.append(fruitinlist)
+    else:
+        nolikelist.append(fruitinlist)
+print("According to your answer, the fruit you like are",likelist)
+
+
+# series 4
+list4=list1
```
natasha-aleksandrova: it is not recommended to make a copy this way, because it will make a shallow copy. see below example:
```
>>> seq
[0, 1, 2]
>>> seq_copy = seq
>>> seq
[0, 1, 2]
>>> seq_copy
[0, 1, 2]
>>> seq.append('new_item')
>>> seq
[0, 1, 2, 'new_item']
>>> seq_copy
[0, 1, 2, 'new_item']
>>>
```
as you can notice we didn't make changes to seq_copy .. but it got the change we applied to seq
that is because they point to the same address.

so to make proper copy, you have couple options
1. use deepcopy: 
```from copy import deepcopy
seq_copy = deepcopy(seq)
```
what's great about deepcopy is you can also apply it to dictionaries

2. use slicing to copy
```
seq_copy = seq[:]
```

(there are others .. but lets stick with those two for now)

students/ZhuojingXie/lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+# main menu
+
+
+
+def donor_name():
+    donor_list=[]
+    for info in data:
+        donor_list.append(info[0])
+    return donor_list
+#this function gonna return a list of donaor, and it will be used in thankprompt function later
```
natasha-aleksandrova: you can also put these type of comments as docstrings like:
```
def donor_name():
    """This functions returns a list of donor names"""
```

students/ZhuojingXie/lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+
+def thankprompt():
+    user_select_thank=input('\n\nType "list" to show a list of the donor names, \n'
+                            'Or type "back" to Back\n')
+    while user_select_thank.lower() != 'back' and user_select_thank.lower() != 'list':
+        user_select_thank=input('\n\nType "list" to show a list of the donor names, \n'
+                            'Or type "back to Back\n\n')
```
natasha-aleksandrova: similar to comment before to avoid duplication, you could have
```
user_select_thank = None
while user_select_thank not in ('back', 'list'):
    user_select_thank = input("....").lower()
```

natasha-aleksandrova:
@ ZhuojingXie great work!

couple general comments:
* run your code through linter for style violations, overall looks good .. but a few violations here and there
* highly recommended to submit each assignment individually, as you can see there are a lot of comments and it is more difficult to provide even more detailed feedback as the Pull Request becomes unmanageable and much harder for you to consume all of the feedback.
* I would recommend doing another pass on list lab and mailroom. It is not required, but if you would like to iterate and improve your python coding then we are happy to review again.

students/ZhuojingXie/lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+#this function will Create a report
+
+def main():
+    a = prompt()
+    if a == '1':
+        thankprompt()
+    elif a =='2':
+        report()
+    elif a == '3':
+        sys.exit()
```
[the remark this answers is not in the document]
ZhuojingXie: a quick question here. At beginning I wrote:

def main():
+    if prompt == '1':
+        thankprompt()
+    elif prompt =='2':
+        report()
+    elif prompt == '3':
+        sys.exit()

Somehow it make sense to me. However when i run the code, it show a critical bug . I still don't understand why and what happen here??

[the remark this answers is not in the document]
natasha-aleksandrova: quick tip on github, use 3 back ticks at the beginning of your code block and 3 back ticks at the end of your code block (backticks -> ```)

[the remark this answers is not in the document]
natasha-aleksandrova: so if I understand correctly in this: `if prompt == '1'` , prompt is the function object right?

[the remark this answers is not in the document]
ZhuojingXie: yes.  just like the main() function I have above. prompt() is another function will return either '1', '2', or '3' based on user select 'Thank you' ' report' or 'exit'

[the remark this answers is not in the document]
natasha-aleksandrova: ok so you were comparing if function object equals to 1, which we would always evaluate to False

what you probably intended to do is to call a function, so it should actually look like: 
`if prompt() == '1'`

however, that is still not desired; your final version is more correct, because you are testing the return value of prompt() function call, that return value is either 1, 2 or 3.

does that make sense?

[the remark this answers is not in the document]
natasha-aleksandrova: this is difference:
```
>>> def test(): return '1'
...
>>> test
<function test at 0x10139de18>
>>> test()
'1'
>>>```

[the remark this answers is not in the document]
ZhuojingXie: roger. it makes so much sense

PR #99, opened 2018-02-08
Grid PR lesson 2

preactive:


robalford:
@preactive Please submit your assignment via the course website as well. Thanks!

preactive:
Copy @robalford will do.  I think i got session timeout issues on edx before I jumped off the train.  


Thanks!

students/preactive/Lesson2/grid-alt.py
```diff
@@ -0,0 +1,24 @@
+def formation(gridSize, gridMulti):
+    plusEl = "+ "
```
robalford: Python conventions is to use 'snake case' for all variable, function and parameter names. So these should be `grid_size`, `grid_multi`, `m_el` etc.

students/preactive/Lesson2/fizzbuzz.py
```diff
@@ -0,0 +1,10 @@
+for i in range(1,100):
```
robalford: This should be for i in range(1, 101) for 1 through 100 inclusive.

students/preactive/Lesson2/fizzbuzz.py
```diff
@@ -0,0 +1,10 @@
...
+    currentprint = str(i)
+    if i % 3 == 0:
+        currentprint = currentprint.replace(str(i),"")
+        currentprint += "fizz"
+    if i % 5 == 0:
+        currentprint = currentprint.replace(str(i),"")
+        currentprint += "buzz"
+    print(currentprint)
+    currentprint = ""
```
robalford: This last line is not necessary since you're resetting the `current print` value at the beginning of the next iteration

robalford:
@preactive saw your note in the course submission field about looking at grid-alt.py as the final version, rather than Grid.py.

The nice thing about git is that you don’t need to save multiple versions of the same file. Just commit your new changes to the file and git will store a history of every change you’ve made, so you can always refer back to your previous work if you want to. 

Please submit the FizzBuzz assignment and upload your file to the course website as well so we can grade it there.

Python has its own style rules that are different than some of the C family languages. I would recommend running all your code through an online linter to catch Python-specific style violations: http://pep8online.com/. Or you can probably configure your text editor to catch pep8 (Python style guide) violations for you.

students/preactive/Lesson2/fizzbuzz.py
preactive: Fixed

students/preactive/Lesson2/fizzbuzz.py
preactive: removed line 10

students/preactive/Lesson2/grid-alt.py
preactive: Adjusted variable names to reflect snake case and pep8

PR #103, opened 2018-02-09
Lesson 03 - Slicing - Emmanuel Twum

etwum:
assignment complete

### DIFF
--- a/students/etwum/lesson03/slicing.py
+++ b/students/etwum/lesson03/slicing.py
@@ -15,9 +15,9 @@ def exchange_first_last(seq):
         return print(new_tuple)
 
     else:
-        last_item2 = seq[-1]
+        last_item2 = [seq[-1]]
         middle_items2 = seq[1:-1]
-        first_item2 = seq[0]
+        first_item2 = [seq[0]]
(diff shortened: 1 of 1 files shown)


students/etwum/lesson03/slicing.py
```diff
@@ -0,0 +1,147 @@
+
+def exchange_first_last(seq):
+    # exchanges the first/last items in a sequence then copies the new sequence into a new string, list or tuple
+
+    # performs this part of the if statement if the sequence is of type string
+    if type(seq) is str:
```
natasha-aleksandrova: `isinstance` check is preferred over `type` checks:
https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/103/files

students/etwum/lesson03/slicing.py
```diff
@@ -0,0 +1,147 @@
...
+        y = seq[-1] + seq[1:-1] + seq[0]
+        new_string = y
+        return new_string
```
natasha-aleksandrova: you can just do `return seq[-1] + seq[1:-1] + seq[0]`, no need for extra variables

PR #106, opened 2018-02-09
Finished lesson02 / Fibonacci series exercise

22e88:
finished series.py


students/elmar_m/lesson02/series.py
```diff
@@ -0,0 +1,140 @@
...
+        startvalues = [0, 1]
+        counter = 0
+        while counter <= n - 2:
+            a = startvalues[0]
+            b = startvalues[1]
+            c = a + b
+            startvalues[0] = b
+            startvalues[1] = c
+            counter += 1
+        return c
```
robalford: This is a totally valid solution to the problem and your code is highly readable and well written throughout. But since we know that the fibonacci series is naturally recursive, that is: `fib(n) = fib(n-2) + fib(n-1)`, can you think of a more concise way to rewrite this as a recursive Python function (a function that calls itself)?

students/elmar_m/lesson02/series.py
```diff
@@ -0,0 +1,140 @@
...
+    return c
+
+'''
+assert statements ensure that functions work in a specified way. They are 
+executed immediately when a module is imported and throw AssertionError if
+their expression does not result to "True". So the following statements
+ensure that the calculated results are correct. 
+'''
+fib_results = [0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 610, 987, 1597, 2584, 4181, 6765]
+for i in range(0, 20):
```
robalford: You could just do `for i in range(len(fib_results)):` here

robalford:
Nice work! Left a couple suggestions for your consideration, but this is ready to merge.

PR #107, opened 2018-02-10
assignment2 part 1 - grid pull request

hortw1991:
I'm sure I put in some unnecessary things but it was fun including things like if name == main.  Any feedback or improvements would be great!

### DIFF
--- a/students/thorn/00-README.md
+++ b/students/thorn/00-README.md
@@ -1,0 +1,1 @@
+UWPCE - Python Cert :: Thomas Horn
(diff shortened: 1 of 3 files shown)


hortw1991:
Added the completed FizzBuzz file.

natasha-aleksandrova:
@hortw1991 grid printer is a very open ended exercise and a good way to jump into python; as we get into further exercises that are more focused we'll be able to provide more specific feedback.

PR #110, opened 2018-02-10
added lesson03/seqslicing.py

johnpharmd:


students/johnpharmd/lesson03/seqslicing.py
```diff
@@ -0,0 +1,104 @@
+def seqxchg(seq):
+    """
+    This function returns a copy of the sequence passed but with
+    the first and last items exchanged.
+    """
+    # new_seq = [seq[-1], seq[1:-1], seq[0]] -- this failed
```
natasha-aleksandrova: I'm glad that you left this comment, because you were close..

let's break it down:
let's say seq is `[1, 2, 3, 4]`
[seq[-1], seq[1:-1], seq[0]] would return: `[4, [2, 3], 1]` not really what we want right? as you notice accessing list by an index returns a single item, but slicing returns another list. So let's stick to slicing only and use `+` to add the list parts together

so we get:
`seq[-1:] + seq[1:-1] + seq[:1]`

natasha-aleksandrova: also for this exercise, there should be no difference in slicing logic for list, string or a tuple so type checking is not necessary.

a side note, it is preferred to use `isinstance` method instead of `type` to check the type when you _do_ need to.

students/johnpharmd/lesson03/seqslicing.py
```diff
@@ -0,0 +1,104 @@
...
+
+    return new_seq
+
+
+def rm_everyother(seq):
+    """
+    This function returns the sequence passed but with every
+    other item removed.
+    """
+    return seq[::2]
```
natasha-aleksandrova: 👍

students/johnpharmd/lesson03/seqslicing.py
```diff
@@ -0,0 +1,104 @@
...
+
+def rm_firstlast4_odds(seq):
+    """
+    The output from this function is a copy of the sequence passed
+    but with the first 4 and last 4 items removed and then only
+    every other item of the remaining original sequence.
+    """
+    new_seq = None
+    if type(seq) == str:
+        new_seq = seq[4:-5]
```
natasha-aleksandrova: you were really close here as well.. instead of `seq[4:-5]` it should be `seq[4:-4]` because you need to remove last 4. lastly..  all that is left is to apply step value same as used in `rm_everyother`

so final solution is: `return seq[4:-4:2]`

(again, no need for type checking)

students/johnpharmd/lesson03/seqslicing.py
```diff
@@ -0,0 +1,104 @@
...
+            new_seq = tuple(new_seq)
+
+    return rm_everyother(new_seq)
+
+
+def rev(seq):
+    """
+    This function returns the sequence passed but in reverse order.
+    """
+    return seq[::-1]
```
natasha-aleksandrova: 👍

students/johnpharmd/lesson03/seqslicing.py
```diff
@@ -0,0 +1,104 @@
...
+    
+    if type(seq) == list or type(seq) == tuple:
+        new_seq = []
+        new_seq.extend(seq[div : div * 2])
+        new_seq.extend(seq[div * 2 : ])
+        new_seq.extend(seq[ : div])
+        if type(seq) == tuple:
+            new_seq = tuple(new_seq)
+
+    return new_seq
```
natasha-aleksandrova: how could you do it with slicing only ? (and apply same solution to list/str/tuple)

PR #112, opened 2018-02-10
completed mailroom assignment

rmart300:


students/rmart300/mailroom/mailroom.py
```diff
@@ -0,0 +1,65 @@
+from datetime import date
+
+#dictionary to hold donors and list of donation amounts
+donation_dict = {}
```
natasha-aleksandrova: looks like you're ahead of the game since we will need to use `dict` in the next part of the mailroom 🙂

students/rmart300/mailroom/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+    
+        for amount in amount_list:
+            donation_list = donation_dict[donor] 
+            donation_list.append(amount)
+            donation_dict[donor] = donation_list
+
+    #prompt user for action and then call function
+    action = ''
+    while action != 'quit':
+        action = input("What would you like to do: \“Send a Thank You\”, \“Create a Report\” or \“quit\”\n")
```
natasha-aleksandrova: what is the purpose of `\` ?

if you attempted to escape the quotes then no need because these are different type of quotes than what is used for the string (note that you could always use `'` single quote for the string if you wanted to use double quotes within the actual string)

or maybe they were supposed to be new lines?

natasha-aleksandrova: suggestion to make the program more user friendly 😉 

asks the user for a number instead of making the user type out the string and possibly misspell it

students/rmart300/mailroom/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+    donor_list = ['Fred Smith','Terrie Ann','Murray Martin','Josh Jones','Jane Doe']
+    amount_list = [500,100,1000]
+    for donor in donor_list:
+        if donor not in donation_dict:
+            donation_dict[donor]=[]
+    
+        for amount in amount_list:
+            donation_list = donation_dict[donor] 
+            donation_list.append(amount)
+            donation_dict[donor] = donation_list
```
natasha-aleksandrova: this is a little bit confusing, let's simplify it a bit!

```
for donor in donor_list:
    donation_dict[donor] = amount_list[:]
```

since you're only initializing here, you don't really need all the extra logic there

students/rmart300/mailroom/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+
+    name = 'list'
+    while name == 'list':
+        name = input("Provide full name: ")
+        if name == 'list':
+            for donor in donation_dict:
+                print(donor)
+    
+    amount = float(input("Provide a donation amount:"))
+    donation_list = donation_dict[name] if name in donation_dict else []
```
natasha-aleksandrova: very nice

PR #113, opened 2018-02-11
Andrew Park | Lesson 4 mailroom2

nDruP:
Pull request for mailroom2 assignment.

### DIFF
--- a/students/nDruP/lesson04/mailroom2.py
+++ b/students/nDruP/lesson04/mailroom2.py
@@ -77,7 +77,7 @@ def craft_thank_u():
         donation_amt = float(donation_amt)
 
         if donor_name not in donor_dict:
-            donor_list[donor_name] = []
+            donor_dict[donor_name] = []
         donor_dict[donor_name].append(donation_amt)
         print_divider()
         message = f"Dearest {donor_name},\n"
@@ -145,6 +145,7 @@ def create_letters():
             letter.write(thank_you)
             letter.close()
         print("Finished writing the letters")
+        break
     return
 
 
--- a/students/nDruP/lesson04/Andrew_Park_2018_02_10.txt
+++ b/students/nDruP/lesson04/Andrew_Park_2018_02_10.txt
@@ -1,0 +1,4 @@
+Dear Andrew Park,
+	We are so grateful for your lifetime generosity of $1234362.00! You now have our eternal loyalty. Use it wisely.
+Sincerely, 
+We're a pyramid scheme and so is Andrew Park
--- a/students/nDruP/lesson04/Bobson_Dugnutt_2018_02_10.txt
+++ b/students/nDruP/lesson04/Bobson_Dugnutt_2018_02_10.txt
@@ -1,0 +1,4 @@
+Dear Bobson Dugnutt,
+	We are so grateful for your lifetime generosity of $2023.56! You now have our eternal loyalty. Use it wisely.
+Sincerely, 
+We're a pyramid scheme and so is Bobson Dugnutt
--- a/students/nDruP/lesson04/Mike_Truk_2018_02_10.txt
+++ b/students/nDruP/lesson04/Mike_Truk_2018_02_10.txt
@@ -1,0 +1,4 @@
+Dear Mike Truk,
+	We are so grateful for your lifetime generosity of $83794.00! You now have our eternal loyalty. Use it wisely.
+Sincerely, 
+We're a pyramid scheme and so is Mike Truk
--- a/students/nDruP/lesson04/Rey_McScriff_2018_02_10.txt
+++ b/students/nDruP/lesson04/Rey_McScriff_2018_02_10.txt
@@ -1,0 +1,4 @@
+Dear Rey McScriff,
+	We are so grateful for your lifetime generosity of $666.00! You now have our eternal loyalty. Use it wisely.
+Sincerely, 
+We're a pyramid scheme and so is Rey McScriff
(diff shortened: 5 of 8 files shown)


students/nDruP/lesson04/mailroom2.py
```diff
@@ -0,0 +1,167 @@
...
+
+def main_menu():
+    """
+    Prompt user to send a Thank You, Create a report, create letters, or quit.
+    """
+    user_prompt = None
+    valid_prompts = {"1": craft_thank_u,
+                     "2": create_report,
+                     "3": create_letters,
+                     "4": sys.exit}
```
natasha-aleksandrova: very nice!

students/nDruP/lesson04/mailroom2.py
```diff
@@ -0,0 +1,167 @@
...
+    while True:
+        print_divider()
+        print("We're a Pyramid Scheme & So Are You! E-Mailroom")
+        print_divider()
+        while user_prompt not in valid_prompts:
```
natasha-aleksandrova: 👍

students/nDruP/lesson04/mailroom2.py
```diff
@@ -0,0 +1,167 @@
...
+            print("Please choose from the following options (1,2,3,4): ")
```
natasha-aleksandrova: could use string format here for options and use values from `valid_promts.keys()`

students/nDruP/lesson04/mailroom2.py
```diff
@@ -0,0 +1,167 @@
...
+            break
+
+        donation_amt = input("Enter their Donation Amount >")
+        if donation_amt.lower() == "exit":
+            break
+        donation_amt = float(donation_amt)
+
+        if donor_name not in donor_dict:
+            donor_dict[donor_name] = []
+        donor_dict[donor_name].append(donation_amt)
```
natasha-aleksandrova: I recommend taking a look at `defaultdict`, it's great for cases like this and makes things much simpler

students/nDruP/lesson04/mailroom2.py
```diff
@@ -0,0 +1,167 @@
...
+        print_divider()
+        message = f"Dearest {donor_name},\n"
+        message += f"Thank you so much for donation of ${donation_amt:.2f}!\n"
+        message += "We will use your donation to create a real living Pokemon."
+        message += f"\nSincerely,\nWe're a Pyramid Scheme & so is {donor_name}"
```
natasha-aleksandrova: note that there are two more ways to do multi-strings that are a bit more readable

use triple quotes (in this case no need for \n)

```
s = f"""Dearest {donor_name}
Thank you so much...
"""

```
or with parens (you will need \n)
```
>>> s = (f"Dearest {donor_name}\n"
...     "thank you so much...")
>>>
```

students/nDruP/lesson04/mailroom2.py
```diff
@@ -0,0 +1,167 @@
...
+        print(message)
+        print_divider()
+
+        break
+    return
```
natasha-aleksandrova: no need for return here

students/nDruP/lesson04/mailroom2.py
```diff
@@ -0,0 +1,167 @@
...
+
+def create_report():
+    """
+    Print a list of donors sorted by total historical donation amount.
+    Donor Name, Total Given, Num Gifts, Average Gift
+    """
+    sort_donors = sorted(donor_dict.items(), key=sum_2tuple_by2, reverse=True)
+    col_name_list = ["Donor Name", "Total Given", "Gifts", "Avg Donation"]
+    billi_len = 12
+    donor_col_len = max([len(name) for name, donations in sort_donors])
```
natasha-aleksandrova: common pattern is to use `_` for var names for data you do not care about to indicate that clearly

for example:
```
 max([len(name) for name, _ in sort_donors])
```

students/nDruP/lesson04/mailroom2.py
```diff
@@ -0,0 +1,167 @@
...
+            thank_you = "Dear " + donor + ",\n\t"
+            thank_you += "We are so grateful for your lifetime generosity of "
+            thank_you += f"${sum(donations):.2f}!"
+            thank_you += " You now have our eternal loyalty. Use it wisely."
+            thank_you += "\nSincerely, \n"
+            thank_you += "We're a pyramid scheme and so is " + donor
+            file_name = donor.replace(' ', '_') + "_" + curdate + ".txt"
+            letter = open(write_dir+"/"+file_name, 'w+')
+            letter.write(thank_you)
+            letter.close()
```
natasha-aleksandrova: it is preferred to use `with` statement when working with files (it is also mentioned in the course material)

it will ensure that the file gets closed

students/nDruP/lesson04/mailroom2.py
```diff
@@ -0,0 +1,167 @@
...
+            write_dir = os.getcwd()
+        for donor, donations in donor_dict.items():
+            thank_you = "Dear " + donor + ",\n\t"
+            thank_you += "We are so grateful for your lifetime generosity of "
+            thank_you += f"${sum(donations):.2f}!"
+            thank_you += " You now have our eternal loyalty. Use it wisely."
+            thank_you += "\nSincerely, \n"
+            thank_you += "We're a pyramid scheme and so is " + donor
+            file_name = donor.replace(' ', '_') + "_" + curdate + ".txt"
+            letter = open(write_dir+"/"+file_name, 'w+')
```
natasha-aleksandrova: this is totally correct, for future reference though you will want to use `os.path.join` instead of manually creating full path

read more: https://stackoverflow.com/questions/13944387/why-use-os-path-join-over-string-concatenation

PR #116, opened 2018-02-11
GIT setup complete

traub01-devtrain:
My git setup is complete, I added my student folder, a readme, and the lesson one break_me.py file.


students/JamesTraub/break_me.py
```diff
@@ -0,0 +1,24 @@
...
+    return terr
+
+def excNameErr():
+    nerr == 2 + 2
+
+#def excSyntaxErr):
+#    a = 0
+#    b = 1
+
+def excAttErr():
```
natasha-aleksandrova: per python style use `snake_case`, so this would be `def exc_att_err`

PR #117, opened 2018-02-11
addressed natasha's comments

rmart300:


students/rmart300/mailroom/mailroom.py
```diff
@@ -44,22 +51,18 @@ def create_report():
...
+        action = input('What would you like to do: 1 - \“Send a Thank You\”, 2 - \“Create a Report\” or 3 - \“quit\”\n')
         
-        if action.lower() == "send a thank you":
+        if not is_number(action) or int(action) < 1 or int(action) > 3:
+            print("please enter a number 1-3")
+        elif int(action) == 1:
             send_thank_you()
-        elif action.lower() == "create a report":
-            print("got here")
+        elif int(action) == 2:
```
natasha-aleksandrova: I think we can make this block even cleaner in the next portion of mailroom where we'll get to use try/except syntax. so keep that in mind for next assignment 😉

PR #118, opened 2018-02-11
added list_lab.py with series1, series2

johnpharmd:
putting in this commit to request feedback on accuracy of series1 and series2, and to ask for your thoughts on my use of string interpolation in series 3. thx!


students/johnpharmd/lesson03/list_lab.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python3
+
+
+def get_fruits():
+    """
+    Create a list that contains “Apples”, “Pears”, “Oranges” and “Peaches”.
+    """
+    fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
```
robalford: You could just declare `fruits` as a global variable and then pass it in to your series functions below

students/johnpharmd/lesson03/list_lab.py
```diff
@@ -0,0 +1,116 @@
...
+    # Remove the last fruit from the list.
+    fruits.pop()
+    # Display the list.
+    print('Popped last item. List now contains:', fruits)
+    
+    # Ask the user for a fruit to delete, find it and delete it.
+    response = input('Enter a fruit to be removed: ')
+    while not response.isalpha():
+        response = input('Only words, please. Remove which fruit?: ')
+    response.lower()
```
robalford: You don't need to call `response.lower()` here since the `capitalize` string method works on any string:

```
In [5]: 'HJHKFDS'.capitalize()
Out[5]: 'Hjhkfds'

In [6]: 'askHHksljdfkK'.capitalize()
Out[6]: 'Askhhksljdfkk'
```

students/johnpharmd/lesson03/list_lab.py
```diff
@@ -0,0 +1,116 @@
...
+    response.capitalize()
+    while response not in fruits:
```
robalford: Because strings are immutable in Python `response.capitalize()` won't change the value of `response` unless you reassign it like `response = response.capitalize()`. Doing this will fix your case check validation here.

students/johnpharmd/lesson03/list_lab.py
```diff
@@ -0,0 +1,116 @@
...
+    # Call get_fruits to create the fruits list.
+    fruits = get_fruits()
+    
+    # Create copy of fruits list, iterate over copy, mutate original.
+    """
+    Ask the user for input displaying a line like “Do you like apples?”
+    for each fruit in the list (making the fruit all lowercase).
+    """
+    new_fruits = fruits
+    for item in new_fruits:
```
robalford: Nice work using the slicing operator to loop through a copy and mutate the original list. Could also be written as `for item in fruits[:]:`

students/johnpharmd/lesson03/list_lab.py
```diff
@@ -0,0 +1,116 @@
...
+        print('Do you eat {:s}?'.format(item).lower())
+        response = input('Enter "y" for yes or "n" for no: ')
+
```
robalford: You may want to do `response.lower()` here to accept 'y' or 'Y'

students/johnpharmd/lesson03/list_lab.py
```diff
@@ -0,0 +1,116 @@
...
+        print('Do you eat {:s}?'.format(item).lower())
+        response = input('Enter "y" for yes or "n" for no: ')
+        
+        # For each “no”, delete that fruit from the list.
+        if response == 'n':
+            fruits.remove(item)
+        """
+        For any answer that is not “yes” or “no”, prompt the user to answer
+        with one of those two values (a while loop is good here).
+        """
```
robalford: This could be rewritten more concisely as an `else` statement since we only end up here if `response is not 'y' or 'n'`

robalford:
@johnpharmd looking good overall, see my comments for a few notes to consider.

students/johnpharmd/lesson03/list_lab.py
robalford: Although I should note that if this module were part of a larger system, you would want to be very careful about defining variables in the global namespace, so the way you're doing this here is fine as well

PR #119, opened 2018-02-11
Dustin_L - Dict Lab and File Lab

rdlaner:
Practice problems for dictionaries and file I/O. Please let me know any recommendations you have on making this more pythonic, particularly file_lab.py. Thanks!


students/Dustin_L/lesson04/file_lab.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+import os
+import os.path as osp
```
natasha-aleksandrova: although this is totally valid, I would use it sparingly

someone reading your code will not automatically know what `osp` module is and might get confused

some of the common use cases for using `as` in modules imports are:
* you are importing two modules with the same name
* you are dealing with different versions of modules and the same lib was renamed or have to use replacement similar lib
* the name of the module is really long or really confusing
* adhering to common pattern in the project or python community

students/Dustin_L/lesson04/file_lab.py
```diff
@@ -0,0 +1,33 @@
...
+
+def file_copy(src_path, dst_path):
+    """Copy file from src_path to dst_path
+
+    Args:
+        src_path (str): path to source file
+        dst_path (str): path to copy destination
+    """
+    chunk_size = 100
+    dst_path = osp.join(dst_path, osp.basename(src_path))
```
natasha-aleksandrova: 👍  on using `join`

students/Dustin_L/lesson04/file_lab.py
```diff
@@ -0,0 +1,33 @@
...
+
+    with open(src_path, 'rb') as src_f:
+        with open(dst_path, 'wb') as dst_f:
```
natasha-aleksandrova: 👍  for using `with`

PR #120, opened 2018-02-11
Lesson02 Fibonacci and Lucas assignment and style edits

DeFeNdog:


students/carlos_novoa/lesson02/series.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def fibonacci():
+    """Return a list containing a fibonacci series (starting at 0, 1)"""
+    return sum_series(False, 0, 1)
+
+
+def lucas():
+    """Return a list containing a lucas series (starting at 2, 1)"""
+    return sum_series(False, 2, 1)
```
natasha-aleksandrova: it looks like you would like the first argument to be optional, in this case you should make `n` keyword argument as well: `def sum_series(n=None, a=0, b=0)` then u could call it `rum_series(a=2, b=1)`

students/carlos_novoa/lesson02/series.py
```diff
@@ -0,0 +1,81 @@
...
+assert sum_series(False, 0, 1) == [0, 1, 1, 2, 3, 5, 8, 13, 21, 34]
+
+
+# sum_series(False, 2, 1) returns correct
+# sequence of Lucas numbers (given 2, 1 args)
+assert sum_series(False, 2, 1) == [2, 1, 3, 4, 7, 11, 18, 29, 47, 76]
+
+
+# sum_series(False, 4, 8) returns a list of a different series
+assert isinstance(sum_series(False, 4, 8), list)
```
natasha-aleksandrova: 👍  for adding tess

students/carlos_novoa/lesson02/series.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+# lucas() returns list that matches correct
+# sequence of Lucas numbers (starting at 2, 1)
+assert lucas() == [2, 1, 3, 4, 7, 11, 18, 29, 47, 76]
+
+
+# sum_series(index) returns explicit True if element prints
+# (prints sequence num in console)
+assert sum_series(3) is True
```
natasha-aleksandrova: you should have your function return the value at the index `n` then that way you can assert:
`assert sum_series(9) == 34  # fibonacci sequence`

students/carlos_novoa/lesson02/series.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def series_int(n, a=0, b=1):
+        for i in range(n):
+            a, b = b, a + b
+        return a
+
+    series_range = 10  # limit 10 to avoid maximum recursion error
+    for i in range(series_range):
+        si = series_int(i, a, b)
+        sequence.append(si)
```
natasha-aleksandrova: looks like your fib/lucas sequences can only go up to ten, can you think of another solution that either uses recursion or for/while loop to generate and find seq at any index?

PR #121, opened 2018-02-11
Lesson 02 Assignments

stuartkershaw:
I wanted to make the same function achieve both grid versions, but found the argument ordering made that a little ugly. To keep aligned with the assignment directions, I broke it up into two functions as directed.

### DIFF
--- a/students/stuartkershaw/lesson02/fizzbuzz.py
+++ b/students/stuartkershaw/lesson02/fizzbuzz.py
@@ -1,0 +1,13 @@
+counter = 1
+limit = 100
+
+while counter <= limit:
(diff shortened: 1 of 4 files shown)


students/stuartkershaw/lesson02/fizzbuzz.py
```diff
@@ -0,0 +1,13 @@
+counter = 1
+limit = 100
+
+while counter <= limit:
```
natasha-aleksandrova: take a look at `range` function in python, it makes these like this a little bit easier (not having to keep track of counter)

PR #124, opened 2018-02-12
Herela-lesson 3: strformat, and lesson 4: mailroom2

harvey-herela:


students/HarveyHerela/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python
+
+def task_one():
+    """Task 1: Print out this tuple."""
+    t1_tuple = (2, 123.4567, 10000, 12345.67)
+
+    # create a string formatter to produce:
+    # 'file_002 :   123.46, 1.00e+04, 1.23e+04'
+
+    print("file_{:0>3d} :   {:.2f}, {:.2e}, {:.3e}".format(t1_tuple[0], t1_tuple[1], t1_tuple[2], t1_tuple[3]))
```
robalford: You can use `{:.2e}` in this case for three significant digits in the final number. For Pep8 line length, the following format is acceptable in Python:

```
    print("file_{:0>3d} :   {:.2f}, {:.2e}, {:.2e}".format(
        t1_tuple[0], t1_tuple[1], t1_tuple[2], t1_tuple[3])
    )
```

students/HarveyHerela/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,83 @@
...
+    
+def task_two():
+    """Task 2: Same as task 1, but different."""
+    t1_tuple = (2, 123.4567, 10000, 12345.67)
+    
+    print(f"file_{t1_tuple[0]:0>3d} :   {t1_tuple[1]:.2f}, {t1_tuple[2]:.2e}, {t1_tuple[3]:.3e}")
+
```
robalford: Pep8 convention is 2 blank lines between module level functions

students/HarveyHerela/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,83 @@
...
+def task_three(tuple_in):
+    """Task 3: Print out arbitrary length tuples."""
+    t3_str = "the {} numbers are: ".format(len(tuple_in))
+    
+    if len(tuple_in) > 0:
+        t3_str += "{}".format(tuple_in[0]);
```
robalford: Pesky semicolon ;)

students/HarveyHerela/lesson04/mailroom2.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python
+
+import datetime
+import os
+
+donors = [{"firstname": "Chase", "lastname": "Zuma", "donations": [25]},
```
robalford: If you were to use a `dict` for your outer data structure here with a unique key (full name?) for each 'record', you could simplify some of your look up code and input prompting loops below

students/HarveyHerela/lesson04/mailroom2.py
```diff
@@ -0,0 +1,130 @@
...
+          {"firstname": "Marshall", "lastname": "Rocky", "donations": [30, 60]},
+          {"firstname": "Skye", "lastname": "Rubble", "donations": [100, 100, 100]}]
+
+def get_name(name_type):    
+    gettingInput = True
```
robalford: Python convention is to use `snake_case` for all variable and function names

students/HarveyHerela/lesson04/mailroom2.py
```diff
@@ -0,0 +1,130 @@
...
+    report += "Donor Name    | Total Given   | Num Gifts | Average Gift"
+    report += "\n"
+    report += "--------------------------------------------------------" 
+    report += "\n"
+    # Create a list with the data to be printed
+    data_list = list()
+    for donor_data in donors:
+        fullname = "{firstname} {lastname}".format(**donor_data)
+        gift_data = [fullname, 0, 0, 0]
+        for donation in donor_data["donations"]:
```
robalford: You could eliminate this inner loop and make your code a bit more readable here by using some of Python's built in functions

```
for donor_data in donors:
        fullname = "{firstname} {lastname}".format(**donor_data)
        total_donations = sum(donor_data["donations"])
        number_of_donations = len(donor_data["donations"])
        avg_donation = total_donations / number_of_donations
        data_list.append(
            [fullname, total_donations, number_of_donations, avg_donation]
        )
```

students/HarveyHerela/lesson04/mailroom2.py
```diff
@@ -0,0 +1,130 @@
...
+    line_fmtr = "{name:<14}  ${total:>13.2f}   {num:>9d}  ${avg:>12.2f}\n".format
+    for d in data_list:
+        report += line_fmtr(name=d[0], total=d[1], num=d[2], avg=d[3])
+    
+    print("\n", report)
+
+def send_letters_everyone():
+    """Creates a letter for everyone in the database, and writes them to file."""
+    num = 0
+    subdir = str(datetime.date.today())
```
robalford: You could use `datetime.datetime.now()` here so that a new directory would be created every time the function is called (in this version the user is limited to one chance per day to send letters). Or better yet, use a timestamp on the letters themselves and just keep adding them to the existing folder.

It's still a great idea to use a `try/except` block here, but instead of trying to guess what the specific error might be, you could have a more generic error message for your user and log the specific error for reference. The simplest way to 'log' an exception is just to use a print statement by calling `print(oserr)` inside your try block, or you could pass `osser` to a format string for a more user friendly error message.

robalford:
@harvey-herela nice work here. Good use of functions for organizing and reusing code in mailroom and it seems like you have a really good handle on string formatting. 

I left a few comments to consider but this is ready to merge. 

I would recommend running all your code through an online linter to catch Python-specific style violations: http://pep8online.com/. Or you can probably configure your text editor to catch pep8 (Python style guide) violations for you.

PR #125, opened 2018-02-12
submitting list_lab.py

tricaonguyen:
Hi Natasha,

I am submitting my list_lab.py.
My while loop in Series 3 is clunky. Maybe you can show me a better way to achieve it.

### DIFF
--- a/students/tri_nguyen/lesson03/list_lab.py
+++ b/students/tri_nguyen/lesson03/list_lab.py
@@ -1,0 +1,179 @@
+#!/usr/bin/env python3
+
+# ------------------------------------------- #
+# Student:   Tri Nguyen
+# Instructor: Natasha Aleksandrova
+# Date:  09-Feb-2018
+# ------------------------------------------- #
+
+# ------------------------------------------- #
+# Series 1:
+# - Create a list that contains 'Apples', 'Pears',
+#   'Oranges' and 'Peaches.
+# - Display the list.
+# - Ask the user for another fruit and add it
+#   to the end of the list.
+# - Display the list.
+# - Ask the user for a number and display the number
+#   back to the user and the fruit corresponding
+#   to that number.
+# - Add another fruit to the beginning of the list
+#   using '+' and display the list.
+# - Add another fruit to the beginning of the list
+# - using insert() and display the list.
+# - Display all the fruits that begins with 'P',
+#   using a for loop.
+
+
+fruit_list = ['Apples', 'Pears', 'Oranges', 'Peaches']
+
+print('--Series 1 of the exercise--'.upper()+'\n')
+
+print('Below is the current list of fruits.')
+for idx, fruit in enumerate(fruit_list):
+    print(idx + 1, '-', fruit)
+
+print()
+user_input = input('Please add another fruit: ')
+fruit_list.append(user_input.title())
+
+print()
+print('Below is a newly updated list of fruits.')
+for idx, fruit in enumerate(fruit_list):
+    print(idx + 1, '-', fruit)
+
+print()
+num_from_user = int(input('Please pick a number to get back the corresponding fruit: '))
+num_from_user = num_from_user - 1  # Python uses zero based indexing
+
+if fruit_list[num_from_user] in fruit_list[num_from_user]:
+    print(num_from_user + 1, '-', fruit_list[num_from_user])
+
+print()
+fruit_list = ['Watermelons'] + fruit_list
+print('Watermelons has been added to the beginning of the list.')
+for idx, fruit in enumerate(fruit_list):
+    print(idx + 1, '-', fruit)
+
+print()
+fruit_list.insert(0, 'Pinnaples')
+print('Pinnaples have been added to the beginning of the list.')
+for idx, fruit in enumerate(fruit_list):
+    print(idx + 1, '-', fruit)
+
+print()
+print('Below is the list of fruits that begin with "P".')
+for fruit in fruit_list:
+    if fruit.startswith('P'):
+        print(fruit)
+
+# End of Series 1
+# ------------------------------------------- #
+
+# ------------------------------------------- #
+# Series 2 - Using the list created in series 1:
+# - Display the list.
+# - Remove the last fruit from the list.
+# - Display the list.
+# - Ask the user for a fruit to delete, find it
+#   and delete it.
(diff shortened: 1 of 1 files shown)


students/tri_nguyen/lesson03/list_lab.py
```diff
@@ -0,0 +1,179 @@
...
+
+        if from_user.lower() == answer_set[1]:
+            print("\nYou don't like " + fruit.lower() + ". It'll be removed.\n")
+            fruit_list.remove(fruit)
+        elif from_user.lower() == answer_set[0]:
+            print('\n' + fruit + ' are good.\n')
+        else:
+            print('\nPlease enter yes or no\n')
+            continue
+        count = count + 1
```
natasha-aleksandrova: Regarding your PR comment about a way to make this cleaner.

------
I would choose a `for` loop for main outer loop because you just need to loop through each fruit and ask the user. A `while` loop is a good option to keep prompting user if they didn't answer with yes or no.

Here's an alternate version of what this could look like:
```
fruit_list = ['Apples', 'Pears', 'Oranges', 'Peaches']

new_fruit_list = []

for fruit in fruit_list:
    like_answer = input(f"Do you like {fruit.lower()}? ")
    
    while like_answer.lower() not in ("yes", "no"):
        like_answer = input("Please answer 'yes' or 'no' ")

    if like_answer == "yes":
        new_fruit_list.append(fruit)

print(new_fruit_list)

```

students/tri_nguyen/lesson03/list_lab.py
```diff
@@ -0,0 +1,179 @@
...
+
+print('--Series 1 of the exercise--'.upper()+'\n')
+
+print('Below is the current list of fruits.')
+for idx, fruit in enumerate(fruit_list):
+    print(idx + 1, '-', fruit)
+
+print()
+user_input = input('Please add another fruit: ')
+fruit_list.append(user_input.title())
```
natasha-aleksandrova: for this exercise it is a good idea to not modify original list since all series/tasks rely on the original content of it.

you could have defined a function for each series and within function scope have the original `fruit_list` or define `fruit_list` at the top of the file and make a copy within each function.

students/tri_nguyen/lesson03/list_lab.py
```diff
@@ -0,0 +1,179 @@
...
+print('Below is a newly updated list of fruits.')
+for idx, fruit in enumerate(fruit_list):
+    print(idx + 1, '-', fruit)
+
+print()
+num_from_user = int(input('Please pick a number to get back the corresponding fruit: '))
+num_from_user = num_from_user - 1  # Python uses zero based indexing
+
+if fruit_list[num_from_user] in fruit_list[num_from_user]:
+    print(num_from_user + 1, '-', fruit_list[num_from_user])
```
natasha-aleksandrova: I'm not sure what the if statement is checking here?

students/tri_nguyen/lesson03/list_lab.py
```diff
@@ -0,0 +1,179 @@
...
+#   and delete it.
+# - (Bonus: Multiply the list times two. Keep asking
+#    until a match is found. Once found, delete all
+#    occurrences).
+
+print('\n'+'--Series 2 of the exercise--'.upper()+'\n')
+
+print('Below is the current list of fruits.')
+for idx, fruit in enumerate(fruit_list):
+    print(idx + 1, '-', fruit)
```
natasha-aleksandrova: your fancy print is a good candidate for a function 🙂 (since it is repeated throughout)

students/tri_nguyen/lesson03/list_lab.py
```diff
@@ -0,0 +1,179 @@
...
+print()
+print('The last fruit ({}) from the list has been removed.'.format(fruit_list[len(fruit_list) - 1]))
+fruit_list.pop()
+for idx, fruit in enumerate(fruit_list):
+    print(idx + 1, '-', fruit)
+
+print()
+to_be_deleted = input('Enter a fruit to delete: ')
+
+if to_be_deleted.title() in fruit_list:
```
natasha-aleksandrova: you can always do shorthand: `input('Enter a fruit to delete: ').title()` so that you don't have to repeat `title()` call

students/tri_nguyen/lesson03/list_lab.py
```diff
@@ -0,0 +1,179 @@
...
+# Series 4 - Using the list from series 1
+# - Make a copy of the list and reverse the letters
+#   in each fruit in the copy.
+# - Delete the last item of the original list.
+# - Display the original and the copy.
+# ------------------------------------------- #
+
+print('\n'+'--Series 4 of the exercise--'.upper()+'\n')
+
+copied_fruit_list = fruit_list.copy()
```
natasha-aleksandrova: even though using `copy()` here works,  warning on copying of objects: 

I would say most of the time you should use deep copy to avoid surprises where your original data structure gets modified. I often use `deepcopy` method because it can apply to lists/dicts:
```
from copy import deepcopy
list_copy = deepcopy(original_list)
```
you can also copy list this way: `original_list[:]`

-----
here is an example of why `copy()` (shallow copy) can do some unexpected things:

```
>>> fruit_list_complex = [['a', 'b'], ['c']]
>>> fruit_list_complex_copy = fruit_list_complex.copy()
>>> fruit_list_complex
[['a', 'b'], ['c']]
>>> fruit_list_complex_copy
[['a', 'b'], ['c']]
>>> fruit_list_complex[1].append('d')
>>> fruit_list_complex
[['a', 'b'], ['c', 'd']]
>>> fruit_list_complex_copy
[['a', 'b'], ['c', 'd']]
>>>
```
----

the top answer [here](https://stackoverflow.com/questions/3975376/understanding-dict-copy-shallow-or-deep) does a great job explaining this in good detail.

tricaonguyen: Wonderful tips and advice. I will clean up the exercises and resubmit.
I was not sure whether I would have to closely follow procedures/instructions from the assignments. I was going to create functions, but the assignment didn't mention so I decided not to. Going forward, I will create functions as much as I can.

natasha-aleksandrova: understandable; you can use functions in pretty much any of the assignments, and certainly not required for early assignment but a great way to organize the code better.

PR #128, opened 2018-02-13
Lesson02 - Fibonacci and Lucas

ghassan24601:


robalford:
@ghassan24601 you're off to a good start here with your fibonacci function. But can you please include all three of the functions as described in the assignment instructions on the course website? 

The first two functions should take only a single input parameter 'n' and return the 'nth' value of the Fibonacci and Lucas series respectively. And the third function will add two optional parameters to define the first two values of the series, defaulting to the Fibonacci series. 

We will approve/merge your pull request, but I would encourage you to submit the full assignment as a new pull request so that we can review in more detail.

Also, please run your code through an online version of the Pep8 linter: http://pep8online.com/. This tool will help you identify and correct conventional Python style violations. Pep8 is the official Python style guide, and following these conventions results in more readable and maintainable Python code.

ghassan24601:
Thanks for the feedback. I have fixed the code or at least, I think I did.
I have also ran it through the Pep8 tool.

robalford:
Looks good! My only suggestion is that the assignment says to return the 'nth' value starting with 0 index, so it should be `fib(0) == 0`, `fib(1) == 1` etc. Also, it's fine to 'hard code' the values from the series in your assertion statements. Here's an example of assertions that you might use for the `fib` function:

```
    assert fib(0) == 0
    assert fib(1) == 1
    assert fib(2) == 1
    assert fib(3) == 2
    assert fib(4) == 3
    assert fib(5) == 5
    assert fib(6) == 8
    assert fib(7) == 13
```

PR #134, opened 2018-02-13
Added series.py - finished.

hortw1991:
Added the finished version of series.py.  I also fixed a type in fizzbuzz that had mislabeled what multiples do what.  Code was unaffected.

### DIFF
--- a/students/thorn/lesson02/series.py
+++ b/students/thorn/lesson02/series.py
@@ -50,12 +50,33 @@ def sum_series(n, first=0, second=1):
         return first
 
 
-# Assertion tests are in the name==main block.
+if __name__ == "__main__":
+    # Test of fibonacci values. 
     print("Fibonacci: ")
-    assert fibonacci(10) ==  55, "correct"
-    
+    # Correct value.
+    try:
+        assert (fibonacci(10) ==  55)
+    except AssertionError:
+        print("Incorrect value.")
+    # Incorrect value.
+    try:
+        assert (fibonacci(8) ==  55)
+    except AssertionError:
+        print("Incorrect value.")
 
+    # Test of lucas values.
+    print("Lucas: ")
+    # Correct values.
+    assert (lucas(7) == 29)
+    assert (lucas(10) == 123)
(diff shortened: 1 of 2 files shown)


students/thorn/lesson02/series.py
```diff
@@ -0,0 +1,82 @@
...
+The following definitions are used to computer the Fibonacci and Lucas Series
+throughout this program.
+    Fibonacci - starts with 0, 1 and sums the previous 2 numbers
+    Lucas - starts with 1, 2 and sums the previous 2 numbers
+'''
+
+def fibonacci(n):
+    """ Returns the nth value in the fibonacci sequence. """
+    if n <= 1:
+        return 0
```
robalford: Remember that the function should return the nth value in the fibonacci series (starting with zero index), so we know that `fibonacci(0) == 0` and `fibonacci(1) == 1` so instead of
```
    if n <= 1:
        return 0
```
this should be 
```
    if n == 0:
        return 0
    if n == 1:
        return 1
```
Other than that, looks good!

students/thorn/lesson02/series.py
```diff
@@ -0,0 +1,82 @@
...
+            n -= 1
+        return first
+
+
+def lucas(n):
+    """ Returns the nth value in the lucas sequence.   Identical to fibonacci
+    except for different starting values. 
+    """
+    if n < 1:
+        return 0
```
robalford: You will need to make a similar revision here for the Lucas series

students/thorn/lesson02/series.py
```diff
@@ -0,0 +1,82 @@
...
+
+def sum_series(n, first=0, second=1):
+    """ sum_series defaults to the fibonacci sequence if only n is assigned a 
+    parameter.  
+        If first and second are given values, they will produce the nth value
+    in a similar fashion to the lucas series beginning with the two optional
+    parameters.
+    """
+    if n < 1:
+        return 0
```
robalford: And here as well, using the `first` and `second` input parameters

PR #135, opened 2018-02-13
Lesson 2 print grid assignment

shibinmathew:
Added readme,  print grid python file, and FizzBuzz file.


students/shibin_mathew/lesson2/FizzBuzz.py
```diff
@@ -0,0 +1,14 @@
+def Fizz(number):
+	for i in range(1,number+1):
+		if (i%3==0) and (i%5==0):
```
natasha-aleksandrova: the parens here are not necessary, can be just: `if i % 3 == 0 and i % 5 == 0`

students/shibin_mathew/lesson2/FizzBuzz.py
```diff
@@ -0,0 +1,14 @@
+def Fizz(number):
```
natasha-aleksandrova: per python style guidelines function names should be lower case snake case, examples:
`def fizz(number)`
`def fizz_buzz(number)`

PR #137, opened 2018-02-14
Dustin_L - Lesson04 - Mailroom 2

rdlaner:
Mailroom 2 assignment in Lesson 04 complete.

Please let me know if you have any suggestions on how I could make this more pythonic. Specifically:
- Is this a reasonable use of defaultdict?
- Line 124, I do a list comprehension without assigning the result to make the code look cleaner. Is this a bad idea and/or is there a better method?

Thanks!


students/Dustin_L/lesson04/mailroom2.py
```diff
@@ -0,0 +1,219 @@
...
+                                              TOTAL_KEY: 0,
+                                              AVE_KEY: 0},
+                        'Angela Davis': {GIFTS_KEY: [74846, 38470, 7570, 50],
+                                         NUM_GIFTS_KEY: 0,
+                                         TOTAL_KEY: 0,
+                                         AVE_KEY: 0},
+                        'Bell Hooks': {GIFTS_KEY: [634547, 47498, 474729, 4567],
+                                       NUM_GIFTS_KEY: 0,
+                                       TOTAL_KEY: 0,
+                                       AVE_KEY: 0}})
```
natasha-aleksandrova: you were definitely on the right track here but I do have couple comments:
* `DONOR_DB` is not a constant (constants don't change), so in this case it should be a lower case global var
* using constants as key name definitions isn't very pythonic although I can see potential benefits
* dictionary keys are _usually_ concise name using camel case or snake case. If you're returning JSON structure it is more common to use camelCase (don't worry if you don't know what JSON is), otherwise it's snake_case.

here is altered version considering my above points:
```
>>> def initialize_donor_data(gifts=None):  # you could add keywords for other data (total etc) but not our use case for now
...     return {"gifts": gifts or [],
...             "num_of_gifts": 0,
...             "total": 0,
...             "avg": 0}
...
>>>
>>> donor_db = defaultdict(lambda: initialize_donor_data(),
...                        {'Toni Morrison': initialize_donor_data(gifts=[1000, 5000, 10000]),
...                         'Mike McHargue': initialize_donor_data(gifts=[12000, 50000, 27000]),
...                         "Flannery O'Connor": initialize_donor_data(gifts=[38734, 6273, 67520]),
...                         'Angela Davis': initialize_donor_data(gifts=[74846, 38470, 7570, 50]),
...                         'Bell Hooks': initialize_donor_data(gifts=[634547, 47498, 474729, 4567])})



```

students/Dustin_L/lesson04/mailroom2.py
```diff
@@ -0,0 +1,219 @@
...
+    names = [donor.lower() for donor in DONOR_DB]
+
+    while True:
+        usr_in = input(name_prompt).strip().lower()
+
+        if usr_in.startswith('q'):
+            break
+        elif usr_in == 'list':
+            print()
+            [print(name.title()) for name in names]
```
natasha-aleksandrova: for your list comprehension question, this is certainly valid but I wouldn't say this is the best practice.

list comprehensions are meant to return a list and so here you are wastefully generating another list and only using it to make it one-liner.

if you would like to keep code cleaner you can always extract a piece of code into a function, but given how simple this part is I do not know that it is worth it.

students/Dustin_L/lesson04/mailroom2.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    select_prompt = ('\nPlease select from the following options:\n'
+                     '\t1. Send a Thank You\n'
+                     '\t2. Create a Report\n'
+                     '\t3. Send letters to all donors\n'
+                     '\t4. quit\n'
+                     ' --> ')
+    usr_in = int(input(select_prompt))
+
+    while usr_in not in (THANK_YOU_OPT, REPORT_OPT, LETTERS_OPT, QUIT_OPT):
```
natasha-aleksandrova: I think we can take it a step further here 😉  let's learn about a cool function `zip`
ref: https://docs.python.org/3.3/library/functions.html#zip

here's a working demonstration:
```
>>> def hello1(): print('hello1')
...
>>> def hello2(): print('hello2')
...
>>> keys = (1, 2)
>>> mapping = dict(zip(keys, (hello1, hello2)))
>>> mapping[1]()
hello1
>>> mapping[2]()
hello2
```

now how can we utilize it in your code?

```
PROMPT_OPTS = (1, 2, 3, 4)
....
while usr_in not in PROMP_OPTS:  # line 70
    print(f"Please try again. Valid options are: {PROMPT_OPTS}")


opt_dict = dict(zip(PROMPT_OPTS, (send_thank_you, create_report, send_letters, quit_mailroom)))  # line 201

....
```

students/Dustin_L/lesson04/mailroom2.py
```diff
@@ -0,0 +1,219 @@
...
+    print('Quitting mailroom...')
+
+
+def send_letters():
+    """Create a letter for each donor and write to disk as a text file"""
+    now = datetime.datetime.today().strftime('%m-%d-%Y')
+
+    for donor, data in DONOR_DB.items():
+        f_name = f'{donor.replace(" ", "_")}_{now}.txt'
+        with open(f_name, 'w') as f:
```
natasha-aleksandrova: 👍  on using `with`

students/Dustin_L/lesson04/mailroom2.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def quit_mailroom():
+    """Exit operations when quitting mail room"""
+    print('Quitting mailroom...')
+
+
+def send_letters():
+    """Create a letter for each donor and write to disk as a text file"""
+    now = datetime.datetime.today().strftime('%m-%d-%Y')
```
natasha-aleksandrova: nice!

students/Dustin_L/lesson04/mailroom2.py
```diff
@@ -0,0 +1,219 @@
...
+    opt_dict = {THANK_YOU_OPT: send_thank_you,
+                REPORT_OPT: create_report,
+                LETTERS_OPT: send_letters,
+                QUIT_OPT: quit_mailroom}
+
+    # Initialize database
+    for values in DONOR_DB.values():
+        values[NUM_GIFTS_KEY] = len(values[GIFTS_KEY])
+        values[TOTAL_KEY] = sum(values[GIFTS_KEY])
+        values[AVE_KEY] = values[TOTAL_KEY] / values[NUM_GIFTS_KEY]
```
natasha-aleksandrova: you could move this logic into my proposed `initialize_donor_data` function

students/Dustin_L/lesson04/mailroom2.py
rdlaner: I like this approach, however, it seems that using the function call in the dict constructor results in the function getting called when the key is selected. When I test ran this, I hit an exception when the add_donation() function was called and it tried to update the database:
File "./mailroom2.py", line 76, in add_donation
    donor_db[donor][GIFTS_KEY].append(amount)
TypeError: 'function' object is not subscriptable

Is there a way to tell python to use the function call only on instantiation of the dict instead of it being invoked with each key access (like is done with opt_dict)?

natasha-aleksandrova: @rdlaner I tried to reproduce your error but everything is working fine:
```
from collections import defaultdict


def initialize_donor_data(gifts=None):  # you could add keywords for other data (total etc) but not our use case for now
     return {"gifts": gifts or [],
             "num_of_gifts": 0,
             "total": 0,
             "avg": 0}


donor_db = defaultdict(lambda: initialize_donor_data(),
                        {'Toni Morrison': initialize_donor_data(gifts=[1000, 5000, 10000]),
                         'Mike McHargue': initialize_donor_data(gifts=[12000, 50000, 27000]),
                         "Flannery O'Connor": initialize_donor_data(gifts=[38734, 6273, 67520]),
                         'Angela Davis': initialize_donor_data(gifts=[74846, 38470, 7570, 50]),
                         'Bell Hooks': initialize_donor_data(gifts=[634547, 47498, 474729, 4567])})

print(donor_db)


donor = "Bell Hooks"
amount = 10
donor_db[donor]["gifts"].append(amount)

donor_db["Natasha"]["gifts"].append(amount)

print(donor_db)
```

if you provide the code that gave you this error I can help you debug, doesn't have to be a PR - you can just push your code to your fork and let me know so I can run it.

rdlaner: Hmmm. Ok, I pushed up my latest changes to my fork. Let me know what you think. I run into the exception after I attempt to send a thank you to a person not in the database.

rdlaner: I can confirm that it works fine when accessing a member that is already in the dictionary, but when trying to add a key that isn't already in the dictionary, it seems that defaultdict is having an issue with the init_donor_data() function.

rdlaner: Ah! I think I found my error. I had left off the parenthesis for the init_donor_data function when called from the lambda on line 41.

natasha-aleksandrova: yay! glad you got it figured out 🙂

PR #139, opened 2018-02-15
Adding the Slicing Lab Assignment

wesmoy17:


students/WesleyM/Lesson3/SlicingLab.py
```diff
@@ -0,0 +1,49 @@
+def floor(n):
```
natasha-aleksandrova: it is preferred to have modules be `snake_case`, => `slicing_lab.py`

students/WesleyM/Lesson3/SlicingLab.py
```diff
@@ -0,0 +1,49 @@
...
+    return a_new_sequence
+
+def thirds(seq):
+    """Divides an input string or tuple into thirds, and arranges the first third to the end"""
+    seq1 = seq[0:floor(len(seq)/3)]
+    seq2 = seq[floor(len(seq)/3):]
+    a_new_sequence = seq2 + seq1
+    return a_new_sequence
+
+if __name__ != "__main__":
```
natasha-aleksandrova: this should be `if __name__ == "__main__":` , your tests didn't actually run, if you fix you will see that some tests are failing

students/WesleyM/Lesson3/SlicingLab.py
```diff
@@ -0,0 +1,49 @@
...
+    res = int(n)
+    return res if res == n or n >= 0 else res-1
+    
+def exchange_first_last(seq):
+    """Reverses the first and last letter or number in a tuple or string"""
+    seqfirst = seq[0:0]
+    seqlast = seq[-1:]
+    remainder = seq[1:-1]
+    a_new_sequence = seqlast + remainder + seqfirst
+    return a_new_sequence
```
natasha-aleksandrova: there is a bug in `seqfirst`, it isn't actually returning the first element.

this simply could be
```
```

PR #142, opened 2018-02-16
submitting strformat.py

tricaonguyen:
I am not sure if I did task six extra correctly. Please let me know if it is wrong.

### DIFF
--- a/students/tri_nguyen/lesson03/strformat_lab.py
+++ b/students/tri_nguyen/lesson03/strformat_lab.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python3
+
+# ------------------------------------------- #
+# Student:   Tri Nguyen
+# Instructor: Natasha Aleksandrova
+# Date:  10-Feb-2018
+# ------------------------------------------- #
+
+# Define functions
+
+
+def format_presentation(a_string):
+    ''' this function prints stuff for readability '''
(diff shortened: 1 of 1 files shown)


students/tri_nguyen/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def task_one():
+    ''' this function will format a 4 element
+        tuple and produce
+        'file_002 :   123.46, 1.00e+04, 1.23e+04' '''
+
+    a_tup = (2, 123.4567, 10000, 12345.67)
+
+    file, floating, exp_one, exp_two = a_tup
```
natasha-aleksandrova: be careful about overwriting built-in functions ([file](https://docs.python.org/2/library/functions.html#file)), some editors will have special syntax highlighting.. but you can also see it in github highlighted orange

natasha-aleksandrova:
as far as your question about task six, consider variable length for your middle column.. if I make one of the values in the rows slightly longer this is what it looks like:
```
Task six returns the formatted string below.
Makiko     33889898498 $1234.23
Yumika     5     $12.00
Kohana     1     $890023.01
Tri        34    $9999999.01```


you're really close 🙂 just need to handle variable length data in the rows

PR #143, opened 2018-02-17
Fibonacci Series Assignment

shibinmathew:
Created a series.py file which contains a function to compute the fibonacci sequence, lucas sequence and a generalize function to compute both sequences.

### DIFF
--- a/students/shibin_mathew/lesson2/series.py
+++ b/students/shibin_mathew/lesson2/series.py
@@ -1,10 +1,10 @@
 def fibonacci(n):
+	'''This function returns the nth value of the fibonacci seires'''
 	first = 0
 	second = 0
 	third = 0
 	listT= []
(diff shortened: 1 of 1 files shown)


students/shibin_mathew/lesson2/series.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+
+def sum_series(n, firstPar = 0, secondPar = 1):
+	#Generatlized function that implements the logic behind the fibonacci sequence and the lucas sequence
+	first = 0 #initial value
+	second = 0 #initial value
+	third = 0 #initial value
+	listS = []
+	for i in range(0,100): #default parameter set to 100
```
natasha-aleksandrova: let's say our max is higher (4mil), how would you solve it without having to generate full sequence only to get let's say 5th element?

PR #146, opened 2018-02-17
added strformat_lab.py with tasks 01 through 05

johnpharmd:


students/johnpharmd/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,72 @@
...
+
+# started with form_st = 'file_{:d}: {:f}, {:d}, {:f}'
+# Internet search pointed me to https://docs.python.org/2/library/string.html,
+# which, apparently, is not (completely) deprecated in Python 3.
+# Also used https://docs.python.org/3.3/library/string.html#formatstrings.
+# iteration: form_st = 'file_{:03d}: {:f}, {:e}, {:e}'
+# iteration: form_st = 'file_{:03d}: {:.2f}, {:e}, {:e}'
+# iteration: form_st = 'file_{:03d}: {:.2f}, {:.2e}, {:e}'
+# Task 1
+ form_st = 'file_{:03d}: {:.2f}, {:.2e}, {:.2e}'
```
robalford: Remove the single space indentation through this section. This will throw an IndentationError and cause the program to crash.

Your string formatting operators look good but you need to call the format method on these strings as you did in the functions below. Adding a print statement is a nice way to test that your code works as expected:

```
numbers = (2, 123.4567, 10000, 12345.67)
form_st = 'file_{:03d}: {:.2f}, {:.2e}, {:.2e}'.format(*numbers)
print(form_st)
```

students/johnpharmd/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,72 @@
...
+
+ # Task 2
+ # started with:
+ # form_st2 = 'file_{integer}: {flo}, {expo1}, {expo2}'
+ # Found this:
+ # https://stackoverflow.com/questions/3228865/how-do-i-format-a-number-with-a-variable-number-of-digits-in-python
+ form_st1 = 'file_{num:0{width}}: {flo:.2f}, {expo1:.2e}, {expo2:.2e}'
+
```
robalford: Call format method on this string as well. Not sure if you've gotten to this in the course materials yet, but you can pass a dict to the format method in the same way that you do below with a tuple, so this can be written like this:

```
numbers = {
    'num': 2,
    'flo': 123.4567,
    'expo1': 10000,
    'expo2': 12345.67,
}

form_st1 = 'file_{num:03d}: {flo:.2f}, {expo1:.2e}, {expo2:.2e}'.format(**numbers)
```

students/johnpharmd/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,72 @@
...
+
+# Task 3
+# Rewrite 'the 3 numbers are: {:d}, {:d}, {:d}".format(1,2,3)' to
+# take an arbitrary number of values.
+
+# form_string = 'the {} numbers are:...'
+# see Chris B's video at 5:10 in
+def formatter(in_tuple):
+    l = len(in_tuple)
```
robalford: Since the variable name `l` doesn't add any meaning here, you could just call `len()` on the `in_tuple` where you need it.

students/johnpharmd/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,72 @@
...
+    output0 = 'the {} numbers are: '
+    output1 = ', '.join(['{}'] * l)
```
robalford: You could use in place assignment here so you don't need the second `output` variable:

`output += ', '.join(['{}'] * len(in_tuple))`

students/johnpharmd/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,72 @@
...
+    (4, 30, 2017, 2, 27)
+
+    use string formating to print:
+
+    '02 27 2017 04 30'
+
+Hint: use index numbers to specify positions.
+"""
+def task4(in_tuple):
+    i_t = in_tuple
```
robalford: No need to reassign `in_tuple` here, you can just use `in_tuple` in place of `i_t` in your function block. In general, it's a best practice for readability in Python (and programming in general) to use meaningful variable names for anything other than short-lived 'throw away' variables like counter variables inside a loop.

PR #150, opened 2018-02-18
lesson03 - thirds

ghassan24601:
I'm not entirely sure what middle third, last third and first third mean. I solved based on what I "think" they mean :)

### DIFF
--- a/students/ghassan/lesson03/fruits.py
+++ b/students/ghassan/lesson03/fruits.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+
+# series1
+
+fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+print(fruits)
+new_fruit = input('Enter a new fruit: ')
+fruits.append(new_fruit)
+print(fruits)
+fruit_index = input('Enter a fruit index: ')
+print(fruits[int(fruit_index)-1])
+another_fruit = input('Enter another fruit: ')
+fruits = [another_fruit] + fruits
+print(fruits)
+yet_another_fruit = input('Enter yet another fruit: ')
+fruits.insert(0, yet_another_fruit)
+print(fruits)
+print([fruit for fruit in fruits if fruit.startswith('P')])
+
+# series 2
+
+fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+print(fruits)
+fruits.pop()
+print(fruits)
+del_fruit = input('Enter a fruit to delete: ')
+fruits.remove(del_fruit)
+
+# series 3
+
+fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+for fruit in fruits:
+    qq = input('Do you like {}?'.format(fruit.lower()))
+    while qq.lower() != 'yes' and qq.lower() != 'no':
+        qq = input('Enter either "yes" or "no"\n')
+    if qq.lower() == 'no':
+        fruits.remove(fruit)
+
+print(fruits)
+
+# series 4
+
+fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+fruitsy = []
+for fruit in fruits:
+    fruitsy.append(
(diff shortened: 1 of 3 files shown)


students/ghassan/lesson03/fruits.py
```diff
@@ -0,0 +1,50 @@
...
+print(fruits)
+fruit_index = input('Enter a fruit index: ')
+print(fruits[int(fruit_index)-1])
+another_fruit = input('Enter another fruit: ')
+fruits = [another_fruit] + fruits
+print(fruits)
+yet_another_fruit = input('Enter yet another fruit: ')
+fruits.insert(0, yet_another_fruit)
+print(fruits)
+print([fruit for fruit in fruits if fruit.startswith('P')])
```
natasha-aleksandrova: nice use of comprehension here 👍 

always a good idea to use `lower()` on a string when doing comparison to be case not sensitive

students/ghassan/lesson03/fruits.py
```diff
@@ -0,0 +1,50 @@
...
+fruits.pop()
+print(fruits)
+del_fruit = input('Enter a fruit to delete: ')
+fruits.remove(del_fruit)
+
+# series 3
+
+fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+for fruit in fruits:
+    qq = input('Do you like {}?'.format(fruit.lower()))
```
natasha-aleksandrova: to avoid having to call `lower()` 3 times below you can just have it on line 33:

`qq = input('Do you like {}?'.format(fruit.lower())).lower()`

students/ghassan/lesson03/slicing.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
+
+
+def first_last(data):
+    new_list = data[:]
+    new_list[0] = data[-1]
+    new_list[-1] = data[0]
+    return new_list
```
natasha-aleksandrova: another way: `return data[-1:] + data[1:-1] + data[:1]`

students/ghassan/lesson03/slicing.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def every_other(data):
+    new_list = data[::2]
+    return new_list
```
natasha-aleksandrova: no need for additional variable, it can be just: `return data[::2]`

students/ghassan/lesson03/slicing.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def fourth_fourth(data):
+    new_list = data[0:4] + data[-4:]
+    return new_list
+
+
+def reversed(data):
```
natasha-aleksandrova: be careful here.. you are overwriting built-in function `reversed` (notice github highlight it different color, so should your editor)

students/ghassan/lesson03/slicing.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def every_other(data):
+    new_list = data[::2]
+    return new_list
+
+
+def fourth_fourth(data):
+    new_list = data[0:4] + data[-4:]
+    return new_list
```
natasha-aleksandrova: this should also do: `, and then every other item in between.`

students/ghassan/lesson03/slicing.py
```diff
@@ -0,0 +1,44 @@
...
+
+def reversed(data):
+    return data[::-1]
+
+
+def thirds(data):
+    first_third = int(len(data) / 3)
+    middle_third = int((len(data) / 3) * 2)
+    last_third = data[-1]
+    return [data[middle_third], last_third, first_third]
```
natasha-aleksandrova: so given input `data = (2, 54, 13, 12, 5, 32)`, your expected data is: `(13, 12, 5, 32, 2, 54)`

you would divide your input into 3 parts, then return middle part + last part + first part

PR #151, opened 2018-02-18
Lesson03 - MailRoom

ghassan24601:


students/ghassan/lesson03/mailroom.py
```diff
@@ -0,0 +1,60 @@
...
+    ('James', (343, 22.11)),
+    ('Ford', (3000, 400))]
+
+
+def prompt():
+    choice = input(
+        'Pick one:\n1) Send a thank you\n2) Create a Report\n3) Quit\n')
+    while choice not in ['1', '2', '3']:
+        choice = input(
+            'Pick one:\n1) Send a thank you\n2) Create a Report\n 3) Quit\n')
```
natasha-aleksandrova: it is a good idea to avoid duplicating code, one option is to initialize `choice`, so it would look like:
```
choice = None
while choice not in ['1', '2', '3']:
    choice = input("...")
return choice
```

Also, it is a good idea to choose function names that hint at what it is doing.. for example, `prompt` would instantly tell me an action is being done, but there is not a hint that it will return something 🙂 

something like `get_user_selection` would indicate that we are getting what user selected (`get_` part is the hint)

students/ghassan/lesson03/mailroom.py
```diff
@@ -0,0 +1,60 @@
...
+    return choice
+
+
+def send_thankyou():
+    donor_name = input('Enter the donor\'s name: ')
+    if donor_name == 'list':
+        for donor in donors:
+            print(donor[0])
+    elif donor_name not in donors:
+        donors.append((donor_name, ))
```
natasha-aleksandrova: your mailroom program should also have functionality to add donor amount to existing donors (which I don't see here).

Since that functionality is missing you can't really see that tuple was the wrong choice for the data structure since it is immutable. If you tried to add another donor amount to existing donor you would have gotten this error: `AttributeError: 'tuple' object has no attribute 'append'`

students/ghassan/lesson03/mailroom.py
```diff
@@ -0,0 +1,60 @@
...
+    prompt()
+
+
+def create_report():
+    print('{:20} | {:15} | {:10} | {:15}'.format(
+        'Donor Name', 'Total Given', 'Num Gifts', 'Average Gift'))
+    print('-'*70)
+    for donor in donors:
+        print('{:20} | {:15} | {:10} | {:15}'.format(
+            donor[0], sum(int(x) for x in donor[1]),
```
natasha-aleksandrova: you should ensure you are only storing ints or floats to have "clean" data then you shouldn't need the extra loop to cast to int

PR #152, opened 2018-02-18
Lesson 03 - Emmanuel Twum - String Formatting

etwum:
string formatting lab complete


students/etwum/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,66 @@
...
+string = f"file{x1:03d}: {x2:.2f}, {x3:.2e}, {x4:.2e}"
+print(string)
+print("----------------------------------------------------------------")
+
+# Task Three
+
+def formatter(*tuple):
+    # creates a string which dynamically prints a string based on the numbers passed to the function
+    form_string = "the " + str(len(tuple)) + " numbers are " + ", ".join(["{}"]*len(tuple)).format(*tuple)
+    return print(form_string)
```
natasha-aleksandrova: `print()` function doesn't actually return anything, so either you should have `return = "the " ... ` or remove `return`

PR #153, opened 2018-02-18
lesson03: added slicing_lab and list_lab

22e88:


students/elmar_m/lesson03/list_lab.py
```diff
@@ -0,0 +1,123 @@
...
+'''
+
+def print_list(l):
+    '''
+    prints the elements of a list in one line, ending with a 
+    single line feed
+    '''
+    for i in l:
+        print(i, '', end='')     
+    print()
```
natasha-aleksandrova: I like the utility print function 👍 

Here is alternate as a one liner: `print(' '.join(map(str, l)))` (note that `map` isn't necessary in your case since your fruits are all strings.

students/elmar_m/lesson03/list_lab.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def series3():
+    '''
+    series 3
+    '''
+    fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+    print('original list:')
+    print_list(fruits)
+    for i in fruits[:]:
```
natasha-aleksandrova: nice on making list copy on the fly 👍

students/elmar_m/lesson03/list_lab.py
```diff
@@ -0,0 +1,123 @@
...
+        print('Do you like', i, '?', 'Type "yes" or "no": >')
+        answer = input()
+        while (answer != 'yes') and (answer != 'no'):
+            answer = input('Please type either "yes" or "no": >\n')
+        if (answer == 'no'):
```
natasha-aleksandrova: no need for parens around answer

students/elmar_m/lesson03/list_lab.py
```diff
@@ -0,0 +1,123 @@
...
+    '''
+    series 3
+    '''
+    fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+    print('original list:')
+    print_list(fruits)
+    for i in fruits[:]:
+        print('Do you like', i, '?', 'Type "yes" or "no": >')
+        answer = input()
+        while (answer != 'yes') and (answer != 'no'):
```
natasha-aleksandrova: as a shorthand you can also do: `answer not in ('yes', 'no')`

students/elmar_m/lesson03/list_lab.py
```diff
@@ -0,0 +1,123 @@
...
+    print('original list items with reversed letters:')    
+    print_list(cfruits)
+
+    print('original list, last item removed:')
+    fruits.pop()
+    print_list(fruits)
+
+
+if __name__ == '__main__':
+    print('This file contains functions for the list lab exercises, please import it to use them')
```
natasha-aleksandrova: list lab looks great!

students/elmar_m/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,67 @@
...
+    '''
+    return a given sequence with the first and last item exchanged
+    '''
+    return a[-1:] + a[1:-1] + a[:1]
+
+def onlyfl(a):
+    '''
+    return only the first and last item, remove everything in between 
+    '''
+    return a[:1] + a[-1:]
```
natasha-aleksandrova: this should be: "with every other item removed." so expected on `a_tuple` is : `(2, 13, 5)`

students/elmar_m/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,67 @@
...
+
+def withoutfl4(a):
+    '''
+    remove the first 4 and the last 4 items of a given sequence and 
+    return everything in between
+    '''
+    return a[4:-4]
```
natasha-aleksandrova: this should be: "with the first 4 and the last 4 items removed, and then every other item in between."

so your `al_tuple` expected is: `(5, 48)`

PR #154, opened 2018-02-18
Lesson 3 pull request

kuhnbt7:


students/kuhnbt/lesson3/strfformat_lab.py
```diff
@@ -0,0 +1,54 @@
...
+
+# Task1
+
+def format_filenames(s):
+    """Make properly formatted filename given 4-element tuple"""
+    element1 = 'file_{:0>3d}: '.format(s[0])
+    element2 = '{:.2f}'.format(s[1])
+    element3 = '{:.2e}'.format(s[2])
+    element4 = '{:.2e}'.format(s[3])
+    return element1 + ', '.join([element2, element3, element4])
```
robalford: Looks good. Could be rewritten as a one liner as follows:

```
def format_filenames(s):
    """Make properly formatted filename given 4-element tuple"""
    return "file_{:03d} : {:.2f}, '{:.2e}', {:.2e}".format(*s)
```

robalford:
@kuhnbt7 Looks great! You've broken up your code into nice small functions, and satisfied all the requirements for these assignments. The only suggestion I have is to run your code through an online Pep8 linter (Python style guide), or configure your text editor to catch Pep8 violations. Mailroom in particular could use some Pep8 love.

I did see your submission for the String Formatting Lab on the course website, but don't see the others. Can you try resubmitting before we merge this PR?

natasha-aleksandrova:
@kuhnbt7 on the course website under each assignment is the submission area. So for slice lab there is submission at the bottom, then on the next page for string format lab there is submission at the bottom, same for mailroom. It sounds like you tried to submit multiple files within the mailroom assignment page.

Do go back to the course website and make submissions under slice and str format labs so that the grade can be recorded, let us know if you have any further questions 🙂 

Merging this PR for now.

kuhnbt7:
Thanks for the feedback Rob and Natasha. I have been using pycodestyle to correct my code but I think I got lazy when it came to the mailroom assignment. Sorry for the screwup with website submissions; I'll try not to repeat that and will go submit the other two assignments now. On a different subject, I took the midcourse survey this afternoon but didn't include a suggestion that I thought of later, so wanted to send it to you guys. I'm taking another course at a different university and they give us 'showcase solutions' after assignments are submitted, which are just solutions that the instructors write. Even when I complete the assignments successfully, it's helpful to look at their code and identify areas where I could have structured things better or more concisely - kind of like Rob's comment above. Anyway, just a suggestion :)

natasha-aleksandrova:
@kuhnbt7 no worries, easy to fix 🙂 

thanks for the suggestion, I agree it would be helpful for students to see the solutions. I will bring it up with the team next week; we would need to figure out how those would be distributed for this self paced format.

PR #155, opened 2018-02-18
completed trigram

rmart300:


students/rmart300/trigram/trigram.py
```diff
@@ -1,45 +1,60 @@
...
+                    if digram in trigram_dict:
+                        trigram_dict[digram].append(words[i+1])
+                    else:
+                        trigram_dict[digram] = [words[i+1]]
 
 def write_book():
+    """write book in trigram syntax"""
     trigram_usage_dict = {}
     for digram in trigram_dict:
         trigram_usage_dict[digram] = 0
```
natasha-aleksandrova: I think something like Counter would be cool to use here:
https://docs.python.org/2/library/collections.html#collections.Counter

here is a demonstration:
```
c = Counter(dict.fromkeys(trigram_dict.keys(), 0))

# to increment
c[digram] += 1

```

what you have is totally valid, just exploring other features of dict 🙂

students/rmart300/trigram/trigram.py
```diff
@@ -1,45 +1,60 @@
...
     
     out = open('my_book.txt','w')
```
natasha-aleksandrova: should use `with` statement here as well

students/rmart300/trigram/trigram.py
```diff
@@ -1,45 +1,60 @@
...
+            words = line.split()
+
+            for i in range(len(words)):
+                if i > 1:
+                    digram = f"{words[i-1]} {words[i]}"
+                if i < len(words)-1:
+                    if digram in trigram_dict:
+                        trigram_dict[digram].append(words[i+1])
+                    else:
+                        trigram_dict[digram] = [words[i+1]]
```
natasha-aleksandrova: good use case for `trigram_dict` to be a defaultdict: `trigram_dict = defaultdict(list)`

removes the need for special logic when adding items to the list

PR #159, opened 2018-02-19
Submitting Listlab.py and correcting SlicingLab.py

wesmoy17:
While submitting Listlab.py Series 3, I noticed that when I tried deleting from my list in my for loop, it would skip the next element. As a result, I used a reverse function to start from the end of my list going backward, but I was curious if there is a way to use the .remove function while going forward?

### DIFF
--- a/students/WesleyM/Lesson3/ListLab.py
+++ b/students/WesleyM/Lesson3/ListLab.py
@@ -1,0 +1,79 @@
+def fruit1():
+    """
+    List four fruits, includes a user input to list,
+    and removes fruit that start with the letter "P"
+    """
+    fruit = ["Apples", "Pears", "Oranges", "Peaches"]
+
+    print(fruit)
+    
+    response = input("Input a fruit > ")
+
+    fruit.append(response)
+
+    print(fruit)
+  
+    numres = int(input("Input a number > "))
+    while numres < 1 or numres > 5:
+        print('Input a number between 1 and 5')
+        numres = int(input("Input a number > "))
+    print(fruit[numres-1])
+    response = input("Input another fruit > ")
+    fruit = [response] + fruit
+    print(fruit)
+    response = input("Input another fruit > ")
+    fruit.insert(0, response)
+
+    for i in fruit:
+        if i[0] == ('P'):
+            print(i, end = ' ')
+    print()
+    return fruit
+
+def fruit2(fruit):
+    """
+    Displays fruit list, removes the last fruit,
+    and deletes a user input fruit
+    """
+    print(fruit)
+    fruit = fruit[:-1]
+    print(fruit)
+    response = input("Delete a fruit > ")
+    while response not in fruit:
+        print("Error. Fruit not included. Doubling list.")
+        fruit *= 2
+        print(fruit)
+        response = input("Delete a fruit > ")
+    while response in fruit:
+        fruit.remove(response)
+    print(fruit)
+
+def fruit3(fruit):
+    """Asks user if they like fruit and deletes disliked fruit"""
+    fruit_like = fruit
+    print(fruit_like)
+    for i in reversed(fruit_like):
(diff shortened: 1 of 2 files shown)


students/WesleyM/Lesson3/ListLab.py
```diff
@@ -0,0 +1,79 @@
...
+    print(fruit)
+    
+    response = input("Input a fruit > ")
+
+    fruit.append(response)
+
+    print(fruit)
+  
+    numres = int(input("Input a number > "))
+    while numres < 1 or numres > 5:
```
robalford: How would you handle other forms of bad input here? e.g. `'three'`

students/WesleyM/Lesson3/ListLab.py
```diff
@@ -0,0 +1,79 @@
...
+        print('Input a number between 1 and 5')
+        numres = int(input("Input a number > "))
+    print(fruit[numres-1])
+    response = input("Input another fruit > ")
+    fruit = [response] + fruit
+    print(fruit)
+    response = input("Input another fruit > ")
+    fruit.insert(0, response)
+
+    for i in fruit:
```
robalford: Since `i` is often used as shorthand for `index` in an enumerated loop, I would use a different variable name here e.g. `for fruit in fruits`

students/WesleyM/Lesson3/ListLab.py
```diff
@@ -0,0 +1,79 @@
...
+        if i[0] == ('P'):
```
robalford: You could use the `lower()` string method here to catch upper and lower case 'p's: `if i[0].lower() == 'p':`. Also no need for the parentheses around the `'P'`

students/WesleyM/Lesson3/ListLab.py
```diff
@@ -0,0 +1,79 @@
...
+    print()
+    return fruit
+
+def fruit2(fruit):
+    """
+    Displays fruit list, removes the last fruit,
+    and deletes a user input fruit
+    """
+    print(fruit)
+    fruit = fruit[:-1]
```
robalford: You can just call `fruit.pop()` here, which is a built-in list method that removes and returns (if you need it) the last item in the list.

students/WesleyM/Lesson3/ListLab.py
```diff
@@ -0,0 +1,79 @@
...
+        response = input("Delete a fruit > ")
+    while response in fruit:
+        fruit.remove(response)
+    print(fruit)
+
+def fruit3(fruit):
+    """Asks user if they like fruit and deletes disliked fruit"""
+    fruit_like = fruit
+    print(fruit_like)
+    for i in reversed(fruit_like):
```
robalford: The simplest way to approach this is probably to loop over a copy of the list by using the slicing operator. So it could just be:

`for i in fruit[:]:`

The `reversed()` method works because it creates a fancy Python object called an `iterator` that allows you to loop through each item in the list in reversed order. But you don't need the list to be reversed here so I'd go with a different option.

In regard to your comment on the PR, the important thing is that we don't attempt to alter the original list while looping through it, because every time we remove an item, the list indexes change in relation to our position in the loop. 

Here's a well written article that explains why we need to loop over a copy in order to alter a list in this way: https://unspecified.wordpress.com/2009/02/12/thou-shalt-not-modify-a-list-during-iteration/

students/WesleyM/Lesson3/ListLab.py
```diff
@@ -0,0 +1,79 @@
...
+    print(fruit_like)
+
+def fruit4(fruit):
+    """
+    Makes a copy of the fruit list with the reverse of the letters,
+    deletes the last item from original list, and displays both lists
+    """
+    fruit_copy = []
+    for i in fruit:
+        fruit_copy += [i[::-1]]
```
robalford: This would be a good use case for the `list.append()` method

PR #160, opened 2018-02-19
submitting mailroom.py

tricaonguyen:
Hi Natasha,

For this exercise, I did not know a good way to sort values within nested lists. As a result, of course, I googled and found a way. You will see it on line #84 and #98. I felt that I cheated because the exercise said: "Finally, use only functions and the basic Python data types you’ve learned about so far. There is no need to go any farther than that for this assignment."

Maybe, you can show me how to sort without "itemgetter".

In addition, I tried:
**full_name = input('Enter a Full Name (or type "list" to get the list of donors): ').title()**
for some reason, it didn't take. So had to do:
**full_name.title()** in each if block.

Thanks,

Tri

### DIFF
--- a/students/tri_nguyen/lesson03/mailroom.py
+++ b/students/tri_nguyen/lesson03/mailroom.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+
+# ------------------------------------------- #
+# Student:   Tri Nguyen
+# Instructor: Natasha Aleksandrova
+# Date:  15-Feb-2018
+# ------------------------------------------- #
+
+# Declare variables and data structure
+
+donor_list = [
+    ['William Gates, III', 653784.49, 2, 326892.24],
+    ['Mark Zuckerberg', 16396.10, 3, 5465.37],
+    ['Jeff Bezos', 877.33, 1, 877.33],
+    ['Paul Allen', 708.42, 3, 236.14],
+    ['Tri Nguyen', 100.00, 1, 100.00]
+    ]
+
+
+# Declare functions
+
+
+def display_menu():
+    ''' this function display a menu
+        of 3 actions '''
+    ACTIONS = ('Send a Thank You', 'Create a Report', 'Quit')
+
+    print(''' Select one of the actions below:
+
+    1. {}
+    2. {}
+    3. {}
+    '''.format(*ACTIONS))
+
+
+def send_thanks():
+    ''' this function composes an email thanking
+        the donor for their generous donation '''
+
+    while True:
+
+        full_name = input('Enter a Full Name (or type "list" to get the list of donors): ')
+
+        if full_name.lower() == 'list':
+            for idx, item in enumerate(donor_list):
+                print(idx + 1, '-', donor_list[idx][0])
+            continue
+
+        elif full_name.title() not in [donor_list[idx][0] for idx, item in enumerate(donor_list)]:
+            donation_amount = float(input('Enter donation amount: '))
+            donor_list.append([full_name.title(), donation_amount, 1, donation_amount])
+
+        elif full_name.title() in [donor_list[idx][0] for idx, item in enumerate(donor_list)]:
+            donation_amount = float(input('Enter donation amount: '))
+            for idx, item in enumerate(donor_list):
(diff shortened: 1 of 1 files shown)


students/tri_nguyen/lesson03/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+        print()
+
+        if action == '1':
+            send_thanks()
+
+        elif action == '2':
+            create_report()
+
+        elif action.lower() == '3':
+            break
```
natasha-aleksandrova: very clean while loop with redirection to functions.

students/tri_nguyen/lesson03/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+    '''.format(*ACTIONS))
+
+
+def send_thanks():
+    ''' this function composes an email thanking
+        the donor for their generous donation '''
+
+    while True:
+
+        full_name = input('Enter a Full Name (or type "list" to get the list of donors): ')
```
natasha-aleksandrova: the inline `title()` call should have worked...

from this code:
```
    full_name = input('Enter a Full Name (or type "list" to get the list of donors): ').title()
    print(full_name)
```
is the output
```
>> Enter a Full Name (or type "list" to get the list of donors): natasha
Natasha
```

I suspect it didn't work from some other bug down the line?

students/tri_nguyen/lesson03/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if full_name.lower() == 'list':
+            for idx, item in enumerate(donor_list):
+                print(idx + 1, '-', donor_list[idx][0])
+            continue
+
+        elif full_name.title() not in [donor_list[idx][0] for idx, item in enumerate(donor_list)]:
```
natasha-aleksandrova: instead have this:
`elif full_name not in [donor[0] for donor in donor_list]`

`enumerate` is not necessary since you only need the items in the list

students/tri_nguyen/lesson03/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+            donation_amount = float(input('Enter donation amount: '))
+            donor_list.append([full_name.title(), donation_amount, 1, donation_amount])
+
+        elif full_name.title() in [donor_list[idx][0] for idx, item in enumerate(donor_list)]:
```
natasha-aleksandrova: you can extract this bit into a function to avoid duplication like so (which also adds clarity):
```
def get_donor_names():
    return [donor[0] for donor in donor_list]
```
and then you can simply call: `elif full_name in get_donor_names():`

students/tri_nguyen/lesson03/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+            donation_amount = float(input('Enter donation amount: '))
+            for idx, item in enumerate(donor_list):
```
natasha-aleksandrova: same here, don't really need `enumerate`. simply: `for donor in donor_list` will give you the donor item, then instead of `donor_list[idx][1]` => `donor[1]` (see how it simplifies code?)

students/tri_nguyen/lesson03/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+            email_template = '''
+                Dear {0},
+
+                Thank you for giving us ${1:,.2f}. Your money will be put to good use.
+                Please come back and donate more.
+
+                Best Regards,
+
+                John Doe
+            '''
```
natasha-aleksandrova: 👍 on using triple quote strings for this

students/tri_nguyen/lesson03/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+
+    print(formatted_columns)
+
+    print('{}'.format('-' * len(formatted_columns)))
+
+    print('Report of Donors is below:\n')
+
+    donor_list.sort(key=itemgetter(1), reverse=True)
+
+    for idx, item in enumerate(donor_list):
```
natasha-aleksandrova: same comment about `enumerate` here

PR #161, opened 2018-02-19
Lesson #3 Assignment Submission

csdotson:
Submitting completed lesson #3 assignments. Submission includes following files:

1) README
2) slicing_lab.py
3) list_lab.py
4) strformat.py
5) mailroom.py

### DIFF
--- a/students/csdotson/lesson03/mailroom.py
+++ b/students/csdotson/lesson03/mailroom.py
@@ -1,55 +1,98 @@
 #!/usr/bin/env python3
 # Lesson 3 - Mailroom... part 1
 
-Donations = [ ["Bill Gates", 10.50], ["Bill Gates", 123.45], ["Bill Gates",    1111.11], ["Jeff Bezos", 7.65], ["Jeff Bezos", 1000], ["Paul Allen", 145.90], ["John Nordstrom", 45.67], ["John Nordstrom", 6519.65], ["Mark Zuck", 789.12] ]
-
-### Prompts and Flow Control ###
-response = input("Welcome Chris! What would you like to do?\n(S)end a Thank You / (C)reate a Report / (Q)uit\n--> ")
-
-while True:
-    if response == "S":
-        send_thank_you()
-        break
-    elif response == "C":
-        print_report()
-        break
-    else:
-        print("Goodbye Sir!")
-        break
+donations = [ ["Bill Gates", 10.50], ["Bill Gates", 123.45], ["Bill Gates",    1111.11], ["Jeff Bezos", 7.65], ["Jeff Bezos", 1000], ["Paul Allen", 145.90], ["John Nordstrom", 45.67], ["John Nordstrom", 6519.65], ["Mark Zuck", 789.12] ]
 
 
 ### Function Definitions ###
+def prompt(which = 0):
+    if which == 0:
+        response = input("Welcome Chris! What would you like to do?\n(S)end a Thank You / (C)reate a Report / (Q)uit\n--> ")
+    else:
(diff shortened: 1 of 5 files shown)


students/csdotson/lesson03/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+def send_thank_you():
+    # Add donor/donations and trigger thank you email
+    response = input("Please enter full name of person you'd like to thank:\nBy the way, you can also type 'list' to get a list of donors or 'P' to return to the main prompt\n--> ")
+
+    if response == "list":
+        print()
+        list_donors()
+        send_thank_you()
+    elif response == "P":
+        prompt(1)
```
natasha-aleksandrova: magic numbers 😉 (https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)

you could define constants for cases of `which` var -> `DFLT_PROMPT = 0, NEXT_PROMPT = 1`

students/csdotson/lesson03/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+    donor_list = donor_names()
+    print("List of donors:")
+    for i in donor_list:
+        print(i)
+    print()
+
+def donor_names():
+    # Create unique list of donor names
+    name_list = []
+    for i, j in donations:
```
natasha-aleksandrova: although one letter vars are ok in comprehensions (and some other cases), it is better to use something more meaningful in a for loop like this, for example: `for donor_name, donor_data in donations`

however in your case since you're not using `j` it is more clear to have:
```
for donation in donations:
    name_list.append(donation[0])
```

students/csdotson/lesson03/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+    return name_list
+
+def print_report():
+    # Print formatted report showing donors and donation stats
+    header = '{:20}|{:^15}|{:^15}|{:>15}'.format("Donor Name", "Total Given", "Num Gifts", "Average Gift")
+    print(header)
+    print("-" * len(header))
+
+    for name in donor_names():
+        print('{:21}{:>15.2f}{:>16}{:>16.2f}'.format(name, total_given(name), num_gifts(name), avg_gift(name)))
```
natasha-aleksandrova: although our data set is really small in our use case, consider a much larger data set and performance issues that could arise here as you are looping over `donations` a total of 4 times.

taking performance aside, having a single loop over donations would actually make your code much simpler:
```
for donation in donations:
    name = donation[0]
    donation_data = donation[1:]
    total = sum(donation_data)
    num_of_gifts = len(donation_data)
    avg = total / num_of_gifts

```

*NOTE: I just realized that your data structure doesn't append to existing donor entry, but essentially duplicates the donor. I would recommend doing the append to existing entry to keep the data nicely organized, which makes access of data much simpler throughout. So most of my comments are around the assumption that donor entries are not duplicated.*

students/csdotson/lesson03/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+    return count
+
+def avg_gift(donor):
+    # Compute/return average gifr for a given donor
+    return (total_given(donor)/num_gifts(donor))
+
+
+if __name__ == "__main__":
+    # Will only execute if mailroom.py is run as a script
+    prompt()
```
natasha-aleksandrova: nice work on the mailroom!

Please review my comments/suggestions, let me know if you have any questions on them. If you would like to take another pass at his we will be happy to review again!

PR #163, opened 2018-02-20
completed dict_lab.py (lesson04)

johnpharmd:


students/johnpharmd/lesson04/dict_lab.py
```diff
@@ -0,0 +1,63 @@
...
+print('"Mango" is a value in the dictionary:', 'Mango' in dict1.values())
+print()
+
+# Dictionaries02
+dict2 = dict1.copy()
+print('dict1:', dict1)
+print('dict2:', dict2)
+
+for k, v in dict1.items():
+	dict2[k] = v.lower().count('t')
```
natasha-aleksandrova: your indentation is off (should be exactly 4 spaces).

most editors can be configured to have this set automatically.

students/johnpharmd/lesson04/dict_lab.py
```diff
@@ -0,0 +1,63 @@
...
+print(dict1)
+
+print(dict1.keys())
+print(dict1.values())
+print('"cake" is a key in the dictionary:', 'cake' in dict1)
+print('"Mango" is a value in the dictionary:', 'Mango' in dict1.values())
+print()
+
+# Dictionaries02
+dict2 = dict1.copy()
```
natasha-aleksandrova: be aware that `copy` makes a shallow copy of dict. it won't make a difference in this simple dict, but if your values hold list for example then it will cause issues.

PR #164, opened 2018-02-20
First swing at Mailroom

MikeS-1:
Looking at other examples definitely helped a lot

### DIFF
--- a/students/MikeShand/lesson03/mailroom.py
+++ b/students/MikeShand/lesson03/mailroom.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python3
+
+
+
+donors=[['Andy', [10.00]],['Bill',[15.00, 25.00]],['Chuck',[20.00,30.00,40.00]]]
+
+def donor_list():
+    donor_list=[]
+    for x in donors:
+        print(x[0])
+    thank_you()
+
+
+def thank_you():
+    print('Please enter the donor name\n (Type "list" for a list of current donor names)\n Press "q" to return to console')
+    cd=input(':')
+
+    if cd.lower()=='list':
+        donor_list()
+
+    elif cd.lower()=='q':
+        console()   
+
+    else:
+
+        amount=float(input('Please enter the donation amount:'))
+
+        for j,k in enumerate(donors):
+            if cd==k[0]:
+                donors[j][1].append(amount)
+                break
+        else:
+            donors.append([cd,[amount]])
+
+        print('Thank you {} for your generous donation of ${:.2f}'.format(cd,amount))
+
+def donor_report():
+    print('Here is a list of donors and contributions')
+    print('Name                 Total                Donations            Average')
+    for data in donors:
+
+        print("|{:<20}|{:<20}|{:<20}|{:<20}|".format(data[0],sum(data[1]),len(data[1]),sum(data[1])/len(data[1])))
+        #print(f'{data[0]:<20}'+f'${sum(data[1]):20}'+f'{len(data[1]):20}'+f'${sum(data[1])/len(data[1]):<20}')
+
+    console()
+
+def console():
+    quit_console=False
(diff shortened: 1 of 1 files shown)


students/MikeShand/lesson03/mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+
+def donor_report():
+    print('Here is a list of donors and contributions')
+    print('Name                 Total                Donations            Average')
+    for data in donors:
+
+        print("|{:<20}|{:<20}|{:<20}|{:<20}|".format(data[0],sum(data[1]),len(data[1]),sum(data[1])/len(data[1])))
+        #print(f'{data[0]:<20}'+f'${sum(data[1]):20}'+f'{len(data[1]):20}'+f'${sum(data[1])/len(data[1]):<20}')
+
+    console()
```
natasha-aleksandrova: these calls back to console actually cause a bug.

if you run your program, and select report option (2) couple times then try to quit with option 3 it won't quit.

in your main function `console` your while loop will put the user back into the main prompt naturally, so having additional calls to `console` in other functions causes execution flow to mess up.

students/MikeShand/lesson03/mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+            donors.append([cd,[amount]])
+
+        print('Thank you {} for your generous donation of ${:.2f}'.format(cd,amount))
+
+def donor_report():
+    print('Here is a list of donors and contributions')
+    print('Name                 Total                Donations            Average')
+    for data in donors:
+
+        print("|{:<20}|{:<20}|{:<20}|{:<20}|".format(data[0],sum(data[1]),len(data[1]),sum(data[1])/len(data[1])))
```
natasha-aleksandrova: note that numbers are normally right justified and strings are left justified in tables.

natasha-aleksandrova: consideration for performance on large data sets:

I would recommend extracting sum and len values into variables so that the avg can be computed efficiently without computing sum and len again.

students/MikeShand/lesson03/mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+
+def thank_you():
+    print('Please enter the donor name\n (Type "list" for a list of current donor names)\n Press "q" to return to console')
+    cd=input(':')
+
+    if cd.lower()=='list':
+        donor_list()
+
+    elif cd.lower()=='q':
+        console()
```
natasha-aleksandrova: return here instead of calling console (see my other comment about program flow)

students/MikeShand/lesson03/mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+        donor_list()
+
+    elif cd.lower()=='q':
+        console()   
+
+    else:
+
+        amount=float(input('Please enter the donation amount:'))
+
+        for j,k in enumerate(donors):
```
natasha-aleksandrova: recommended to use more meaningful names: `for idx, donor in enumerate (donors)`

on another note how about a more simplified loop:
```
for name, donor_data in donors:
    if name == cd:
        donor_data.append(amount)
        break
else:
    donors.append([cd, [amount]])
```

MikeS-1:
Thank you for the feedback! I’ve submitted an update with the recommended changes.

I couldn’t figure out how to past the “pull” url on the website for a second submission, and I didn’t know if you needed it our not, so here it is just in case.

https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/173


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

natasha-aleksandrova:
yep that works, if you're doing updates to already submitted assignment just note so in the PR comment so we know and thats it.

PR #165, opened 2018-02-20
Lesson 2 Grid Printer files

traub01-devtrain:
I am adding the Lesson 2 Grid Printer files grid_printer.py and grid_printer2.py. I created a copy of break_me.py and put it in the Lesson 01 folder. I plan to remove the other break_me.py located in the students/jamestraub folder where my README is located for cleanup.

### DIFF
--- a/students/JamesTraub/Lesson02/JT-Fizz-Buzz.py
+++ b/students/JamesTraub/Lesson02/JT-Fizz-Buzz.py
@@ -1,0 +1,14 @@
+# Write a program that prints the numbers from 1 to 100 inclusive.
+# But for multiples of three print “Fizz” instead of the number.
+# For the multiples of five print “Buzz” instead of the number.
+# For numbers which are multiples of both three and five print “FizzBuzz” instead.
+
+for number in range(1, 101):
+    if number % 3 == 0 and number % 5 == 0:
+        print('FizzBuzz')
+    elif number % 5 == 0:
+        print('Buzz')
+    elif number % 3 == 0:
+        print('Fizz')
+    else:
+        print(number)
--- a/students/JamesTraub/Lesson02/grid_printer2.py
+++ b/students/JamesTraub/Lesson02/grid_printer2.py
@@ -1,0 +1,21 @@
+# chose to divide n by 2 to give columns an even width
+
+def print_grid(n):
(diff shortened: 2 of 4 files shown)


students/JamesTraub/Lesson02/JT-Fizz-Buzz.py
```diff
@@ -0,0 +1,14 @@
...
+
+for number in range(1, 101):
+    if number % 3 == 0 and number % 5 == 0:
+        print('FizzBuzz')
+    elif number % 5 == 0:
+        print('Buzz')
+    elif number % 3 == 0:
+        print('Fizz')
+    else:
+        print(number)
```
natasha-aleksandrova: your file name should be `jt_fizz_buzz.py` 

for reference: https://www.python.org/dev/peps/pep-0008/#package-and-module-names

otherwise fizz buzz looks great!

students/JamesTraub/Lesson02/grid_printer2.py
```diff
@@ -0,0 +1,19 @@
+# chose to divide n by 2 to give columns an even width
+
+def print_grid(n,s):
+    pls = '+'
+    mns = '-'
+    pl = '|'
+    spc = ' '
+    dventry = s
+    grdsz = (dventry * mns)
```
natasha-aleksandrova: I think this might be a C lang convention to use var names with no vowels?

either way it is more readable to have `grid_size` than `grdsz` 🙂 (I'm picking on `grdsz` but it is a general comment.)

sometimes it makes sense to have short var names, and sometimes it makes sense to spell it out.

students/JamesTraub/Lesson02/grid_printer2.py
```diff
@@ -0,0 +1,19 @@
...
+    spcsz = (dventry * spc)
+    plsln = (pls + grdsz + pls + grdsz + pls)
+    plline = (pl + spcsz + pl + spcsz + pl)
+    for b in range(n):
```
natasha-aleksandrova: it is a common convention in python to use `_` var name for throw-away vars in loops. it clearly indicates that this var is never used and pretty much ignore it. Example:
```
for _ in range(n)
```

PR #167, opened 2018-02-20
Lesson03 - Slicing Lab Exercise

DeFeNdog:


students/carlos_novoa/lesson03/slicing.py
```diff
@@ -0,0 +1,60 @@
...
+def reversed(seq):
+    """elements reversed (just with slicing)"""
+    return seq[::-1]
+
+
+def thirds(seq):
+    """order: middle third, last third, first third """
+    return seq[int(len(seq) / 3):int((len(seq) / 3) * 2)] +\
+        seq[-int(len(seq) / 3):] +\
+        seq[:int(len(seq) / 3)]
```
natasha-aleksandrova: to make this more readable extract `int(len(seq) / 3)` into var, not everything has to be a one liner 😉

PR #168, opened 2018-02-20
My First Mailroom Assignment

alexLaws:
I spoke with Natasha mid last week before I went on vacation about this assignment. I had done the full assignment more or less and then saw that it said not to use topics we hadn't covered yet. I'd already written it using dictionaries.  She said that would be ok and I should make a note of it here.

### DIFF
--- a/students/alexLaws/lesson03/mailroom.py
+++ b/students/alexLaws/lesson03/mailroom.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python3
+
+donor_list = []
+donor_list.append({'Name': 'Ryan Moore', 'Donation': 500})
+donor_list.append({'Name': 'Ryan Moore', 'Donation': 250})
+donor_list.append({'Name': 'Ted Laws', 'Donation': 1000})
+donor_list.append({'Name': 'Ted Laws', 'Donation': 100})
+donor_list.append({'Name': 'Ben Snell', 'Donation': 150})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 2000})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 2000})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 4000})
+donor_list.append({'Name': 'Beth Ross', 'Donation': 400})
+
+
+def user_action():
+    print("Choose an action from this list:")
+    print("Send a Thank You")
+    print("Create a Report")
+    print("Quit")
+    while True:
+        option = input("What is your choice: ")
+        if option.lower() in ["send a thank you", "create a report", "quit"]:
+            return option
+            break
+        else:
+            print("That is not an option. Choose again.")
+
+
+def dnrs(x):
+    names = []
+    for dnr in x:
+        person = dnr.get("Name")
+        if person in names:
+            pass
+        else:
+            names.append(person)
+    return names
+
+
+def send_to(x):
+    donors = dnrs(x)
+    recipient = input("Who is the thank you note for? Enter a full name or 'list' to see a list: ")
+    if recipient.lower() == 'list':
+        print("Here are all the donors:")
+        for name in donors:
+            print(name)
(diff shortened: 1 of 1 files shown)


students/alexLaws/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+
+donor_list = []
```
robalford: This could be written as a list of dict literals rather than making a call to `append` for each item in the list (something like this):

```
donor_list = [
    {'Name': 'Ryan Moore', 'Donation': 500},
    {'Name': 'Ryan Moore', 'Donation': 250},
    {'Name': 'Ted Laws', 'Donation': 1000},
 ]
```

students/alexLaws/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+donor_list.append({'Name': 'Ryan Moore', 'Donation': 500})
+donor_list.append({'Name': 'Ryan Moore', 'Donation': 250})
+donor_list.append({'Name': 'Ted Laws', 'Donation': 1000})
+donor_list.append({'Name': 'Ted Laws', 'Donation': 100})
+donor_list.append({'Name': 'Ben Snell', 'Donation': 150})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 2000})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 2000})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 4000})
+donor_list.append({'Name': 'Beth Ross', 'Donation': 400})
```
robalford: Can you think of a way to store multiple donation amounts in each 'record' so you don't have so much repeated data here? A dict is a great outer data structure to use here, but perhaps you could combine with a list of donation amounts for each donor? This could also simplify some of your lookup and calculation code as noted below.

students/alexLaws/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+def user_action():
+    print("Choose an action from this list:")
+    print("Send a Thank You")
+    print("Create a Report")
+    print("Quit")
+    while True:
+        option = input("What is your choice: ")
+        if option.lower() in ["send a thank you", "create a report", "quit"]:
+            return option
+            break
```
robalford: `return` will exit the function so no need to include `break` here

students/alexLaws/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+        else:
+            print("That is not an option. Choose again.")
+
+
+def dnrs(x):
+    names = []
+    for dnr in x:
+        person = dnr.get("Name")
+        if person in names:
```
robalford: Using a single 'record' for each donor in the `door_list` would clean this up a bit

students/alexLaws/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+            donor_total += int(i.get('Donation'))
+    return donor_total
+
+
+def name_cnt(person, x):
+    donor_count = 0
+    for i in x:
+        if i.get('Name') == person:
+            donor_count += 1
+    return donor_count
```
robalford: The `name_sum` and `name_cnt` functions could be replaced by Python's built in `sum` and `len` functions if donation data were stored in a list.

students/alexLaws/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+    amounts = []
+    for n in to_sort:
+        amounts.append(n.get('Total'))
+    amounts.sort(reverse=True)
+    ordered = []
+    for num in amounts:
+        for i in to_sort:
+            if i.get('Total') == num:
+                ordered.append(i.get('Name'))
+    return ordered
```
robalford: Check out the `key` parameter of Python's `list.sort()` method. Perhaps you could employ that to simplify things here?

PR #169, opened 2018-02-20
Finished slice with testing.

preactive:
Was trying to play with yield with fib function in last lesson.  Is there a better way to adjust variables in function with Yield and not use global variables?

Is asserts inline acceptable in the fashion I did them?

### DIFF
--- a/students/preactive/Lesson2/series.py
+++ b/students/preactive/Lesson2/series.py
@@ -7,6 +7,20 @@ def fib(n):
 # print(fib(8))
 
 
+a,b = 0,1
+def fib_gen():
+    global a, b
+    while True:
+        a, b = b, a + b
+        yield a
+f = fib_gen()
+next(f)
+next(f)
+next(f)
+next(f)
+print(next(f))
+
+
 def lucas(n):
     a, b = -1, 2
     for i in range(n):
--- a/students/preactive/Lesson3/slice.py
+++ b/students/preactive/Lesson3/slice.py
@@ -1,0 +1,40 @@
+def slicer():
+    a_string = "this is a string"
+    a_tuple = (2, 54, 13, 12, 5, 32)
+    a_arr = ["tide", "pod", "challenge", "is", "not", "for", "the", "weak"]
+    first_last(a_tuple)
+
+
+def first_last(slicer_input):
+    # with the first and last items exchanged
+    first = slicer_input[0]
+    last = slicer_input[-1]
+    swapped = last + slicer_input[1:len(slicer_input) - 1] + first
+    return swapped
+
+
+def odd_out():
+    # with every other item removed
+    print()
+
+
+def ambigus_ao_fourth():
+    # with the first 4 and the last 4 items removed,
+    # and then every other item in between
+    print()
+
+
+def reversed():
+    # with the elements reversed (just with slicing)
+    print()
+
+
+def thirds():
+    # with the middle third, then last third, then the
+    # first third in the new order
+    print()
+
+
+
+
+slicer()
--- a/students/preactive/Lesson3/slice.py
+++ b/students/preactive/Lesson3/slice.py
@@ -3,40 +3,59 @@ def slicer():
     a_tuple = (2, 54, 13, 12, 5, 32)
     a_arr = ["!", "!", "tide", "pod", "challenge", "is",
              "not", "for", "the", "weak", "!", "!"]
-    thirds(a_arr)
+    three_types = (a_string, a_tuple, a_arr)
+
+    for slice_item in three_types:
+        for_type = type(slice_item)
+        first_last(slice_item, for_type)
+        odd_out(slice_item, for_type)
+        ambigus_ao_fourth(slice_item, for_type)
+        reversed(slice_item, for_type)
+        thirds(slice_item, for_type)
 
 
-def first_last(slicer_input):
+
+def first_last(slicer_input, for_type):
     # with the first and last items exchanged
     swapped = slicer_input[-1:] + slicer_input[1:len(slicer_input) - 1] \
               + slicer_input[:1]
+    if for_type == "str":
+        print("str")
+    if for_type == "<class 'tuple'>":
+        print("tuple")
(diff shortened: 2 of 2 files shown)


preactive:
I did a `git commit --amend` for PEP8 fixes for slice.py

natasha-aleksandrova:
@preactive `git commit --amend` is appropriate when there is a need for it 🙂 

Typically you would make a new commit, remember that in a way commits tell a story of progression of the code.

example of commit history, most recent at the top
--- fix style for pep8
--- add fizz buzz assignment

when someone else or you a year later trying to figure out why something was added, commit history is a great tool.

amazingly that are even guidelines for commit messages 😃 
https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

students/preactive/Lesson2/series.py
```diff
@@ -7,6 +7,20 @@ def fib(n):
 # print(fib(8))
 
 
+a,b = 0,1
+def fib_gen():
+    global a, b
+    while True:
+        a, b = b, a + b
+        yield a
```
natasha-aleksandrova: as far as your question on yield (btw awesome that you are trying out `yield`)...

you shouldn't need globals here, this works as well:
```
def fib_gen():
    a,b = 0,1
    while True:
        a, b = b, a + b
        yield a
```

students/preactive/Lesson3/slice.py
```diff
@@ -0,0 +1,86 @@
...
+    # with the first and last items exchanged
+    swapped = slicer_input[-1:] + slicer_input[1:len(slicer_input) - 1] \
+              + slicer_input[:1]
+    if for_type is str:
+        assert swapped == "ghis is a strint", "String Failed"
+    if for_type is tuple:
+        assert swapped == (32, 54, 13, 12, 5, 2), "Tuple Failed"
+    if for_type is list:
+        assert swapped == ["!", "!", "tide", "pod", "challenge", "is",
+             "not", "for", "the", "weak", "!", "!"], "List failed"
```
natasha-aleksandrova: For your assert question.. I would keep asserts separate from the actual code.


in real project tests would go at the root of the project in the `tests` folder, in small scripts like this they would go into `if __name__ == '__main__'` block

which would now make your test logic much simpler:

`assert first_last("This is a string") == "ghis is a strint"` without the need for type checking

students/preactive/Lesson3/slice.py
```diff
@@ -0,0 +1,86 @@
...
+        first_last(slice_item, for_type)
+        odd_out(slice_item, for_type)
+        ambigus_ao_fourth(slice_item, for_type)
+        reversed(slice_item, for_type)
+        thirds(slice_item, for_type)
+
+
+def first_last(slicer_input, for_type):
+    # with the first and last items exchanged
+    swapped = slicer_input[-1:] + slicer_input[1:len(slicer_input) - 1] \
```
natasha-aleksandrova: note that the middle part can be simplified to: `slicer_input[1:-1]`

students/preactive/Lesson3/slice.py
```diff
@@ -0,0 +1,86 @@
...
+    if for_type is tuple:
+        assert slicer_reversed == (32, 5, 12, 13, 54, 2), "Tuple Failed"
+    if for_type is list:
+        assert slicer_reversed == ['!', '!', 'weak', 'the', 'for', 'not',
+                                   'is', 'challenge', 'pod', 'tide', '!', '!'], \
+            "List failed"
+    return slicer_reversed
+
+
+def thirds(s_i, for_type):
```
natasha-aleksandrova: more readable to call the var either of these: `seq` or `sequence`

students/preactive/Lesson3/slice.py
```diff
@@ -0,0 +1,86 @@
...
+    # with the middle third, then last third, then the
+    # first third in the new order
+    if len(s_i) % 3 == 0:
+        L = round(len(s_i) / 3)
```
natasha-aleksandrova: var names should be lower case, and avoid one letters (unless comprehensions and other short hand cases)

students/preactive/Lesson3/slice.py
```diff
@@ -0,0 +1,86 @@
...
+    a_arr = ["!", "!", "tide", "pod", "challenge", "is",
+             "not", "for", "the", "weak", "!", "!"]
+    three_types = (a_string, a_tuple, a_arr)
+
+    for slice_item in three_types:
+        for_type = type(slice_item)
+        first_last(slice_item, for_type)
+        odd_out(slice_item, for_type)
+        ambigus_ao_fourth(slice_item, for_type)
+        reversed(slice_item, for_type)
```
natasha-aleksandrova: be careful about overwriting built-in functions, as you can see github highlights `reversed`, so will some code editors.

natasha-aleksandrova:
@preactive I don't believe I see your submission for series assignment in the course website?

Please be sure to submit PR link to get the grade.

students/preactive/Lesson2/series.py
preactive: added the test and a modifed one that leads series with 0

students/preactive/Lesson3/slice.py
preactive: Fixed

students/preactive/Lesson3/slice.py
preactive: I was getting lazy.  Thank your for holding me to this.  Fixed.

students/preactive/Lesson3/slice.py
preactive: Fixed, changed verbage

preactive:
I checked the others and I did submit them in the portal

students/preactive/Lesson3/slice.py
preactive: fixed

students/preactive/Lesson3/slice.py
preactive: Moved asserts into name/main and refactored.

PR #173, opened 2018-02-21
Updated with recommended changes

MikeS-1:
I couldn't figure out how to get the PEP8 linter working in Sublime, and the web based version I found was less than helpful, so I just used PyCharm. I like Sublime, but PyCharm seems pretty awesome. Are there any known drawbacks to using one over the other?

### DIFF
--- a/students/MikeShand/lesson03/mailroom.py
+++ b/students/MikeShand/lesson03/mailroom.py
@@ -1,51 +1,51 @@
 #!/usr/bin/env python3
 
+donors = [['Andy', [10.00]], ['Bill', [15.00, 25.00]], ['Chuck', [20.00, 30.00, 40.00]]]
 
-
-donors=[['Andy', [10.00]],['Bill',[15.00, 25.00]],['Chuck',[20.00,30.00,40.00]]]
 
 def donor_list():
-    donor_list=[]
     for x in donors:
         print(x[0])
     thank_you()
 
 
 def thank_you():
-    print('Please enter the donor name\n (Type "list" for a list of current donor names)\n Press "q" to return to console')
-    cd=input(':')
+    print('Please enter the donor name\n (Type "list" for a list of current donor names)\n '
+          'Press "q" to return to console')
+    cd = input(':')
 
-    if cd.lower()=='list':
+    if cd.lower() == 'list':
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
Sublime is a generic editor that can be setup (via plugins) for any language. Pycharm is specifically built for python dev, and is a full featured, powerful IDE. I switched from sublime to pycharm over a year ago and haven't looked back.

students/MikeShand/lesson03/mailroom.py
```diff
@@ -1,51 +1,51 @@
...
+
+        print('Thank you {} for your generous donation of ${:.2f}'.format(cd, amount))
 
-        print('Thank you {} for your generous donation of ${:.2f}'.format(cd,amount))
 
 def donor_report():
     print('Here is a list of donors and contributions')
     print('Name                 Total                Donations            Average')
     for data in donors:
+        x = int(sum(data[1]))
```
natasha-aleksandrova: I would avoid `x` and `y` variables, remember usually someone else will be reading your code 😉 

`total` and `num_of_gifts` are more clear

x and y type of variable are more appropriate in comprehensions (example: `[len(x) for x in sequence]`)

students/MikeShand/lesson03/mailroom.py
```diff
@@ -1,51 +1,51 @@
...
+        y = int(len(data[1]))
```
natasha-aleksandrova: `len` will return an `int` so no need to cast to `int`

students/MikeShand/lesson03/mailroom.py
```diff
@@ -1,51 +1,51 @@
...
 
     else:
 
-        amount=float(input('Please enter the donation amount:'))
+        amount = float(input('Please enter the donation amount:'))
 
-        for j,k in enumerate(donors):
-            if cd==k[0]:
+        for j, k in enumerate(donors):
+            if cd == k[0]:
```
natasha-aleksandrova: I would void `j` and `k` vars as they are ambiguous, and I would also not use `enumerate` here. 

You don't need the index, you can just loop over: `for donor in donors` since you end up accessing donor by index later on anyway.

PR #174, opened 2018-02-21
Herela: Lesson 5: exception, mailroom3

harvey-herela:
My previous attempt at submitting the exceptions exercise failed, so ignore the GitHub link on the startlearning website for exceptions. This commit here, this pull-request, contains both the exceptions and the mailroom3.

### DIFF
--- a/students/HarveyHerela/lesson05/mailroom3.py
+++ b/students/HarveyHerela/lesson05/mailroom3.py
@@ -1,0 +1,152 @@
+#!/usr/bin/env python
+
+import datetime
+import os
+
+donors = [{"name": ("Chase", "Zuma"), "donations": [25 * x for x in range(2)]},
+          {"name": ("Marshall", "Rocky"), "donations": [30 * x for x in range(2, 4)]},
+          {"name": ("Skye", "Rubble"), "donations": [45 * x for x in range(10, 13)]}]
+
+
+def get_name(prompt):
+    """Anything can be a name, so accept anything as a name."""
+    getting_name = True
+
+    while getting_name:
+
+        print("Type 'list' for a list of names in the database, '~' to go back")
+        response = input(prompt)
+        response = response.strip()
+
+        if response.lower() == "list":
+            for d in donors:
+                print("{0} {1}".format(*d["name"]))
+        else:
+            getting_name = False
+    return response
+
+
+def send_thank_you():
+    """Prompts for a donor name and donation amount, then prints a 'thank you' message to the screen."""
+    firstname = get_name("First name: ")
+
+    if firstname == "~":
+        return
+
+    lastname = get_name("Last name: ")
+
+    if lastname == "~":
+        return
+
+    amount = "nada"
+    while not amount.isnumeric():
+        print("Donation amount, or '~' to go to main menu:")
+        amount = input("$ ")
+        amount = amount.strip()
+        if amount == '~':
+            return
+    amount = float(amount)
+
+    # Add the donation
+    name = (firstname, lastname)
+    for d in donors:
+        if name in d:
+            # If it's an existing donor, add it to their donation list.
+            d["donations"].append(amount)
+            break
+    else:
+        # If it's a new donor, create a new entry.
+        donors.append({"name": name, "donations": [amount]})
+
+    thank_you_msg = "Thank you {firstname} {lastname} for your generous donation of ${amount:.2f}".format(
+        firstname=firstname, lastname=lastname, amount=amount)
+    print(thank_you_msg)
+
+
+def get_total(donor_data):
+    return donor_data[1]
+
+
+def create_row(donor_data):
+    fullname = "{0} {1}".format(*donor_data["name"])
+    total = sum(donor_data["donations"])
+    num = len(donor_data["donations"])
+    avg = total / num
+    return [fullname, total, num, avg]
+
+
+def create_report():
+    """Prints a list of all donors, sorted by total amount given."""
+    report = str()
+    report += "Donor Name    | Total Given   | Num Gifts | Average Gift"
+    report += "\n"
+    report += "--------------------------------------------------------"
+    report += "\n"
+
+    # Create a list with the data to be printed
+    data_list = [create_row(d) for d in donors]
+
+    # Sort the list by total amount given
+    data_list.sort(key=get_total, reverse=True)
+
+    # Define the formatter
+    line_fmtr = "{name:<14}  ${total:>13.2f}   {num:>9d}  ${avg:>12.2f}\n".format
+    for d in data_list:
+        report += line_fmtr(name=d[0], total=d[1], num=d[2], avg=d[3])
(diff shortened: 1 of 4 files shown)


natasha-aleksandrova:
@harvey-herela glad that you got it worked out 🙂and thanks for leaving us a comment as a reminder!

students/HarveyHerela/lesson05/mailroom3.py
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env python
+
+import datetime
+import os
+
+donors = [{"name": ("Chase", "Zuma"), "donations": [25 * x for x in range(2)]},
+          {"name": ("Marshall", "Rocky"), "donations": [30 * x for x in range(2, 4)]},
+          {"name": ("Skye", "Rubble"), "donations": [45 * x for x in range(10, 13)]}]
```
robalford: Don't think I'd use comprehensions here, since it's more readable to see the donation amounts listed out one at a time

students/HarveyHerela/lesson05/mailroom3.py
```diff
@@ -0,0 +1,152 @@
...
+        return
+
+    amount = "nada"
+    while not amount.isnumeric():
+        print("Donation amount, or '~' to go to main menu:")
+        amount = input("$ ")
+        amount = amount.strip()
+        if amount == '~':
+            return
+    amount = float(amount)
```
robalford: This could be a good use case for a `try/except` block in place of the `while` loop type check

students/HarveyHerela/lesson05/mailroom3.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # Add the donation
+    name = (firstname, lastname)
+    for d in donors:
+        if name in d:
```
robalford: This should be `if name in d.values():` because the `in` operator checks for the existence of keys in the dict

students/HarveyHerela/lesson05/mailroom3.py
```diff
@@ -0,0 +1,152 @@
...
+def create_report():
+    """Prints a list of all donors, sorted by total amount given."""
+    report = str()
+    report += "Donor Name    | Total Given   | Num Gifts | Average Gift"
+    report += "\n"
+    report += "--------------------------------------------------------"
+    report += "\n"
+
+    # Create a list with the data to be printed
+    data_list = [create_row(d) for d in donors]
```
robalford: Nice!

students/HarveyHerela/lesson05/mailroom3.py
```diff
@@ -0,0 +1,152 @@
...
+        try:
+            with open(filename, 'w') as outfile, open("template.txt", 'r') as infile:
+                donations = d["donations"]
+                last = donations[len(donations) - 1]
+                total = 0
+                for gift in donations:
+                    total += gift
+                name = d["name"]
+                outfile.write(infile.read().format(firstname=name[0], lastname=name[1], amount=last, total=total))
+        except FileNotFoundError as fnferr:
```
robalford: I would move this `except` block closer to the code that might raise `FileNotFoundError`. And I'd move the code that sets up the `name`, `last` and `total` variables out of the `with` block, since I don't think any of this depends on the file being open to run. That would let you use a separate `try/except` block to catch the `KeyError` and move `FileNotFoundError` and `ValueError` exception blocks closer to the calls to `open` and `write` that could potentially raise these errors. Does that make sense?

students/HarveyHerela/lesson05/mailroom3.py
```diff
@@ -0,0 +1,152 @@
...
+            print(fnferr)
+            print("\nThere was an error reading the template file")
+            return
+        except KeyError as kerr:
+            print("Key Error:", kerr, "is a bad key.")
+            print("There was an error with the template {keys}. Verify the {keys} are correct.")
```
robalford: Need to call `string.format()` here

students/HarveyHerela/lesson05/mailroom3.py
```diff
@@ -0,0 +1,152 @@
...
+            return
+        except ValueError as verr:
```
robalford: Not currently doing anything with the `verr` variable here

students/HarveyHerela/lesson05/mailroom3.py
harvey-herela: It makes sense, and it's something I wanted to do when I wrote the code, but I wasn't sure how to do it. How should I use a "try/except" block with a "with" block?

robalford: I think you're taking the right approach overall by placing the 'with' block inside the 'try' block and catching the 'FileNotFoundError'. I would just pull out the code that might raise the KeyError into a separate 'try' block mainly for readability and to be extra careful about the possibility of unrelated or uncaught exceptions bubbling up in an unexpected place in the code.

robalford: @natasha-aleksandrova do you have any thought on this?

harvey-herela: You mean like nested try/except blocks?

robalford: No just two separate blocks like this (I have not tested this so apologies if this breaks your code in some way I'm not seeing):

```
    for d in donors:
        try:
            donations = d["donations"]  
            last = donations[len(donations) - 1]
            total = 0
            for gift in donations:
                total += gift
            name = d["name"]
        except KeyError as kerr:  
            print("Key Error:", kerr, "is a bad key.")
            print("There was an error with the template {keys}. Verify the {keys} are correct.")  
            return
        filename = "./{dir}/{0}_{1}.txt".format(*d["name"], dir=subdir)
        try:
            with open(filename, 'w') as outfile, open("template.txt", 'r') as infile:
                outfile.write(infile.read().format(firstname=name[0], lastname=name[1], amount=last, total=total))
        except FileNotFoundError as fnferr:  
            print(fnferr)
            print("\nThere was an error reading the template file")
            return
        except ValueError as verr:  
            print("There was an error while reading the formatting strings.")
            return
        num += 1
```

But the `KeyError` `try` block should probably be refined from here as well because it's still kind of a lot of code to place within a single try block and there are multiple key lookups in there. Generally you just want to handle the exception as close to where it occurred as possible.

natasha-aleksandrova: the above makes sense (moving the dictionary logic out of `with` block).. although I would say the try / except for dictionary logic may be an overkill. `donors` is accessed all over the place and kind of under a contract for expected keys, if you put try/except for KeyError everywhere those keys are accessed that would be a bit of a mess.

my view on try/except is to add to places that depend on external things like file existing, user input, http calls to some APIs or known potential failures in general; when within my own code I usually know what to expect (of course there are always bugs but those should be getting fixed).

PR #178, opened 2018-02-22
Lesson 4 assignments

kuhnbt7:


students/kuhnbt/lesson4/mailroom_v2.py
```diff
@@ -0,0 +1,105 @@
...
+
+def send_to_everyone(directory=os.getcwd()):
+    """Write thank you notes to specified directory for each donor in
+       donor_info"""
+    user_dir = input('Change output directory (y/n)?')
+    if user_dir.lower() == 'y':
+        directory = input('Please enter desired directory:')
+    today = str(datetime.datetime.today())[:10]
+    for donor in donor_info:
+        filename = '/'.join([directory, donor])+'_' + today + '.txt'
```
natasha-aleksandrova: check out `os.path.join`, it is the recommended way to create file paths

kuhnbt7:
Hi Natasha, thanks for the helpful comments! You're right, I didn't actually call my functions in the kata script. I've corrected this and also amended my print function to make the text appear more like the real thing. I'll submit another pull request with the changes.

natasha-aleksandrova:
sounds good, thank you!

PR #179, opened 2018-02-22
Lesson 03 - Emmanuel Twum - mail room

etwum:
assignment complete

### DIFF
--- a/students/etwum/lesson03/mailroom.py
+++ b/students/etwum/lesson03/mailroom.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+
+if __name__ == "__main__":
+    print()
+
+
+def send_thank_you():
+
+    return print("works")
+
+
+def create_report():
+
+    return print("works")
+
+print("Welcome to the Mail Room")
+print("------------------------------------------------------------------------")
+# ****Input/Output****
+while True:
+    print("""
+    Menu of Options
+    1) Send Thank You Note
+    2) Create Report
+    3) Exit Program
+    """)
+    strChoice = None
+    try:
+        strChoice = int(input("Which option would you like to perform? Input a number [1 to 3] "))
+        if strChoice == 1 or strChoice == 2 or strChoice == 3:
+            print()
+        else:
+            raise Exception
+    except Exception:
+        print("Please input a valid option: 1, 2, or 3")
+
+    if (strChoice == 1):
+        send_thank_you()
+        continue
+
+    elif (strChoice == 2):
+        create_report()
+        continue
+    elif (strChoice == 3):
+        break 
--- a/students/etwum/lesson03/mailroom.py
+++ b/students/etwum/lesson03/mailroom.py
@@ -3,15 +3,26 @@
 if __name__ == "__main__":
     print()
 
+lstHeader = [["Donor Name", "| Total Donation(s)", "| # of Donations", "| Avg Donation"]]
+lstDonors = [["Lionel Messi", 1000000.00, 5, 20000.00],["Thierry Henry", 500, 1, 500], ["Michael Jordan", 45000, 3, 15000],
+        ["Kobe Bryant", 8000, 2, 4000]]
 
 def send_thank_you():
 
     return print("works")
 
+def sort_list(lstDonors):
+    # used to sort the donor list by the total donations column
+    return lstDonors[1]
 
 def create_report():
+    # creates a report of the the donors
+    for x in lstHeader:
+        print('{:<25}{:<20}{:<17}{:<15}'.format(*x))
+    print("----------------------------------------------------------------------------")
+    for x in sorted(lstDonors,key=(sort_list), reverse= True):
+        print('{:<25} $ {:<20}{:^14} $ {:<15}'.format(*x))
 
-    return print("works")
 
 print("Welcome to the Mail Room")
 print("------------------------------------------------------------------------")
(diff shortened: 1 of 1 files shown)


students/etwum/lesson03/mailroom.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
+
+if __name__ == "__main__":
+    print()
```
natasha-aleksandrova: for readability and consistency (if __name__ usually goes at the bottom) I would move this down at the bottom of the script and have the main guts there, calling support functions and no extra code outside of that.

students/etwum/lesson03/mailroom.py
```diff
@@ -0,0 +1,117 @@
...
+
+# headers used in table
+lstHeader = [["Donor Name", "| Total Donation(s)", "| # of Donations", "| Avg Donation"]]
+
+# initial list of donors
+lstDonors = [["Lionel Messi", 1000000.00, 5, 20000.00],["Thierry Henry", 500, 1, 500], ["Michael Jordan", 45000, 3, 15000],
```
natasha-aleksandrova: `lstDonors` and `lstHeader` should be snake_case -> `lst_donors` and `lst_header` (something like `header_list` and `donors_list` would be even better)

students/etwum/lesson03/mailroom.py
```diff
@@ -0,0 +1,117 @@
...
+    Menu of Options
+    1) Send Thank You Note
+    2) Create Report
+    3) Exit Program
+    """)
+    strChoice = None
+
+    # try/except block to make sure the user inputs a valid option
+    try:
+        strChoice = int(input("Which option would you like to perform? Input a number [1 to 3] "))
```
natasha-aleksandrova: `str_choice` (snake_case for var names)

students/etwum/lesson03/mailroom.py
```diff
@@ -0,0 +1,117 @@
...
+        if strChoice == 1 or strChoice == 2 or strChoice == 3:
```
natasha-aleksandrova: cleaner version: `if str_choice in (1, 2, 3)`

students/etwum/lesson03/mailroom.py
```diff
@@ -0,0 +1,117 @@
...
+            print()
+        else:
+            raise Exception
+    except Exception:
+        print("Please input a valid option: 1, 2, or 3")
+
+    if (strChoice == 1):
```
natasha-aleksandrova: no need for parens here

students/etwum/lesson03/mailroom.py
```diff
@@ -0,0 +1,117 @@
...
+        send_thank_you()
+        continue
+
+    elif (strChoice == 2):
+        create_report()
+        continue
```
natasha-aleksandrova: no need for continue statements here and above, your program will continue naturally

students/etwum/lesson03/mailroom.py
```diff
@@ -0,0 +1,117 @@
...
+    return lstDonors[1]
+
+def create_report():
+    # creates a report of the the donors
+
+    for x in lstHeader:
+        print('{:<25}{:<20}{:<17}{:<15}'.format(*x))
+    print("----------------------------------------------------------------------------")
+    for x in sorted(lstDonors,key=(sort_list), reverse= True):
+        print('{:<25} $ {:<20}{:^14} $ {:<15}'.format(*x))
```
natasha-aleksandrova: note that commonly in tables numbers are right justified and strings are left justified

students/etwum/lesson03/mailroom.py
```diff
@@ -0,0 +1,117 @@
...
+
+def sort_list(lstDonors):
+    # used to sort the donor list by the total donations column
+
+    return lstDonors[1]
+
+def create_report():
+    # creates a report of the the donors
+
+    for x in lstHeader:
```
natasha-aleksandrova: since this is only used once it would make more sense to have it defined locally in this function rather than at the top of the file.

students/etwum/lesson03/mailroom.py
```diff
@@ -0,0 +1,117 @@
...
+        new_donation_amount = None
+        boolValid = False
+        while boolValid != True:
+            try:
+                new_donation_amount = float(input("Please input a donation amount in order to add the donor to the list. "))
+            except ValueError as e:
+                print(e)
+                print("Please input a valid number")
+            else:
+                boolValid = True
```
natasha-aleksandrova: you can also just call `break` here with a `while True` loop

students/etwum/lesson03/mailroom.py
```diff
@@ -0,0 +1,117 @@
...
+    donor_full_name = donor_first_name.capitalize() + " " + donor_last_name.capitalize()
+    print('-----------------------------------------------------')
+    print()
+
+    # prints an email if the donors name is currently in the list
+    for x in lstDonors:
+        if x[0] == donor_full_name:
+            email = "Dear {a},\n\nThank you for your generous donation of ${b:.2f} to our charity.\n".format(a=x[0], b=x[1])
+            print(email)
+            break
```
natasha-aleksandrova: you should have logic here to add add donation amount to existing donors.

students/etwum/lesson03/mailroom.py
etwum: I'm guessing you're talking about the header list?

natasha-aleksandrova: correct, referring to `lstHeader`

PR #180, opened 2018-02-22
lesson03: finished mailroom.py

22e88:


students/elmar_m/lesson03/mailroom.py
```diff
@@ -0,0 +1,88 @@
...
+    print(opts)
+
+
+def report():
+    '''
+    Show an overview of current donors and donations
+    '''
+    # get the highest number of digits to create formatstring accordingly:
+    sumlist = []
+    for i in donors.keys():
```
natasha-aleksandrova: since you only care about values, you should loop through values:
```
for donor_data in donors.values():
    sumlist.append(sum(donor_data))
```

students/elmar_m/lesson03/mailroom.py
```diff
@@ -0,0 +1,88 @@
...
+        sumlist.append(sum(donors[i]))
+    maxn = len(str(max(sumlist)))
+    maxn += 2
```
natasha-aleksandrova: here is another pattern to get max value:
```
max_val = 0
for donor_data in donors.values():
    max_val = max(max_val, sum(value))
```

students/elmar_m/lesson03/mailroom.py
```diff
@@ -0,0 +1,88 @@
...
+
+    fstring = '{:<20} ' + '|' + '{:>' + str(maxn) + '} ' + '|' + '{:>9}' + '|' + '{:>20}' 
+    print(fstring.format('Donor Name', 'Total', 'Num Gifts', 'Average Gift'))  
+    print('-' * (maxn + 54)) 
+    for i in donors.keys():
```
natasha-aleksandrova: better to use `for name, donor_data in donors.items()`

PR #181, opened 2018-02-22
[DennisLee][Lesson 02] Grid Printer exercise submission.

DennisLee-DennisLee:
Everything works when I test the file. I'm confident in the TableGrid class code, but I'm actually not sure why the tool invocation works in the __name=='main' clause. In line 77, when I instantiate the class, the interpreter kept complaining whenever I tried to add the row/col count and cell height/width as parameters. After I called the class without parameters, the tool operated as desired, though I still feel like the call is incorrect.


students/DennisLee/lesson02/grid_printer_class.py
```diff
@@ -0,0 +1,78 @@
...
+    else:
+        rows_and_cols = 4
+
+    if args.cell_size:
+        cell_size: int = args.cell_size
+    else:
+        cell_size = 6
+
+    tg = TableGrid()
+    tg.print_grid()
```
natasha-aleksandrova: we're not doing classes til way later in the course 🙂 so this is a way advanced first assignment!

I would recommend sticking to what has been learned so far in the class material for future assignment as that will make the most sense instead of working with classes for them.

students/DennisLee/lesson02/grid_printer_class.py
```diff
@@ -0,0 +1,78 @@
...
+
+        :return:  True if the function succeeds; False if the function fails.
+        """
+
+        horizontal_rule: str = self.build_line(True)
+        interior_line: str = self.build_line(False)
+
+        print((horizontal_rule + (interior_line * cell_size)) * rows_and_cols
+             + horizontal_rule)
+        return True
```
natasha-aleksandrova: you don't need return True here it's sort of a `void` function

students/DennisLee/lesson02/grid_printer_class.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    Create a square table grid, specifying the number of rows/columns
+    and the cell height/width.
+
+    :rows_and_cols:  The number of rows and columns in the grid.
+
+    :cell_size:  The height and width of a grid cell.
+    """
+
+    def __init(self, rows_and_cols: int, cell_size: int):
```
natasha-aleksandrova: the reason this wasn't working for you is because this is supposed to be `__init__` not `__init`

students/DennisLee/lesson02/grid_printer_class.py
```diff
@@ -0,0 +1,78 @@
...
+
+        :border:  Whether to create a horizontal rule. True indicates yes;
+                  False (default) specifies an interior line (containing just
+                  the vertical rules, with spaces in between those borders for
+                  the cell interior).
+
+        :return:  The text string for a table line.
+        """
+
+        PLUS, MINUS, PIPE, SPACE, LINEFEED = '+', '-', '|', ' ', '\n'
```
natasha-aleksandrova: typically constants should go at the top of the file, otherwise they should be regular lower case vars when within a method.

students/DennisLee/lesson02/grid_printer_class.py
```diff
@@ -0,0 +1,78 @@
...
+        vertical_line_char: str = PIPE
+        inside_char: str = SPACE
+
+        # horizontal rules contain plus & minus characters, not pipes & spaces
+        if border:
+            vertical_line_char, inside_char = PLUS, MINUS
```
natasha-aleksandrova: what is the purpose of reassigning the constant vars?

students/DennisLee/lesson02/grid_printer_class.py
```diff
@@ -0,0 +1,78 @@
...
+import argparse
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-t", "--table_size", type = int, 
+                        help = "specify the number of rows or columns in the \
+                        table (default 4)")
+    parser.add_argument("-c", "--cell_size", type = int,
+                        help = "specify the height and width of a table cell \
+                        (default 6)")
+    args = parser.parse_args()
```
natasha-aleksandrova: fancy!

students/DennisLee/lesson02/grid_printer_class.py
```diff
@@ -0,0 +1,78 @@
...
+
+        # horizontal rules contain plus & minus characters, not pipes & spaces
+        if border:
+            vertical_line_char, inside_char = PLUS, MINUS
+        
+        return (vertical_line_char + SPACE + (inside_char + SPACE) * cell_size
+            ) * rows_and_cols + vertical_line_char + LINEFEED
+
+
+import argparse
```
natasha-aleksandrova: the imports should go at the top of the file

students/DennisLee/lesson02/grid_printer_class.py
DennisLee-DennisLee: Sorry, I've been a bit confused by the information flow of the course. The Lesson 2 Content page had a link from the Tutorials section to Useful Python Learning Resources, which led me to the Python for You and Me web site, so I read the Class page there (along with the rest of the tutorial). It's hard to know when to stop reading, actually.

students/DennisLee/lesson02/grid_printer_class.py
DennisLee-DennisLee: vertical_line_char and inside_char are not constants - those represent the text characters that change depending on whether a border line or an interior cell line are being drawn.

students/DennisLee/lesson02/grid_printer_class.py
DennisLee-DennisLee: That explains it. And since I used **rows_and_cols** and **cell_size** in the class's methods without prefixing with **self.**, and then used the exact same names in the **if __name__=='main'** clause, those values were in scope to carry over into the methods without causing an error.

students/DennisLee/lesson02/grid_printer_class.py
natasha-aleksandrova: yes there are a lot of additional links and I can see how that can be confusing (we have already provided that feedback to the content developers).

I would stick to topics on the main course website (content and videos) as "this is what we are learning in the lesson".

PR #183, opened 2018-02-22
commit for submitted code, ready for review

JeroenSweerts:
Please review this code for the Slicing Lab Exercise.
Thanks.

Best regards,

Jeroen

### DIFF
--- a/students/JeroenSweerts/Lesson3/Slicing_Lab_Exercise.py
+++ b/students/JeroenSweerts/Lesson3/Slicing_Lab_Exercise.py
@@ -1,0 +1,128 @@
+a_string = "this is a string"
+a_tuple = (2, 54, 13, 12, 5, 32)
+food = ['spam', 'eggs', 'ham']
+
+
+
+
+def exchange_first_last(seq):
+    """the first and last items exchanged"""
+    tipe = type(seq)
+    seq = list(seq)
+    first = seq[0]
+    last = seq[-1]
+    a_new_sequence = seq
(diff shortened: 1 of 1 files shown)


students/JeroenSweerts/Lesson3/Slicing_Lab_Exercise.py
```diff
@@ -0,0 +1,128 @@
...
+    last = seq[-1]
+    a_new_sequence = seq
+    a_new_sequence[0] = last
+    a_new_sequence[-1] = first
+    if tipe == str:
+        return ''.join(map(str, a_new_sequence))
+    elif tipe == tuple:
+        return tuple(a_new_sequence)
+    else:
+        return a_new_sequence
```
natasha-aleksandrova: @JeroenSweerts although it is important to get the 'right answer', it is also important to exercise the right tools 🙂 

since this is a slicing lab, you should be almost exclusively utilizing [slicing functionality](https://www.pythoncentral.io/how-to-slice-listsarrays-and-tuples-in-python/)

if you're only using slicing you also don't need to handle special cases for different data types (string vs tuple vs list).

here is example of slicing to exchange first / last:
```

```
above is using slicing, will work for string, tuple, and a list and very simple 🙂

JeroenSweerts:
Hi,

I rewrote the code and did a new pull request. On the website I can’t enter the URL and can’t upload the file. The pull request number is 187.

Best regards,

Jeroen

PR #184, opened 2018-02-23
Lesson04 Activity

alexLaws:
Code for the dict Lab and File Lab.

### DIFF
--- a/students/alexLaws/lesson04/dict_lab.py
+++ b/students/alexLaws/lesson04/dict_lab.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python3
+
+import os
+
+# Dictionaries 1
+'''
+person = {'name': 'Chris', 'city': 'Seattle', 'cake': 'Chocolate'}
+print(person)
+person.pop('cake')
+person['fruit'] = 'Mango'
+print(person)
+
+print(person.keys())
+print(person.values())
+print('cake' in person)
+print('Mango' in person.values())
+
+# Dictionaries 2
+
+num_Ts = {}
+for i in person:
+    num_Ts[i] = person[i].lower().count('t')
+
+print(num_Ts)
+
+# Sets 1
+
+s2 = set(filter(lambda x: x % 2 == 0, range(21)))
+s3 = set(filter(lambda x: x % 3 == 0, range(21)))
+s4 = set(filter(lambda x: x % 4 == 0, range(21)))
+
+print(s2)
+print(s3)
+print(s4)
+
+print(s3.issubset(s2))
+print(s4.issubset(s2))
+
+# Sets 2
+
+char_Set = set("python")
+char_Set.add("i")
+print(char_Set)
+
+froz_set = frozenset("marathon")
+print(froz_set)
+
+print(char_Set.union(froz_set))
+print(char_Set.intersection(froz_set))
+
+# File Practice
+
+'''
+def print_dir():
+    for i in os.listdir():
+        print(i)
+
+
+def copy_file(file, dest=None):
+    buffer = 256
+    copy = "{}\{}_copy{}".format(dest, file[:-4], file[-4:])
+    f = open(file, 'rb')
+    j = open(copy, 'wb')
+    while True:
+        temp = f.read(buffer)
+        if len(temp) > 0:
+            j.write(temp)
+        else:
+            break
+    j.close()
--- a/students/alexLaws/lesson04/dict_lab.py
+++ b/students/alexLaws/lesson04/dict_lab.py
@@ -3,7 +3,7 @@
 import os
 
 # Dictionaries 1
-'''
+
 person = {'name': 'Chris', 'city': 'Seattle', 'cake': 'Chocolate'}
 print(person)
 person.pop('cake')
@@ -50,7 +50,7 @@ print(char_Set.intersection(froz_set))
 
 # File Practice
 
-'''
+
 def print_dir():
     for i in os.listdir():
         print(i)
--- a/students/alexLaws/lesson04/mailroom.py
+++ b/students/alexLaws/lesson04/mailroom.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python3
+
+donor_list = []
+donor_list.append({'Name': 'Ryan Moore', 'Donation': 500})
+donor_list.append({'Name': 'Ryan Moore', 'Donation': 250})
+donor_list.append({'Name': 'Ted Laws', 'Donation': 1000})
+donor_list.append({'Name': 'Ted Laws', 'Donation': 100})
+donor_list.append({'Name': 'Ben Snell', 'Donation': 150})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 2000})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 2000})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 4000})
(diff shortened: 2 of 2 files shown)


students/alexLaws/lesson04/dict_lab.py
```diff
@@ -0,0 +1,70 @@
...
+print('cake' in person)
+print('Mango' in person.values())
+
+# Dictionaries 2
+
+num_Ts = {}
+for i in person:
+    num_Ts[i] = person[i].lower().count('t')
+
+print(num_Ts)
```
natasha-aleksandrova: `num_ts` (lower case snake_case)

students/alexLaws/lesson04/dict_lab.py
```diff
@@ -0,0 +1,70 @@
...
+print(s2)
+print(s3)
+print(s4)
+
+print(s3.issubset(s2))
+print(s4.issubset(s2))
+
+# Sets 2
+
+char_Set = set("python")
```
natasha-aleksandrova: `char_set` (lower case snake_case)

students/alexLaws/lesson04/dict_lab.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def print_dir():
+    for i in os.listdir():
+        print(i)
+
+
+def copy_file(file, dest=None):
+    buffer = 256
+    copy = "{}\{}_copy{}".format(dest, file[:-4], file[-4:])
```
natasha-aleksandrova: be careful about overwriting built-in functions, as you can see github here highlights `file` in orange color

you can call it `source_file`

natasha-aleksandrova: preferred to use os.path.join instead of manually creating file path by using `\` (windows) or `/` (unix)

students/alexLaws/lesson04/dict_lab.py
```diff
@@ -0,0 +1,70 @@
...
+    f = open(file, 'rb')
+    j = open(copy, 'wb')
+    while True:
+        temp = f.read(buffer)
+        if len(temp) > 0:
+            j.write(temp)
+        else:
+            break
+    j.close()
```
natasha-aleksandrova: doesn't look like you're closing `f` with `f.close()`

it is preferred to use `with` statement when working with files which handles auto closing the file handle for you.

example:
```
with open(filename, 'r') as f:
    f.read(buffer)
```

natasha-aleksandrova: nice work on dict and file labs!

students/alexLaws/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+def user_action():
+    print("Choose an action from this list:")
+    print("Send a Thank You")
+    print("Create a Report")
+    print("Quit")
+    while True:
+        option = input("What is your choice: ")
+        if option.lower() in ["send a thank you", "create a report", "quit"]:
+            return option
+            break
```
natasha-aleksandrova: no need for `break` here, your `return` will act like a break

students/alexLaws/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+donor_list = []
+donor_list.append({'Name': 'Ryan Moore', 'Donation': 500})
+donor_list.append({'Name': 'Ryan Moore', 'Donation': 250})
+donor_list.append({'Name': 'Ted Laws', 'Donation': 1000})
+donor_list.append({'Name': 'Ted Laws', 'Donation': 100})
+donor_list.append({'Name': 'Ben Snell', 'Donation': 150})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 2000})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 2000})
+donor_list.append({'Name': 'Andrew Crawford', 'Donation': 4000})
+donor_list.append({'Name': 'Beth Ross', 'Donation': 400})
```
natasha-aleksandrova: I'm going to review below code as is.. but I would really recommend to reconsider the data structure chosen here which will change a lot of your code.

The idea to use dict for donor DB is to make things much easier and because you are still dealing with a list it is not easier.

here are two options: 
```
donors = {
    "Ryan Moore": [234, 333, 234343],
    "Ted Laws": [444, 100, 300]
}

donors = {
    "Ryan Moore": {"donations": [234, 333, 234343]},
    "Ted Laws": {"donations": [444, 100, 300]}
}
```

first option is bare minimum for your needs and simple, second is an extendable version in case you would like to associate more data to your donor.

notice that donor name is a natural key, look up by key is very cheap and you are not duplicating data

once you go with a `dict` for your data structure, next check out `defaultdict` and how it can help you here 🙂

students/alexLaws/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+    while True:
+        option = input("What is your choice: ")
+        if option.lower() in ["send a thank you", "create a report", "quit"]:
+            return option
+            break
+        else:
+            print("That is not an option. Choose again.")
+
+
+def dnrs(x):
```
natasha-aleksandrova: plz don't use x for var names 🙂 you can use short hand var names for comprehension loops where it is throw away and clear what you are doing.

this tells me no indication what `x` is and i had to trace down the code to find out

natasha-aleksandrova: function would probably be better called `get_donor_names` ?

students/alexLaws/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+    names = []
+    for dnr in x:
+        person = dnr.get("Name")
+        if person in names:
+            pass
+        else:
+            names.append(person)
+    return names
```
natasha-aleksandrova: you can utilize a `set` here to clean it up:
```
names_set = set()
for donor in donor_list:
    names_set.add(donor['Name'])
return list(names_set)
```

much simpler!

students/alexLaws/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+                thank_you(name, donor_list)
+            else:
+                amt = int(input("Wait, A new donation! How much: "))
+                donor_list.append({'Name': name, 'Donation': amt})
+                thank_you(name, donor_list)
+        if act == "create a report":
+            report(donor_list)
+        if act == "quit":
+            break
+        print("Would you like to do something else...")
```
natasha-aleksandrova: here are some general comments about the "main" execution block:

* you don't really need to pass around `donor_list` into functions since it is defined globally
* `if act` on lines 106, 114, 116 should really be `if .. elif.. elif` it will only ever go into one of them
* for user action it is much more user friendly to have user type a number, options 1, 2 and 3 and in your code check for that instead of full strings (error prone)
* consider implementing: `See if you can use a dict to switch between the users selections` per course instructions. I will give example below. That would make the code  much cleaner in the main block.


Here is the example of user selection switch using dict (note I am using numbers to switch on user selection for simplicity), which would be inside while True loop:
```
selection_map = {
    "1": send_thank_you,
    "2": create_report,
    "3": quit  # or sys.exit
}
user_selection = input("select ... ")
if user_selection not in selection_map:
    print("bad selection!")
selection_map[user_selection]() 
```

natasha-aleksandrova:
@alexLaws I don't see your submission on the course website, please make the submission when you get a chance. Merging this PR now.

PR #185, opened 2018-02-23
Emmanuel Twum - updated mailroom program code for lesson 3

etwum:
made some updates to the mailroom program based on feedback

### DIFF
--- a/students/etwum/lesson03/mailroom.py
+++ b/students/etwum/lesson03/mailroom.py
@@ -1,13 +1,12 @@
 #!/usr/bin/env python3
 
-if __name__ == "__main__":
-    print()
+
 
 # headers used in table
-lstHeader = [["Donor Name", "| Total Donation(s)", "| # of Donations", "| Avg Donation"]]
+lst_Header = [["Donor Name", "| Total Donation(s)", "| # of Donations", "| Avg Donation"]]
 
 # initial list of donors
-lstDonors = [["Lionel Messi", 1000000.00, 5, 20000.00],["Thierry Henry", 500, 1, 500], ["Michael Jordan", 45000, 3, 15000],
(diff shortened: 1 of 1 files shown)


students/etwum/lesson03/mailroom.py
```diff
@@ -1,28 +1,23 @@
...
 
-if __name__ == "__main__":
-    print()
-
-# headers used in table
-lstHeader = [["Donor Name", "| Total Donation(s)", "| # of Donations", "| Avg Donation"]]
 
 # initial list of donors
-lstDonors = [["Lionel Messi", 1000000.00, 5, 20000.00],["Thierry Henry", 500, 1, 500], ["Michael Jordan", 45000, 3, 15000],
+lst_Donors = [["Lionel Messi", 1000000.00, 5, 20000.00],["Thierry Henry", 500, 1, 500], ["Michael Jordan", 45000, 3, 15000],
```
natasha-aleksandrova: per pep8: your regular var names should be all lower case `lst_donors` and `lst_header`

students/etwum/lesson03/mailroom.py
```diff
@@ -34,9 +29,21 @@ def send_thank_you():
...
+            print('-----------------------------------------------------')
+            print()
+            if str_donation_choice == 'yes':
+                donate_more = float(input("How much would you like to donate? "))
+                total_donations = x[1] + donate_more
+                count_donations = x[2] + 1
+                avg_donation = total_donations/count_donations
+                x[1] = total_donations
+                x[2] = count_donations
+                x[3] = '{:.2f}'.format(avg_donation)
```
natasha-aleksandrova: something like `donor` instead of `x` would be a better name for a var for whoever is reading your code 🙂

PR #186, opened 2018-02-23
Added funcionality to kata_script

kuhnbt7:


students/kuhnbt/lesson4/kata_script.py
```diff
@@ -40,4 +40,22 @@ def generate_text(trigrams_dict, length):
                                       len(next_word_choices)-1)]
         start_key = (start_key[1], next_word)
         results.append(next_word)
-    return ' '.join(results)
+        # Lines tend to be about 7 words long
+        reshaped_results = []
+        for i, j in enumerate(results):
+            if i>0 and i % 7 == 0:
```
robalford: Add space around the `>` for Pep8, but don't think you actually need the `if i > 0` check here

natasha-aleksandrova:
👍

PR #187, opened 2018-02-23
code with only slicing

JeroenSweerts:
Hi,

As suggested after reviewing my code, I rewrote the code applying only slicing.
Could you please review?

Thanks.

Jeroen

### DIFF
--- a/students/JeroenSweerts/Lesson3/Slicing_Lab_Exercise.py
+++ b/students/JeroenSweerts/Lesson3/Slicing_Lab_Exercise.py
@@ -7,19 +7,7 @@ food = ['spam', 'eggs', 'ham']
 
 def exchange_first_last(seq):
     """the first and last items exchanged"""
-    tipe = type(seq)
-    seq = list(seq)
-    first = seq[0]
-    last = seq[-1]
-    a_new_sequence = seq
-    a_new_sequence[0] = last
-    a_new_sequence[-1] = first
-    if tipe == str:
-        return ''.join(map(str, a_new_sequence))
-    elif tipe == tuple:
-        return tuple(a_new_sequence)
-    else:
-        return a_new_sequence
+    return seq[-1:] + seq[1:-1] + seq[:1]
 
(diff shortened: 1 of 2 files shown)


students/JeroenSweerts/Lesson3/Slicing_Lab_Exercise.py
```diff
@@ -7,19 +7,7 @@
```
natasha-aleksandrova: pep8: your python file names should be all lower case -> `slicing_lab_exercise.py`

students/JeroenSweerts/Lesson3/Slicing_Lab_Exercise.py
```diff
@@ -28,101 +16,52 @@ def exchange_first_last(seq):
...
-            del seq[i]
-        except:
-            break
-    if tipe == str:
-        return ''.join(map(str, seq))
-    elif tipe == tuple:
-        return tuple(seq)
-    else:
-        return seq
+    return seq[0:len(seq):2]
```
natasha-aleksandrova: note that you can do just `[::2]` the zero and length of sequence are default

students/JeroenSweerts/Lesson3/Slicing_Lab_Exercise.py
```diff
@@ -28,101 +16,52 @@ def exchange_first_last(seq):
...
-
-    if tipe == str:
-        return ''.join(map(str, seq))
-    elif tipe == tuple:
-        return tuple(seq)
-    else:
-        return seq
-
-assert mlf_third(range(9)) == [3, 4, 5, 6, 7, 8, 0, 1, 2]
+    return seq[int(len(seq)/3):-int(len(seq)/3)] + seq[-int(len(seq)/3):] + seq[:int(len(seq)/3)]
```
natasha-aleksandrova: I would extract this into a var `int(len(seq)/3` for readability

PR #196, opened 2018-02-25
lesson 2 submission

prgrover:


students/prgrover/lesson02/series.py
```diff
@@ -0,0 +1,88 @@
...
+    n:  Calculate up to the nth value of the Fibonacci Series.
+
+    Returns:
+    The nth value in the Fibonacci Series.
+    """
+
+    if n <= 1:
+        return n
+    else:
+        return (fibonacci(n-1) + fibonacci(n-2))
```
natasha-aleksandrova: don't need the extra parens, should just be: `return fibonacci(n-1) + fibonacci(n-2)`

PR #198, opened 2018-02-25
submitting lesson05 exercise

rmart300:


students/rmart300/lesson05/except_exercise.py
```diff
@@ -0,0 +1,64 @@
...
+
+try:
+   more_joke = more_fun(langs[0])
+except IndexError as e:
+    log.write('caught the index error' + str(e))
+    try:
+        next_joke = more_fun(langs[1])
+        #print(next_joke)
+    except Exception:
+        log.write(e)
```
robalford: It's generally a better idea to catch specific exceptions that you know might occur and that you know how to handle rather than catching the base Exception object. The danger in catching the base object (or using bare `except`) is that there are so many different unknown exceptions/errors that could be slipping by silently and make it harder to debug. 

But I believe what you're trying to do here is log Exception in order to identify any unanticipated errors, and that's a valid approach as well, but this should be:

```
    except Exception as e:  
        log.write(e) 
```
FYI the [logging](https://docs.python.org/3/library/logging.html#logging.Logger.exception) module in the Python Standard Library provides an `exception` method for just this purpose.

students/rmart300/lesson05/except_exercise.py
```diff
@@ -0,0 +1,64 @@
...
+    else:
+        last_joke = more_fun(langs[2])
+finally:
+    #pass
```
robalford: Remove dead code

students/rmart300/mailroom/mailroom.py
```diff
@@ -1,18 +1,20 @@
...
 donation_dict = {}
 donor_dict = {}
 
 def is_number(n):
+    """ tests conversion to float """
     try: 
         float(n)
         return True
-    except:
+    except ValueError as e:
```
robalford: No need to assign `e` here since you're not using it in the `except` block

students/rmart300/mailroom/mailroom.py
```diff
@@ -26,15 +28,24 @@ def send_thank_you():
...
     
-    amount = float(input("Provide a donation amount:"))
-    amount_list = donation_dict[name] if name in donation_dict else [] 
-    amount_list.append(amount)
-    donation_dict[name] = amount_list    
-    donor_dict[name] = { 'firstname': name.split(' ')[0], 'lastname': name.split(' ')[1] }
-    thank_you_string = f"Hi {name}\nThank you for your donation of {amount} to the mailroom!\n"
-    print(thank_you_string)    
+    try:
+        amount = float(input("Provide a donation amount:"))
```
robalford: Can you use your `is_number` function here?

students/rmart300/mailroom/mailroom.py
```diff
@@ -26,15 +28,24 @@ def send_thank_you():
...
+    except ValueError as e:
+        print('invalid donation amount: ' + str(e))
+    else:
+        try:
+            donor_dict[name] = { 'firstname': name.split()[0], 'lastname': name.split()[1] }
```
robalford: I would just do:

```
try:
    name.split()[1]
```
Since that's the code that might raise the IndexError and then move the `donor_dict` assignment into the `else` block below.

students/rmart300/mailroom/mailroom.py
```diff
@@ -45,38 +56,45 @@ def create_report():
...
+            message += f"It will be put to very good use.\n\n    Sincerely,\n"
+            message += f"        -The Team"
+            with open(f"{letter_directory}{first_name}_{last_name}_{donation_num}.txt",'w') as f:
+                f.write(message)
+            donation_num += 1
 
 if __name__ == '__main__':
-    #build initial dictionary of donors and donation amounts
-    donor_list = ['Fred Smith','Terrie Ann','Murray Martin','Josh Jones','Jane Doe']
+    # build initial dictionary of donors and donation amounts
```
robalford: I would recommend moving all your initial data structure setup logic to the top of the file where you define the data structures.

I think this would make your code more readable, and because the `if __name__ == '__main__':` block only executes when the file is run as a program, this would allow the module to be imported along with the initial data that you need to call your functions.

PR #199, opened 2018-02-25
Herela: Trigrams

harvey-herela:


students/HarveyHerela/lesson04/trigram.py
```diff
@@ -0,0 +1,211 @@
...
+                new_lines = True
+            else:
+                if new_lines:
+                    add_word_to_trigram('\n\n')
+                new_lines = False
+                
+            for w in words:
+                add_word_to_trigram(w)
+
+            line = next(line_it)
```
natasha-aleksandrova: I think this could be simplified to:
```
with open(book_filename, 'r') as book:
    for line in book:
        if line[:3] == "***":
            continue
        else:
            process_line(line)
```

PR #200, opened 2018-02-25
exception exercise

JerHsi:


students/JerryH/Lesson05/except_exercise.py
```diff
@@ -0,0 +1,62 @@
...
+from except_test import fun, more_fun, last_fun
+
+
+# Figure out what the exception is, catch it and while still
+# in that catch block, try again with the second item in the list
+first_try = ['spam', 'cheese', 'mr death']
+
+try:
+    joke = fun(first_try[0])
+    joke2 = fun(first_try[1])
```
robalford: `joke2` should be assigned in the `except` block after catching the `NameError`

students/JerryH/Lesson05/except_exercise.py
```diff
@@ -0,0 +1,62 @@
...
+# Finally, while still in the try/except block and regardless of whether
+# there were any exceptions, call the function last_fun with no
+# parameters. (pun intended)
+
+langs = ['java', 'c', 'python']
+
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    print()
```
robalford: In order to print the error, you need to assign it to a variable using the following syntax:

```
except IndexError as e:
     print(e)
```

students/JerryH/Lesson05/except_exercise.py
```diff
@@ -0,0 +1,62 @@
...
+
+try:
+    next_joke = more_fun(langs[1])
```
robalford: This should be moved to the `except` block above along with the `else` and `finally` blocks below, in other words these should all occur within the scope of a single try statement.

students/JerryH/Lesson05/except_exercise.py
```diff
@@ -0,0 +1,62 @@
...
+except:
```
robalford: Try to avoid using a bare `except` statement, since there are so many possible exceptions that could arise it's better to hone in on the specific errors/exceptions that you know might occur.

PR #204, opened 2018-02-26
Mailroom 04

alexLaws:
Cleanup from Mailroom 3 and Mailroom 04 updates


students/alexLaws/lesson04/mailroom.py
```diff
@@ -1,118 +1,92 @@
...
-    recipient = input("Who is the thank you note for? Enter a full name or 'list' to see a list: ")
+def send_to():
+    recipient = input("Who is the thank you note for?\n"
+                      "Enter a full name or 'list' to see a list: ")
     if recipient.lower() == 'list':
         print("Here are all the donors:")
-        for name in donors:
-            print(name)
+        for i in donor_dictionary:
+            print(i)
```
robalford: This can be written without `else` statement as:
```
    if recipient.lower() == 'list':
        print("Here are all the donors:")
        for i in donor_dictionary:
            print(i)
        recipient = input("Who is the thank you note for? ")
    return recipient
```

alexLaws:
@robalford I think I do that in lines 47-51 of the code? I didn't think we needed to have a way to explicitly add a new donation.  I thought we only needed to add if they user entered a new name.

robalford:
@alexLaws the requirement I'm referring to is in the 'Sending a Thank You' section of the Mailroom Part 1 assignment from lesson 3 on the course website:

* If the user types a name in the list, use it.
* Once a name has been selected, prompt for a donation amount.
* Turn the amount into a number – it is OK at this point for the program to crash if someone types a bogus amount.
* Once an amount has been given, add that amount to the donation history of the selected user.

alexLaws:
Yup, I see that now.  Thanks for flagging that.  I remember thinking it was weird the assignment didn't provide a way to add new donations to existing donors.  I'll add it in the next round.

robalford:
@alexLaws 👍

PR #208, opened 2018-02-26
first commit for String Formatting Exercise

JeroenSweerts:
Hi, 

Please review my code for the String Formatting Exercise.

Thanks.

Best regards,

Jeroen

### DIFF
--- a/students/JeroenSweerts/Lesson3/string_formatting_exercise.py
+++ b/students/JeroenSweerts/Lesson3/string_formatting_exercise.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+
+###################TASK 1#######################################################
+'''Write a format string that will take the following four element tuple:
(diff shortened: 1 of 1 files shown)


students/JeroenSweerts/Lesson3/string_formatting_exercise.py
```diff
@@ -0,0 +1,62 @@
...
+
+###################TASK 3#######################################################
+def formatter(*t):
+    form_string = ""
+    for el in t[0]:
+        form_string = form_string + "{:d},"
+    #delete last comma of form_string
+    form_string = form_string[:-1]
+
+    print("the "+str(len(*t))+" numbers are: " +form_string.format(*t[0]))
```
natasha-aleksandrova: I would generally try to stick with one type of string concat, it makes it a bit harder to read with different flavors.

either ony f-string or format would work well here.

PR #210, opened 2018-02-27
lesson04 - ungraded

ghassan24601:


students/ghassan/lesson04/dict_lab.py
```diff
@@ -0,0 +1,45 @@
...
+}
+
+print(cakes)
+cakes.pop('cake')
+print(cakes)
+cakes['fruit'] = 'Mango'
+print(cakes)
+print(cakes.keys())
+print(cakes.values())
+print('cake' in cakes.keys())
```
robalford: This can be `'cake' in cakes` (`in` checks for existence of key in dict without needing to call `keys()`

students/ghassan/lesson04/dict_lab.py
```diff
@@ -0,0 +1,45 @@
...
+
+# sets1
+s2 = set(s for s in range(0, 21) if s % 2 == 0)
+s3 = set(s for s in range(0, 21) if s % 3 == 0)
+s4 = set(s for s in range(0, 21) if s % 4 == 0)
+print(s2)
+print(s3)
+print(s4)
+print(s3.issubset(s2))
+print(s4.issubset(s2))
```
robalford: 👍

PR #211, opened 2018-02-27
Submitting dict_lab.py

tricaonguyen:
I might have misunderstood the requirements of Dictionaries 2.


students/tri_nguyen/lesson04/dict_lab.py
```diff
@@ -0,0 +1,139 @@
...
+    if 'cake' in a_dict:
+        print('\nYes, there is cake.')
+    else:
+        print('\nNo, there is no cake')
+
+    # Check if "Mango" is a value
+
+    value_list = list(a_dict.values())
+
+    if 'Mango' in value_list:
```
robalford: `dict.values()` returns a list of values so you can just loop through that without the `value_list` assignment: `if 'Mango' in a_dict.values():`

students/tri_nguyen/lesson04/dict_lab.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def dict_two():
+    ''' practice Python dictionary '''
+
+    a_dict = {'name': 'Tri', 'city': 'Seattle', 'cake': 'cheese'}
+
+    new_dict = {'name': a_dict['name'] + ('t'.upper() * 3),
+                'city': a_dict['city'] + ('t' * 3),
+                'cake': a_dict['cake'] + ('t'.upper() * 3)}
```
robalford: You're looking for the `count` of `t`'s in each value here

PR #212, opened 2018-02-27
[DennisLee][Lesson 03][Mailroom Lab P1] Ready for submission.

DennisLee-DennisLee:


students/DennisLee/lesson03/mailroom.py
```diff
@@ -0,0 +1,139 @@
...
+        send_thank_you()  # Try getting a donor name again
+
+    else:
+        if response not in donors:  # Add to donor list if it's a new name
+            donors.append(response)
+            donor_history.append([response])
+
+        donation = '...'  # Seed donation variable to continue the while loop
+        # Make sure the donation amount is a valid, positive number
+        while donation.count('.') > 1 or donation.strip('0123456789.'
```
robalford: It could make this section a bit more readable to just use an empty string for the donation initialization and then `while not donation`, or use some variation of `while True` with the number validation code in an `if` block, but that's just a suggestion.

DennisLee-DennisLee: I had a lot of trouble with this section, as first I tried to use **.isnumeric()** and **.isdecimal()**, but those methods kept returning **False** whenever I tested a value with cents in it. The last part of the **while** statement was to ensure that **float(donation)** was positive, but that expression errored out if I set **donation** to `''`, so I gave in and initialized donation to a value that would flip the large Boolean expression to **True** on the first clause.

I wish the **while True** pathway had occurred to me. :( It would've saved me a headache to just exit the loop using a **break** statement by writing a conditional to look for the correct characteristics, instead of filtering out the incorrect stuff.

robalford: This section is tricky to do well, especially with the toolkit available at this point in the course. We'll be adding exceptions to mailroom shortly which will help with validating for different forms of input. The `math` module of the Python Standard Library includes some methods that might be helpful to you here as well.

PR #214, opened 2018-02-27
Lesson 4 Ungraded Activities

csdotson:
Not sure if we needed to open a PR to commit the Activities for Lesson 4, but figured I'd add them in. If you get a moment, any feedback would be helpful. Thanks!


students/csdotson/lesson04/dict_lab.py
```diff
@@ -0,0 +1,71 @@
...
+print(d.values())
+
+print("Is 'cake' a key in the dictionary?", 'cake' in d)
+print("Is 'Mango' a value in the dictionary?", 'Mango' in d.values())
+print()
+
+### Dictionaries 2 ###
+print("*" * 5 + " Dictionaries 2 " + "*" * 5)
+print("Dictionary from part 1:", d)
+for key in d.keys():
```
natasha-aleksandrova: note that you don't have to call `keys()`. `for key in d:` will iterate over keys

students/csdotson/lesson04/file_lab.py
```diff
@@ -0,0 +1,16 @@
...
+    print(f'File {index + 1}: ', f)
+
+# Copy files from a source to a destination
+source_file = 'file_to_move.txt'
+destination = pathlib.Path.home()
+outfile = open(destination / 'output.txt', 'w')
+
+for line in open(source_file):
+    outfile.write(line)
+outfile.close()
```
natasha-aleksandrova: couple notes here:

 * prefer to use `os.path.join` to create file path ([read here](https://stackoverflow.com/questions/13944387/why-use-os-path-join-over-string-concatenation))
* use `with` when opening files, here you're not closing source_file and if write crashes, outfile will be not be closed as well.

PR #218, opened 2018-03-01
Lesson04 (minus the trigrams)

MikeS-1:


students/MikeShand/lesson04/dict_lab.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python
+
+dictionary = {"name": "Chris", "city": "Seattle", "cake": "chocolate"}
+
+print(dictionary)
+
+dictionary.popitem()
```
natasha-aleksandrova: I would probably get used to using `pop()` vs `popitem()`, 

`pop` can do everything `popitem` can, plus more

students/MikeShand/lesson04/dict_lab.py
```diff
@@ -0,0 +1,56 @@
...
+
+print(dictionary)
+
+dictionary["fruit"] = "Mango"
+
+print(dictionary.keys())
+
+print(dictionary.values())
+
+print("cake" in dictionary.keys())
```
natasha-aleksandrova: note that you can also do `"cake" in dictionary` without calling `keys()` as it will iterate over keys by default

students/MikeShand/lesson04/file_lab_list.py
```diff
@@ -0,0 +1,4 @@
+import pathlib
+
+for file in pathlib.WindowsPath().rglob("*"):
+    print(file.absolute())
```
natasha-aleksandrova: 1. don't use `file` as a var name, you can see github highlights it differently that is because it is a built-in function
2. how could you make it platform independant?

students/MikeShand/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,89 @@
...
+            batch_file()
+
+        elif cmd in ('4', 'q', 'qqq'):
+            quit_console = True
+
+        # elif cmd.lower()=='qqq':
+            # quit_console=True
+
+        else:
+            print('Please try again')
```
natasha-aleksandrova: can you use a dict to switch between the users selections?

students/MikeShand/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,89 @@
...
+            thank_you()
+
+        elif cmd == '2':
+            donor_report()
+
+        elif cmd == '3':
+            batch_file()
+
+        elif cmd in ('4', 'q', 'qqq'):
+            quit_console = True
```
natasha-aleksandrova: there are various flavors on doing the while loop.. in your particular case based on your code here it seems simpler to call `break` here and do a `while True`

students/MikeShand/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env python3
+
+# donors = [['Andy', [10.00]], ['Bill', [15.00, 25.00]], ['Chuck', [20.00, 30.00, 40.00]]]
+
+donors = {'Andy': [10.00], 'Bill': [15.00, 25.00], 'Chuck': [20.00, 30.00, 40.00]}
```
natasha-aleksandrova: 👍

students/MikeShand/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,89 @@
...
+        else:
+            donors[new_donor] = [amount]
+
+        print('Thank you {} for your generous donation of ${:.2f}'.format(new_donor, amount))
+
+
+def donor_report():
+    print('Here is a list of donors and contributions')
+    print('Name                 Total                Donations            Average')
+    for data in donors:
```
natasha-aleksandrova: this is technically for `name in donors`
since you also need the dict values, instead of accessing by key just loop over both keys and values (a more common pattern)
`for name, data in donors.items():`

students/MikeShand/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,89 @@
...
+        print('|{:<20}|{:>20}|{:>20}|{:>20}|'.format(data, total_donation, avg_donation, total_donation / avg_donation))
+        # print(f'{data[0]:<20}'+f'${sum(data[1]):20}'+f'{len(data[1]):20}'+f'${sum(data[1])/len(data[1]):<20}')
+
+
+def batch_file():
+    for data in donors:
+        filename = data.replace(" ", "_") + ".txt"
+        total_donation = sum(donors[data])
+        letter = ('Thank you {} for you generous contributions totalling {:.2f}!'.format(data, total_donation))
+        open(filename, 'w').write(letter)
```
natasha-aleksandrova: preferred to use `with` when working with files

students/MikeShand/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,89 @@
...
+
+    else:
+
+        amount = float(input('Please enter the donation amount:'))
+
+        if new_donor in donors:
+            donors[new_donor].append(amount)
+
+        else:
+            donors[new_donor] = [amount]
```
natasha-aleksandrova: check out `defaultdict` and see how it could be useful for cases like above 🙂

MikeS-1:
https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/238

Updated mailroom with recommended changes.

PR #220, opened 2018-03-01
submitting mailroom.py

tricaonguyen:
I don't know how to make the filename for William Gates, III look like this **William_Gates_III.txt**.
Mine looks like this **William_Gates,_III.txt** with a comma.

### DIFF
--- a/students/tri_nguyen/lesson04/mailroom.py
+++ b/students/tri_nguyen/lesson04/mailroom.py
@@ -1,0 +1,175 @@
+#!/usr/bin/env python3
+
+# ------------------------------------------- #
+# Student:   Tri Nguyen
+# Instructor: Natasha Aleksandrova
+# Date:  25-Feb-2018
+# ------------------------------------------- #
+
+# Import stuff
+from operator import itemgetter
+
+# Declare variables and data structure
+
+donor_dict = [
+    {'name': 'William Gates, III', 'donation': 653784.49, 'count': 2, 'average': 326892.24},
+    {'name': 'Mark Zuckerberg', 'donation': 16396.10, 'count': 3, 'average': 5465.37},
+    {'name': 'Jeff Bezos', 'donation': 877.33, 'count': 1, 'average': 877.33},
+    {'name': 'Paul Allen', 'donation': 708.42, 'count': 3, 'average': 236.14},
+    {'name': 'Tri Nguyen', 'donation': 100.00, 'count': 1, 'average': 100.00}
+]
+    
+# Declare functions
+
+
+def display_menu():
+    ''' this function display a menu
+        of 3 actions '''
+    ACTIONS = ('Send a Thank You', 'Create a Report', 'Send letters to everyone', 'Quit')
+
+    print(''' Select one of the actions below:
+
+    1. {}
+    2. {}
+    3. {}
+    4. {}
+    '''.format(*ACTIONS))
+
+
+def display_donor_list():
+    ''' this function displays list of donors
+        from donor_dict '''
+
+    for idx, item in enumerate(donor_dict):
+        print(idx + 1, '-', donor_dict[idx]['name'])
+
+
+def send_thanks():
+    ''' this function composes an email thanking
+        the donor for their generous donation '''
+
+    name_list = [item['name'] for item in donor_dict]
+
+    while True:
+
+        full_name = input('Enter a Full Name (or type "list" to get the list of donors): ').title()
+
+        if full_name.lower() == 'list':
+            display_donor_list()
+            continue
+
+        elif full_name not in name_list:
+            donation_amount = float(input('Enter donation amount: '))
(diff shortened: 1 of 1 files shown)


students/tri_nguyen/lesson04/mailroom.py
```diff
@@ -0,0 +1,175 @@
...
+
+# Declare variables and data structure
+
+donor_dict = [
+    {'name': 'William Gates, III', 'donation': 653784.49, 'count': 2, 'average': 326892.24},
+    {'name': 'Mark Zuckerberg', 'donation': 16396.10, 'count': 3, 'average': 5465.37},
+    {'name': 'Jeff Bezos', 'donation': 877.33, 'count': 1, 'average': 877.33},
+    {'name': 'Paul Allen', 'donation': 708.42, 'count': 3, 'average': 236.14},
+    {'name': 'Tri Nguyen', 'donation': 100.00, 'count': 1, 'average': 100.00}
+]
```
natasha-aleksandrova: although this data structure can work, it is not ideal.

in this particular case `name` is our unique key for donors. consider that you need to check if donor exists to determine if you should add donations to existing one or if you need to add new donor, this is what you do:
```
for donor in donor_list:
    if donor['name'] == name_entered:
        # append to existing donor
```
1. it's way more complex to manage donors 2. it is expensive (imagine you have a large list of donors)

now if you actually used a dictionary:
```
donor_dict = {"Paul Allen": [10, 100, 120], "Jeff Bezos": [3, 44, 45]}
if donor_dict.get(name_entered):
    donor_dict[name_entered].append(new_donation)
else:
    donor_dict[name_entered] = [new_donation]
```
1. simple, no more looping looking for something in the list 2. it's *cheap*, O(1) time which means that, no matter how much data, it will execute in constant time (that is not the case if a list look up).

students/tri_nguyen/lesson04/mailroom.py
```diff
@@ -0,0 +1,175 @@
...
+        print()
+        if action == '4':
+            break
+
+        options = {
+            '1': send_thanks,
+            '2': create_report,
+            '3': send_all_letters}
+
+        options.get(action, default)()
```
natasha-aleksandrova: I like the `default` functionality!

students/tri_nguyen/lesson04/mailroom.py
```diff
@@ -0,0 +1,175 @@
...
+
+        first_name = each_list[0][0]
+        last_name = each_list[0][1]
+
+        with open(template.format(*each_list[0]), 'w') as f:
+            f.write('{0} {1} {2},\n\n'.format(email_dict['greetings'], first_name, last_name))
+            f.write('\t{0}${1:,.2f}\n\n'.format(email_dict['body_one'], each_list[1]))
+            f.write('\t{}\n\n'.format(email_dict['body_two']))
+            f.write('\t\t{}\n'.format(email_dict['end']))
+            f.write('\t\t    {}'.format(email_dict['signature']))
```
natasha-aleksandrova: I wouldn't use `email_dict` as a way to populate the file content string. 

what makes a bit more sense is to have a template at the top of the file
```
first_name = "natasha"
template = "Dear {first_name}, thank you for your donation!"
# then in the loop:
template.format(first_name=first_name)
```
the way it is right now makes it kind of hard to read

students/tri_nguyen/lesson04/mailroom.py
```diff
@@ -0,0 +1,175 @@
...
+        'end': 'Sincerely',
+        'signature': '-The Team'}
+
+    value_list = [list(item.values()) for item in donor_dict]
+
+    for each_list in value_list:
+
+        each_list[0] = each_list[0].split()
+
+        template = '{}_' * (len(each_list[0]) - 1) + '{}.txt'
```
natasha-aleksandrova: for your question on `,` in the file name:
```
>>> s = 'William_Gates,_III.txt'
>>> s.replace(',', '')
'William_Gates_III.txt'
```
you can use replace to just remove the comma, there are also other ways to do it if you have a set of chars you want to remove.

students/tri_nguyen/lesson04/mailroom.py
tricaonguyen: Thanks, Natasha. Your instructions and explanations here are very helpful. I kind of knew that there would be a better way to organize. I will make changes.

students/tri_nguyen/lesson04/mailroom.py
tricaonguyen: Having the template at the top of  the file meaning outside of the function  at the top of the script?

students/tri_nguyen/lesson04/mailroom.py
natasha-aleksandrova: 👍

students/tri_nguyen/lesson04/mailroom.py
natasha-aleksandrova: sorry I meant around line 119 outside the loop but inside of `send_all_letters` function

PR #221, opened 2018-03-01
Lesson03 incorporate suggested fixes

csdotson:
Thanks a lot for the suggestions, Natasha! I tried to incorporate everything you had mentioned. Let me know if you still see any issues.


natasha-aleksandrova:
this is great. in the next assignment you will be incorporating `dict` data structure in your mailroom which will make the code much easier and cleaner.

PR #222, opened 2018-03-01
lesson04 - new mailroom

ghassan24601:


students/ghassan/lesson04/new_mailroom.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
+
+import sys
+
+global donors
+donors = {
```
robalford: No need to declare `global` here, since `donors` is defined in the global namespace it becomes a global variable by default

students/ghassan/lesson04/new_mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+        'Jessie': [100, 300, 500],
+        'James': [343, 22.11],
+        'Ford': [3000, 400]}
```
robalford: Nice use of `dict` for you main data structure

students/ghassan/lesson04/new_mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def main_menu():
+    choice = input(
+        'Pick one:\n1) Send a thank you\n2) Create a Report\n\
+3) Send letters to everyone\n4) Quit\n')
+    return str(choice)
```
robalford: In Python3, `input` returns a string by default so no need to call `str()` here

students/ghassan/lesson04/new_mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def send_thankyou():
+    donor_name = input('Enter the donor\'s name: ')
+    if donor_name == 'list':
```
robalford: Remember to prompt the user for the 'list' option

students/ghassan/lesson04/new_mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+        print(donor.keys())
```
robalford: This should be `donors` plural and will throw a NameError as is

students/ghassan/lesson04/new_mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+    elif donor_name not in donors.keys():
+        donors[donor_name] = []
+    donation_amount = input(
+        'Please enter a donation amount: ')
+    donors[donor_name].append(donation_amount)
```
robalford: Remember to convert this to a `float` so that you can use in the report calculations

students/ghassan/lesson04/new_mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+                'Thank you {} for yout generous donation of {}\n'.format(
+                    donor, sum(amounts)))
+
+
+def quittt():
+    sys.exit()
+
+
+def main():
+    users_choice = main_menu()
```
robalford: The program needs to return to the main prompt and continue running after exiting each selection. A `while` loop is a great construct to use for this purpose.

students/ghassan/lesson04/new_mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+
+def main_menu():
+    choice = input(
+        'Pick one:\n1) Send a thank you\n2) Create a Report\n\
+3) Send letters to everyone\n4) Quit\n')
+    return str(choice)
+
+
+def send_thankyou():
+    donor_name = input('Enter the donor\'s name: ')
```
robalford: I would recommend using a `while` loop here, so that you can return to the main prompt after showing the list of donor names

PR #223, opened 2018-03-01
Adding slice.py for Lesson03

sbolsen:


students/sbolsen/lesson03/slice.py
```diff
@@ -0,0 +1,33 @@
+def first_last_exchanged(sequence):
+    sequence[0], sequence[-1] = sequence[-1], sequence[0]
+    return sequence
```
natasha-aleksandrova: this is technically not slicing 🙂 

slicing is `[<start>:<stop>:<step>]`

so how could you do this with slicing only? (hint: it's a one liner)

students/sbolsen/lesson03/slice.py
```diff
@@ -0,0 +1,33 @@
...
+
+def thirds(sequence):
+    a_list = []
+    for i in sequence:
+        a_list.append(i)
+    section = int(len(a_list) / 3)
+    first = a_list[:section]
+    third = a_list[section * 2:section * 3]
+    second = a_list[section:section * 2]
+    return second + third + first
```
natasha-aleksandrova: same here 🙂 

how can you do it with slicing only?

students/sbolsen/lesson03/slice.py
```diff
@@ -0,0 +1,33 @@
...
+
+def every_other(sequence):
+    return sequence[1::2]
```
natasha-aleksandrova: this is technically: `sequence[::2]` (first one should be zero but we omit because it's the default)

students/sbolsen/lesson03/slice.py
```diff
@@ -0,0 +1,33 @@
...
+
+def four_and_four(sequence):
+    b = sequence[4:-4]
+    c = b[1::2]
+    return c
```
natasha-aleksandrova: in one line: `return [4:-4:2]`

PR #224, opened 2018-03-01
first commit for mailroom exercise

JeroenSweerts:


students/JeroenSweerts/Lesson3/mailroom.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
+
+donors = ['papa', 'mama', 'bompa', 'bobonne', 'onbekende']
+donations = [[100, 5, 15], [12, 200, 2, 66], [1000], [500, 500], [1000000]]
```
robalford: Can you think of a way to combine these into a single data structure? Then you wouldn't need to keep the indexes of the two lists in sync and this would simplify some of your lookup code below.

students/JeroenSweerts/Lesson3/mailroom.py
```diff
@@ -0,0 +1,64 @@
...
+
+def thankyou():
+    '''If the user types ‘list’, show them a list of the donor names and re-prompt
```
robalford: Use triple double-quoted strings for docstrings: `"""`

students/JeroenSweerts/Lesson3/mailroom.py
```diff
@@ -0,0 +1,64 @@
...
+    Once a name has been selected, prompt for a donation amount.
+    Turn the amount into a number – it is OK at this point for the program to
+    crash if someone types a bogus amount.
+    Once an amount has been given, add that amount to the donation history of
+    the selected user.
+    Finally, use string formatting to compose an email thanking the donor for their
+    generous donation. Print the email to the terminal and return to the original
+    prompt.'''
+    name = input('Please enter a name or type "list" to see a list of names > ')
+    while name == 'list':
```
robalford: In Python, you can initialize an infinite loop by using the syntax `while True:` for these kinds of situations and then use `break` to exit the loop whenever your exit condition is met. 

Or you can set up a Boolean variable such as `asking_for_name = True` for a more readable version of the same construct in `while asking_for_name`. 

One of the advantages of using an approach like this here is that you wouldn't need to repeat the same call to `input` at various points in your code. So the whole block would look something like this:

```
    while True:
        name = input('Please enter a name or type "list" to see a list of names > ')
        if name == 'list':
            print(donors)
        else:
            break
```

students/JeroenSweerts/Lesson3/mailroom.py
```diff
@@ -0,0 +1,64 @@
...
+        print(donors)
+        name = input('Please enter a name or type "list" to see a list of names > ')
+    if not name in donors:
```
robalford: This should be `if name not in donors:` to check for membership in list

students/JeroenSweerts/Lesson3/mailroom.py
```diff
@@ -0,0 +1,64 @@
...
+        donors.append(name)
+        donations.append([])
+    donation = input('Please enter a donation amount > ')
+    donations[donors.index(name)].append(donation)
+    print(f'Dear {name} , thank you very much for your generous donation of\
+    {donation}. Looking forward receiving even more money next time.')
+    print(donors)
+    print(donations)
+
```
robalford: Use two blank lines between functions for Pep8

students/JeroenSweerts/Lesson3/mailroom.py
```diff
@@ -0,0 +1,64 @@
...
+     those above and below)
+     After printing this report, return to the original prompt.
+     At any point, the user should be able to quit their current task and return
+     to the original prompt.
+     From the original prompt, the user should be able to quit the script cleanly.'''
+    print(f'{"Donor Name":20s} {"|  Total Given":20s} {"|  Num Gifts  |":20s} {"Average Gift":20s}')
+    print(f'{"-"*80}')
+    for name in donors:
+        print(f'{name:20s} ${sum(donations[donors.index(name)]):20.2f}\
+        {len(donations[donors.index(name)]):12d}\
```
robalford: Can you adjust the formatting here to match the widths of row value cells above?

students/JeroenSweerts/Lesson3/mailroom.py
```diff
@@ -0,0 +1,64 @@
...
+        ${sum(donations[donors.index(name)])/len(donations[donors.index(name)]):20.2f}')
+
+if __name__ == '__main__':
+    response = input('Please choose between the following 3 actions:\
+    1. Send a Thank You; 2. Create a Report; 3. quit >  ')
+    while response != '3':
```
robalford: See my comment above on using `while True` or a Boolean variable to set up an infinite loop and simplify some of the code here.

students/JeroenSweerts/Lesson3/mailroom.py
```diff
@@ -0,0 +1,64 @@
...
+    generous donation. Print the email to the terminal and return to the original
+    prompt.'''
+    name = input('Please enter a name or type "list" to see a list of names > ')
+    while name == 'list':
+        print(donors)
+        name = input('Please enter a name or type "list" to see a list of names > ')
+    if not name in donors:
+        donors.append(name)
+        donations.append([])
+    donation = input('Please enter a donation amount > ')
```
robalford: Remember to convert the user input from a string to the correct type here. The program currently crashes at line 51 below with the following error message:

```
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

PR #225, opened 2018-03-02
Lesson 05 - Emmanuel Twum - Mail Room Part 3

etwum:


students/etwum/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env python3
+
+
+# initial list of donors
+lst_donors = [{"donor name": "Lionel Messi", "total donations": 1000000.00, "number donations": 5, "avg donation": 20000.00},
+              {"donor name": "Thierry Henry", "total donations": 500, "number donations":1, "avg donation": 500},
+              {"donor name": "Michael Jordan", "total donations": 45000, "number donations": 3, "avg donation": 15000},
+              {"donor name": "Kobe Bryant", "total donations": 8000, "number donations": 2, "avg donation": 4000}]
```
natasha-aleksandrova: this is not the appropriate data structure for this particular use case

for performance considerations and usability, you should be using a dictionary not a list

students/etwum/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,178 @@
...
+    for x in lst_donors:
+        if x["donor name"] == donor_full_name:
+
+            while True:
+
+                # raise exception if the user inputs anything other than 'yes' or 'no'
+                try:
+                    str_donation_choice = input("Would you like to add another donation? (yes or no) ")
+                    if str_donation_choice not in ("yes","no"):
+                        raise Exception
```
natasha-aleksandrova: adding the try except here seems a little extra, could have been done simply with if/else which is already there

students/etwum/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,178 @@
...
+                except Exception:
+                    print("Please input 'yes' or 'no'")
+                else:
+                    break
+
+            print('-----------------------------------------------------')
+            print()
+            if str_donation_choice == 'yes':
+                donate_more = float(input("How much would you like to donate? "))
```
natasha-aleksandrova: this would be a good place for try/except

students/etwum/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,178 @@
...
+            # input donor name to print an email
+            donor_first_name = input("Please input the first name of the donor. ")
+            donor_last_name = input("Please input the last name of the donor. ")
+            if donor_first_name == "" or donor_last_name == "":
+                raise Exception
+        except Exception:
+            print("\nPlease input a first and last name")
+            print()
+        else:
+            break
```
natasha-aleksandrova: a simple if/else statement here would be simpler

PR #226, opened 2018-03-02
lesson04: mailroom.py part2

22e88:


students/elmar_m/lesson04/mailroom.py
```diff
@@ -0,0 +1,154 @@
...
+'''
+
+import time
+import collections
+
+
+main_p = '\n> Main menu. Options: t == thankyou | r == report | q == quit program\n'
+sub_p = '\n>> Sub menu "thankyou". Options: l == list current donors | a == add donation / donor | m == send mail to every donor | x == exit submenu\n'
+
+donors = collections.defaultdict(list)
```
natasha-aleksandrova: 👍 on using default dict

students/elmar_m/lesson04/mailroom.py
```diff
@@ -0,0 +1,154 @@
...
+
+def writefile(name, content):
+    '''
+    Create mail files
+    
+        ARGS:
+    name: filename / name of donor
+    content: mail body text
+    '''
+    ts = time.strftime('%Y%m%d-%H%M%S')
```
natasha-aleksandrova: nice!

students/elmar_m/lesson04/mailroom.py
```diff
@@ -0,0 +1,154 @@
...
+    fname = '{}.{}.txt'.format(name, ts)
+    f = open(fname, 'w')
+    f.write(content)
+    f.close
```
natasha-aleksandrova: prefer to use `with` when working with file, this ensures that the file is closed

students/elmar_m/lesson04/mailroom.py
```diff
@@ -0,0 +1,154 @@
...
+    Add new donation and / or donor
+    '''
+    dname = input('>> Please give donor name: ')
+    if dname in donors:
+        print('>>', dname, 'already in list')
+        add_amount(dname)
+    elif not dname in donors:
+        print('>>', dname, 'not in list, adding it ')
+        add_amount(dname)
+
```
natasha-aleksandrova: this should really be if/else and you can take the `add_amount` outside of those blocks:
```
if dname in donors:
    print("already in list")
else:
    print("not in list")
add_amount(dname)
```

even shorter (if you wanted):
```
print(">> {dname} {msg}".format(dname=dname, msg="already in list" if dname in donors else "not in list, adding"))
add_amount(dname)

```

students/elmar_m/lesson04/mailroom.py
```diff
@@ -0,0 +1,154 @@
...
+        ARGS:
+    p:  prompt which is shown to the user
+    d:  the dispatcher dictionary which holds the menu options
+    '''
+    try:
+        while True:
+            response = input(p)
+            if d[response]() == 'exiting':
+                break
+    except KeyError:
```
natasha-aleksandrova: nice!

PR #228, opened 2018-03-03
[DennisLee][Lesson 04][Mailroom P2] Ready for submission.

DennisLee-DennisLee:


students/DennisLee/lesson04/dict_lab.py
```diff
@@ -0,0 +1,43 @@
...
+
+def dict1():
+    d = {'name': 'Chris', 'city': 'Seattle', 'cake': 'Chocolate'}
+    print(d)
+    del d['cake']
+    print(d)
+    d['fruit'] = 'Mango'
+    print(d.keys())
+    print(d.values())
+    print('cake' in d.keys())
```
robalford: The `in` operator test for key membership in dicts so this can be `print('cake' in d)`

students/DennisLee/lesson04/dict_lab.py
```diff
@@ -0,0 +1,43 @@
...
+    print('Mango' in d.values())
+
+def dict2():
+    d = {'name': 'Chris', 'city': 'Seattle', 'cake': 'Chocolate'}
+    for item in d:
```
robalford: You could do `for key, value in d` here

students/DennisLee/lesson04/mailroom.py
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/env python3
+
+import os
+import pathlib
```
robalford: You don't appear to be using `pathlib`so will want to remove unused import

students/DennisLee/lesson04/mailroom.py
```diff
@@ -0,0 +1,189 @@
...
+
+# Initial donor list and the amounts they have donated
+donor_history = {
+            'Red Herring': [65820.5, 31126.37, 15000],
+            'Papa Smurf': [210.64, 1000],
+            'Pat Panda': [55324.4],
+            'Karl-Heinz Berthold': [3545.2, 10579.31],
+            'Mama Murphy': [156316.99, 8500.3, 12054.33],
+            'Daphne Dastardly': [82]
+        }
```
robalford: Nice use of a dict for your donor data structure

students/DennisLee/lesson04/mailroom.py
```diff
@@ -0,0 +1,189 @@
...
+    :return:  None.
+    """
+    # create a dictionary of menu items, menu text, and menu caller functions
+    choices = {'1': ("Send a thank you", send_thank_you), 
+               '2': ("Create a report", create_a_report),
+               '3': ("Send letters to everyone", send_all_letters),
+               '4': ("Quit", exit_screen)}
+    
+    response = ''
+    while response != '4':  # Show menu forever until user exits
```
robalford: I think this could be slightly simplified by using `while True` for this loop as well and placing `response` assignment inside the loop, so that you only need to assign once to an empty string. But you'd also need an `exit` function that actually exits the program

students/DennisLee/lesson04/mailroom.py
```diff
@@ -0,0 +1,189 @@
...
+        # Print the menu list (with numbered choices)
+        print()
+        for i in choices:
+            print(i, choices[i][0])
+
+        # Get the selection number
+        response = ''
+        while not response in choices:
```
robalford: This should be `while response not in choices:`

students/DennisLee/lesson04/mailroom.py
```diff
@@ -0,0 +1,189 @@
...
+      ).strip()
+
+    if response.lower() in ('', 'quit'):
+        exit_screen()
+        return
+
+    elif response.lower() == 'list':
+        print("\nLIST OF DONORS:")
+        for donor in donor_history:
+            print(donor)
```
robalford: We'll be adding unit tests to mailroom in a future assignment, so you might consider factoring the donor list generation out into its own function that returns the list and calling that here to make it more testable.

students/DennisLee/lesson04/mailroom.py
```diff
@@ -0,0 +1,189 @@
...
+                return
+            # Make sure the donation amount is a valid, positive number
+            elif donation.strip('0123456789.') == '' and len(
+                    donation) > 0 and donation.count('.') <= 1 and float(
+                    donation) > 0.0:
+                break
+
+        # Add the donation to the master donor history and print the letter
+        donation = float(donation)
+        donor_history.setdefault(response, [])
```
robalford: Nice use of `setdefault()`

PR #229, opened 2018-03-03
second commit

JeroenSweerts:
Please review the recommended updates. I couldn't make it work to switch to an alternative data structure without making the code complexer and without using dictionaries. So I kept the data structures that I initially used.

Thanks.

Jeroen

### DIFF
--- a/students/JeroenSweerts/Lesson3/mailroom.py
+++ b/students/JeroenSweerts/Lesson3/mailroom.py
@@ -4,7 +4,7 @@ donors = ['papa', 'mama', 'bompa', 'bobonne', 'onbekende']
 donations = [[100, 5, 15], [12, 200, 2, 66], [1000], [500, 500], [1000000]]
 
 def thankyou():
-    '''If the user types ‘list’, show them a list of the donor names and re-prompt
+    """If the user types ‘list’, show them a list of the donor names and re-prompt
     If the user types a name not in the list, add that name to the data structure
     and use it.
     If the user types a name in the list, use it.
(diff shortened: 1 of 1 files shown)


students/JeroenSweerts/Lesson3/mailroom.py
```diff
@@ -15,12 +15,14 @@ def thankyou():
...
         name = input('Please enter a name or type "list" to see a list of names > ')
-    if not name in donors:
+        if name == 'list':
+            print(donors)
+        else:
+            break
+    if name not in donors:
         donors.append(name)
         donations.append([])
     donation = input('Please enter a donation amount > ')
```
robalford: If you convert and store all new donations as `int`s here, then you won't need to call `int()` on every donation where you generate the report below. And I would suggest using `float`, since we're dealing with money.

students/JeroenSweerts/Lesson3/mailroom.py
```diff
@@ -42,13 +45,11 @@ def report():
...
-     From the original prompt, the user should be able to quit the script cleanly.'''
+     From the original prompt, the user should be able to quit the script cleanly."""
     print(f'{"Donor Name":20s} {"|  Total Given":20s} {"|  Num Gifts  |":20s} {"Average Gift":20s}')
-    print(f'{"-"*80}')
+    print(f'{"-"*76}')
     for name in donors:
-        print(f'{name:20s} ${sum(donations[donors.index(name)]):20.2f}\
-        {len(donations[donors.index(name)]):12d}\
-        ${sum(donations[donors.index(name)])/len(donations[donors.index(name)]):20.2f}')
+        print(f'{name:20s} ${sum(int(i) for i in donations[donors.index(name)]):20.2f} {len(donations[donors.index(name)]):13d}${sum(int(i) for i in donations[donors.index(name)])/len(donations[donors.index(name)]):20.2f}')
```
robalford: I'd suggest breaking these calculations out of the format string and assigning to variables like 'total_gifts', 'average_gift' etc. for readability. Also, you can use triple quoted strings for extra long strings that span multiple lines to keep under Pep8 recommended line length. See my other comment on casting type of user input so you don't need to loop through and cast donations to `int`s here.

PR #230, opened 2018-03-03
request for feedback

johnpharmd:


students/johnpharmd/lesson04/trigrams.py
```diff
@@ -28,30 +28,86 @@
...
+    read_data = ' '.join(read_data.split('\n'))
+    read_data = ' '.join(read_data.split('--'))
+    read_data = ' '.join(read_data.split('-'))
+    read_data = ' '.join(read_data.split(', '))
+    read_data = ' '.join(read_data.split('. '))
+    read_data = ''.join(read_data.split('('))
+    read_data = ''.join(read_data.split(')'))
+    read_data = read_data.strip('.')
+    read_data = read_data.strip('  ')
+    read_data = read_data.split(' ')
```
natasha-aleksandrova: you could rewrite your join/split code into a loop using `replace` with replace char map
```
replace_map = {"": ("\r", '"'), " ": ("(", ")")}
for new_char, chars in replace_map.items():
    for c in chars:
        read_data = read_data.replace(c, new_char)
```

students/johnpharmd/lesson04/trigrams.py
```diff
@@ -28,30 +28,86 @@
...
+            s2 = read_data[i + 1]
+            st_key = (s1, s2)
+            s3 = read_data[i + 2]
+            if st_key not in st_dict:
+                st_dict[st_key] = [s3]
+            else:
+                st_dict[st_key] += [s3]
+        # count += 1
+    print(st_dict)
+# see comment line above
```
natasha-aleksandrova: the indentation is off in this function, also your comment should have one indent from function definition

students/johnpharmd/lesson04/trigrams.py
```diff
@@ -28,30 +28,86 @@
...
+        # if potential_k in st_dict:
+        new_st_lst.append(random.choice(st_dict[potential_k]))
+        potential_k = (new_st_lst[-2], new_st_lst[-1])
+        n += 1    
+    print('len(new_st_lst):', len(new_st_lst))
+    print(new_st_lst)
+
+
+clean_data()
+make_st_dict(read_data)
```
natasha-aleksandrova: prefer this pattern:
```
cleaned_data = clean_data()
make_st_dict(cleaned_data)
```

especially since you are returning it in `clean_data` already

students/johnpharmd/lesson04/trigrams.py
```diff
@@ -28,30 +28,86 @@
...
 s3 = ''
-count = 0
-st_set = set()
 st_dict = {}
+count = 0
 pron_set = set(['I', 'you', 'she', 'he', 'it', 'we', 'they'])
-
+n = 0
+new_st_lst= []
+testfile_dict = {1: 'sherlock_small.txt', 2: 'sherlock.txt'}
```
natasha-aleksandrova: normally the top of the file would only include constants, or instantiating logging module for example to make logging accessible in the whole module.

most of these should live with the function where they are used, consider readability aspect - every time you would need to go search for definition of vars instead of them being right where they are used.

students/johnpharmd/lesson04/trigrams.py
```diff
@@ -28,30 +28,86 @@
...
+    for i, word in enumerate(read_data):
+        if i + 2 < len(read_data):
+            s1 = word
+            s2 = read_data[i + 1]
+            st_key = (s1, s2)
+            s3 = read_data[i + 2]
+            if st_key not in st_dict:
+                st_dict[st_key] = [s3]
+            else:
+                st_dict[st_key] += [s3]
```
natasha-aleksandrova: looks like a good candidate for `st_dict` to be a `defaultdict` with default being a `list`

students/johnpharmd/lesson04/trigrams.py
natasha-aleksandrova: it's a good idea to rename a var when you change the data type (from str to a list):
`read_data = read_data.split(' ')` 
to reduces confusion -> `read_data_list = read_data.split(' ')`

PR #232, opened 2018-03-04
Submitting StringFormatLab.py

wesmoy17:
Submitting Lab 3 String Formatting Lab Assignment

### DIFF
--- a/students/WesleyM/Lesson3/StringFormatLab.py
+++ b/students/WesleyM/Lesson3/StringFormatLab.py
@@ -1,0 +1,52 @@
+def task_one():
+    """Converts (2, 123.4567, 10000, 12345.67) to 'file_002 :   123.46, 1.00e+04, 1.23e+04'"""
+    string = (2, 123.4567, 10000, 12345.67)
+    print("file_{:03d}: {:.2f}, {:.2e}, {:.2e}".format(string[0], string[1],string[2],string[3]))
+    return
+
+def task_two():
+    """Converts (2, 123.4567, 10000, 12345.67) to 'file_002 :   123.46, 1.00e+04, 1.23e+04'using f-strings"""
+    string = (2, 123.4567, 10000, 12345.67)
+    print(f"file_{string[0]:03d}: {string[1]:.2f}, {string[2]:.2e}, {string[3]:.2e}")
+    return
+    
+def task_three(in_tuple):
+    """Takes a user input tuple and writes in the form of 'the X numbers are: a, b, c'"""
+    l = len(in_tuple)
+    form_string = ("The {} numbers are: " + ", ".join(["{}"]*l)).format(l, *in_tuple)
+    form_string.format(*in_tuple)
+    return form_string.format(*in_tuple)
+
+def task_four(string2):
+    """Rewrites a tuple (4, 30, 2017, 2, 27) into '02 27 2017 04 30'"""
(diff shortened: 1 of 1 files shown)


students/WesleyM/Lesson3/StringFormatLab.py
```diff
@@ -0,0 +1,52 @@
+def task_one():
+    """Converts (2, 123.4567, 10000, 12345.67) to 'file_002 :   123.46, 1.00e+04, 1.23e+04'"""
+    string = (2, 123.4567, 10000, 12345.67)
```
robalford: I think it's a little confusing to name the tuple `string`, and in this case you could just pass the tuple straight to `format()` as args, since you're not changing the order of the items:

`print("file_{:03d}: {:.2f}, {:.2e}, {:.2e}".format(2, 123.4567, 10000, 12345.67))`

students/WesleyM/Lesson3/StringFormatLab.py
```diff
@@ -0,0 +1,52 @@
...
+
+def task_two():
+    """Converts (2, 123.4567, 10000, 12345.67) to 'file_002 :   123.46, 1.00e+04, 1.23e+04'using f-strings"""
+    string = (2, 123.4567, 10000, 12345.67)
+    print(f"file_{string[0]:03d}: {string[1]:.2f}, {string[2]:.2e}, {string[3]:.2e}")
+    return
+    
+def task_three(in_tuple):
+    """Takes a user input tuple and writes in the form of 'the X numbers are: a, b, c'"""
+    l = len(in_tuple)
```
robalford: Nice work! It would be a bit more readable to just use the call to `len()` in your code, rather than assigning to `l`:

`form_string = ("The {} numbers are: " + ", ".join(["{}"] * len(in_tuple))).format(len(in_tuple), *in_tuple)`

students/WesleyM/Lesson3/StringFormatLab.py
```diff
@@ -0,0 +1,52 @@
...
+    """Uses f-strings to write a message"""
+    string3 = ['orange', 1.3, 'lemon', 1.1]
+
+    print(f"The weight of an {string3[0]} is {string3[1]} and the weight of a {string3[2]} is {string3[3]}")
+    print(f"The weight of an {string3[0].upper()} is {string3[1]*1.2} and the weight of a {string3[2].upper()} is {string3[3]*1.2}")
+    return
+
+def task_six():
+    """Prints a table with Name, Age and Cost. Also given a tuple of ten consecutive numbers, it prints into a column 5 characters wide"""
+    table_data = [['Name', 'Age', 'Cost'], ['Robert', '23', '100'], ['Mark', '1000', '2.0'], ['Alice', '20', '1000.02'], ['Bob','32', '10000000'], ['Charles','4', '1'], ['Derek','59', '10000000000000'], ['Billy Bob Bo Bobby Bill Bixworth','59', '25.324']]
```
robalford: The way you write `table_data` here is Pep8 compliant, however, I find the following format more readable because it looks like a table and it has the advantage of placing each 'record' in your structure on its own line, making it easier to track changes to the data in version control:

```
table_data = [
        ['Name', 'Age', 'Cost'],
        ['Robert', '23', '100'],
        ['Mark', '1000', '2.0'],
        ['Alice', '20', '1000.02'],
        ['Bob', '32', '10000000'],
        ['Charles', '4', '1'],
        ['Derek', '59', '10000000000000'],
        ['Billy Bob Bo Bobby Bill Bixworth', '59', '25.324'],
    ]
```

PR #233, opened 2018-03-04
lesson07 - question

JerHsi:
@natasha-aleksandrova , I keep getting AttributeError: 'Element' object has no attribute 'append' even I have "append" class method defined.  Why is that?

### DIFF
--- a/students/JerryH/Lesson07/html_render.py
+++ b/students/JerryH/Lesson07/html_render.py
@@ -1,0 +1,20 @@
+#!/usr/bin/env python
+
+class Element(object):
+
+    tag = "html"
+    indent = "  "
+
+
+    def __init__(self, content=None):
+        self.content = [content] if content else []
+
+    def append(self, words):
+        self.content.append(words)
+
+'''
+el = Element("First content here")
+print(el.content)
+el.append("second sentence")
+print(el.content)
+'''
--- a/students/JerryH/Lesson07/run_html_render.py
+++ b/students/JerryH/Lesson07/run_html_render.py
@@ -1,0 +1,228 @@
+#!/usr/bin/env python
+
+"""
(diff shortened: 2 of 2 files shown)


natasha-aleksandrova:
@JerHsi what are you running to get that error?

natasha-aleksandrova:
I ran `run_html_render.py` but got `AttributeError: 'Element' object has no attribute 'render'`

JerHsi:
@natasha-aleksandrova 
This is what I got:
In [133]: %run run_html_render.py
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/Documents/Python/Self_Paced-Online/students/JerryH/Lesson07/run_html_render.py in <module>()
     36 page = hr.Element()
     37 
---> 38 page.append("Here is a paragraph of text -- there could be more of them, "
     39             "but this is enough  to show that we can do some text")
     40 

AttributeError: 'Element' object has no attribute 'append'

natasha-aleksandrova:
Since I'm not getting the same stack trace my only guess is something else is getting imported. Worth trying:
```
page = hr.Element()
print(hr.__file__)
```
after you create instance of Element print module file then open that file and very it is the correct class definition and has your append method

JerHsi:
@natasha-aleksandrova I closed my iPython session and re-opened a new one, and now it works as it should be.  Not quite sure why it didn't work before though.  Thanks anyway.

natasha-aleksandrova:
ah! iPython won't auto pick up changes in your file unless you set a magic variable.

JerHsi:
How should I do that?

natasha-aleksandrova:
follow this guide here:
https://ipython.org/ipython-doc/3/config/extensions/autoreload.html

PR #235, opened 2018-03-04
Herela: Added lesson 06

harvey-herela:


students/HarveyHerela/lesson06/mailroom4.py
```diff
@@ -0,0 +1,185 @@
...
+
+    print(num, "letters created.")
+
+
+if __name__ == "__main__":
+    # Default donor list
+    donors = dict()
+    donors[("Chase", "Zuma")] = {"donations": [25]}
+    donors[("Marshall", "Rocky")] = {"donations": [30, 60]}
+    donors[("Skye", "Rubble")] = {"donations": [100, 100, 100]}
```
robalford: I would suggest defining your donors data structure outside the `if __name__ == "__main__":` block so that you can import it into your test module.

students/HarveyHerela/lesson06/mailroom4_test.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python
+
+import datetime
+import unittest
+from unittest.mock import patch
+from unittest.mock import mock_open
```
robalford: be sure to remove any unused imports

students/HarveyHerela/lesson06/mailroom4_test.py
```diff
@@ -0,0 +1,99 @@
...
+
+import mailroom4
+
+class mailroom4_unit_test(unittest.TestCase):
```
robalford: Class names use CapWords convention in Python e.g. `Mailroom4UnitTest`

harvey-herela:
What is the correct way to test units that have input/print/open?

natasha-aleksandrova:
By using  mocks, which is what you have 🙂 

(mocking/testing `print` statements isn't very common though)

PR #237, opened 2018-03-05
lesson01

natrod22:
Had a bit of trouble getting GitHub to push my work, but now I think I've got a handle on it. 
Will be uploading Lesson 2 and 3 in the next day.


students/NatalieRodriguez/Lesson01/break_me.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+"""Syntax Error"""
+def SyntaxError() #writing the code without a colon throws a Syntax Error
+
+
+"""Attribute Error"""
+anInt = 8
+anInt.append(4)
+
```
natasha-aleksandrova: everything here is supposed to be invalid 🙂 .. but as you are starting out with the assignment, couple notes on styling which we will check in assignments:
* function names should be all lower case with underscores (called snake_case) -> `def name_error`
* var names are also snake_case -> `an_int`
* `""" """` triple quotes strings are common in doc strings for a function or multi-line strings, for comments use `# this is a comment`
* there are a lot more tedious spacing rules etc that I won't get into 😆 
* highly recommended to use something like pycharm or run the code via linter (pep8, pyflakes etc)

natasha-aleksandrova:
glad you got github working for you, please don't hesitate to reach out to us via course email or slack if you need any further help!

natasha-aleksandrova:
make sure to submit the PR link in the course website at the same time so that it can get graded.

you will want to submit a link like this: `https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/237` which is created right after you hit Create Pull Request button.

PR #239, opened 2018-03-05
Lesson04 - Trig

ghassan24601:


ghassan24601:
Just a feedback - and it might be only me -. 
I kind of feel like the phrasing of the assignments are not always intuitive.

robalford:
@ghassan24601 if you can point us to specific places in the assignments where you think the requirements are unclear or underspecified, we can put in a request to the course authors to revise the materials. 

As instructors, we want to this course to be as accessible and easy to navigate as possible for our students so we're definitely open to this kind of feedback and have already made a few changes like this based on student input.

PR #240, opened 2018-03-05
submitting trigrams.py

tricaonguyen:
Hi Nathasha or Rob,

For some unknown reason, I found this exercise difficult. It took me a long time, and it still not perfect.
Is there a cleaner, better way to accomplish this exercise?

### DIFF
--- a/students/tri_nguyen/lesson04/trigrams.py
+++ b/students/tri_nguyen/lesson04/trigrams.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+# ------------------------------------------- #
+# Student:   Tri Nguyen
+# Instructor: Natasha Aleksandrova
+# Date:  28-Feb-2018
+# ------------------------------------------- #
+
+import random
+
+
+def get_word_list():
+    ''' this function create a list of words from 
+        sherlock_small.txt '''
+
+    word_list = []
+
+    with open('sherlock_small.txt') as f:
+        for line in f:
+            line.rstrip()
+            for char in line:
+                line = line.replace('--', ' ')
+            words = line.split()
+
+            for word in words:
+                word_list.append(word)
+
+    return word_list
+
+
+def create_word_dict():
+    ''' create a dictionary that looks like the
+        trigram table '''
+
+    word_dict = {}
+
+    txt_words = get_word_list()
+
+    for idx in range(len(txt_words) - 2):
+        word_dict.setdefault((txt_words[idx], txt_words[idx + 1]), []).append(txt_words[idx + 2])
+
+    return word_dict
+
+
+word_dict = create_word_dict()
+word_list = get_word_list()
+word_list_idx = list(range(len(word_list)))
+
+start_point = random.choice(word_list_idx)
+
+new_text = [word_list[start_point], word_list[start_point + 1]]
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
@tricaonguyen yes it's an interesting but also challenging problem to solve 🙂 

I usually write the "design" on a piece of paper to help me think it through and then write out the code, at least that helps me.

I have uploaded a fairly simple solution (here)[https://github.com/natasha-aleksandrova/Self_Paced-Online/blob/master/solutions/lesson4/trigrams.py], it does not handle more complex text processing, which can be easily expanded, or writing final text a to file (but that's simple enough).

students/tri_nguyen/lesson04/trigrams.py
```diff
@@ -0,0 +1,66 @@
...
+    ''' create new text recursively '''
+
+    if rand_pair in word_dict:
+        next_word = random.choice(word_dict[rand_pair])
+        new_text.append(next_word)
+        create_new_text((rand_pair[1], next_word))
+
+
+# Run create_new_text recursively to get new_text
+create_new_text(word_pair)
```
natasha-aleksandrova: try to organize your code so that it has a flow, for example it is easy to miss code on lines 45-52 because I wouldn't normally expect somewhat random code in between functions.

following common patterns like having `if __name__ == "__main__":` as an entry point makes it much easier to read.

students/tri_nguyen/lesson04/trigrams.py
```diff
@@ -0,0 +1,66 @@
...
+def create_word_dict():
+    ''' create a dictionary that looks like the
+        trigram table '''
+
+    word_dict = {}
+
+    txt_words = get_word_list()
+
+    for idx in range(len(txt_words) - 2):
+        word_dict.setdefault((txt_words[idx], txt_words[idx + 1]), []).append(txt_words[idx + 2])
```
natasha-aleksandrova: this could be a bit simpler by using `defaultdict` -> `word_dict = defaultdict(list)`

then you have:
```
word_dict((txt_words[idx], txt_words[idx + 1])).append(txt_words[idx + 2])
```

students/tri_nguyen/lesson04/trigrams.py
```diff
@@ -0,0 +1,66 @@
...
+
+def get_word_list():
+    ''' this function create a list of words from 
+        sherlock_small.txt '''
+
+    word_list = []
+
+    with open('sherlock_small.txt') as f:
+        for line in f:
+            line.rstrip()
```
natasha-aleksandrova: strings are immutable, so if you want to actually strip you need to assign result: `line = line.rstrip()`

students/tri_nguyen/lesson04/trigrams.py
```diff
@@ -0,0 +1,66 @@
...
+        word_dict.setdefault((txt_words[idx], txt_words[idx + 1]), []).append(txt_words[idx + 2])
+
+    return word_dict
+
+
+word_dict = create_word_dict()
+word_list = get_word_list()
+word_list_idx = list(range(len(word_list)))
+
+start_point = random.choice(word_list_idx)
```
natasha-aleksandrova: I think `random.randrange(len(word_list))` might be a better option here? (reduces the need for line 47 as well)

PR #243, opened 2018-03-06
Prelim lesson 6 code

kuhnbt7:
Hi Natasha and Rob, as we discussed, I'm submitting my unit test code for review prior to actually turning it in. If you have any suggestions for improvement I'd appreciate them! Thanks


students/kuhnbt/lesson5/mailroom_v3.py
```diff
@@ -0,0 +1,131 @@
...
+donor_info = defaultdict(list)
+for author, donation in [
+             ('Ernest Hemingway', [35, 65.25, 40]),
+             ('John Steinbeck', [25.50, 20, 10]),
+             ('Joseph Heller', [45, 20, 105.50]),
+             ('Kurt Vonnegut', [60, 40, 400]),
+             ('Michael Chabon', [100, 200, 300, 400])
+             ]:
+    donor_info[author] = donation
+# Is there a simpler way to set initial values for a defaultdict?
```
natasha-aleksandrova: yes! pass your initial dict as a second param, here's an example:

```
>>> from collections import defaultdict
>>> donors = defaultdict(list, {"Andy": [10.00]})
>>> donors["natasha"].append(1000)
>>> donors
defaultdict(<class 'list'>, {'Andy': [10.0], 'natasha': [1000]})
```

students/kuhnbt/lesson6/mailroom_unittests.py
```diff
@@ -0,0 +1,43 @@
...
+        with self.assertRaises(SystemExit):
+            quit_program()
+
+    def test_thankyou(self):
+        hemingway_text = 'Dear Ernest Hemingway:\nThank you for your'\
+            ' generous donation of $40.00.\nI really appreciate your '\
+            '3\ndonations to our organization.\nI assure you that your'\
+            ' contributions will be put to\ngood use!\n\nRegards,\nBen'
+
+        self.assertEqual(hemingway_text, get_thank_you('Ernest Hemingway'))
```
natasha-aleksandrova: very nice!

a good idea is to test variations of `get_thank_you`, since you have special logic for "donations" vs "donation"

so you could have:
`test_get_thank_you_single_donation`
`test_get_thank_you_multiple_donations`

you should be able to over write `donor_info` with your test set of data to allow you to test various cases without compromising mailroom data set.
```
mailroom_v4.donor_info = defaultdict(list, {"Test Donor1": [1], "Test Donor2": [20, 24]})

```

students/kuhnbt/lesson6/mailroom_unittests.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # I refactored the create_report function by splitting
+    # out functions that create the report header and report rows,
+    # allowing me to unit test those parts of the function. But
+    # I'm not if the added complexity is worth it...
```
natasha-aleksandrova: I think your refactor was justifiable and does look more readable

students/kuhnbt/lesson6/mailroom_unittests.py
```diff
@@ -0,0 +1,43 @@
...
+                                        [25.5, 20, 10]), 16),
+                         expected_row)
+
+    # I'm not sure how to test start_program because it doesn't
+    # return a value and all functionality is based on user input
+
+    # Same problem for send_thankyou, though test_thankyou ensures
+    # that the string used in send_thankyou is as expected. Same
+    # deal for send_to_everyone, which writes to files and doesn't
+    # return anything
```
natasha-aleksandrova: in order to test `start_program` or `send_thankyou` you need to use `mock` module, this is an advanced concept and not required for this lesson. (next course in series goes over that).

if you have the extra time and want to learn, you can give it a try. Otherwise, this is good enough. The goal was to get exposure to `unittest` module, and looks like you got 🙂

kuhnbt7:
Ok thanks Natasha. I made a couple changes (I actually noticed an error because of the single donation unit test!) and will push again and submit on the course site.

PR #245, opened 2018-03-06
Lesson04 mailroom part2

csdotson:
Completed mailroom_2.py for Lesson04.

### DIFF
--- a/students/csdotson/lesson04/mailroom_2.py
+++ b/students/csdotson/lesson04/mailroom_2.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+# Lesson 3 - Mailroom... part 1
+
+# Initial list of donors/donations
+donations = [["Bill Gates", 10.50, 123.45, 1111.11],
+             ["Jeff Bezos", 7.65, 1000],
+             ["Paul Allen", 145.90],
+             ["John Nordstrom", 45.67, 6519.65],
+             ["Mark Zuck", 789.12]]  # Convert this to a dict
+
+
+### Function Definitions ###
+def prompt():
+    # Original prompt for main menu selections
+    while True:
+        response = input("What would you like to do?\n1 - Send a Thank You\n2 - Create a Report\n3 - Quit\n--> ")
+
+        if response == "1":
+            print()
+            send_thank_you()
+            break
+        elif response == "2":
+            print()
+            print_report()
+            break
+        elif response == "3":
+            break
+        else:
+            print("Invalid response, please try again...")
+            print()
+            prompt()
+
+def send_thank_you():
+    # Add donor/donations and trigger thank you email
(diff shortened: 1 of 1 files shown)


students/csdotson/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,107 @@
...
+
+def send_thank_you():
+    # Create sub_prompt for creating thank you's and adding new donors/donations
+    prompt(thank_you_prompt, thank_you_menu)
+
+
+def add_donation():
+    # Append donation to existing donor, or add donor/donation to existing list
+    response = input("\nPlease enter a donor name: ")
+    if response not in donations.keys():
```
robalford: `in` checks for key membership in dict, so this can be `if response not in donations:`

students/csdotson/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,107 @@
...
+        new_donor = response
+        new_donation = input(f"Please enter a donation amount for {new_donor}: ")
+        donations[new_donor] = [float(new_donation)]
+        print_email(new_donor, new_donation)
+    else:
+         current_donor = response
```
robalford: Looks like the code in the `else` block here is over indented by a space

students/csdotson/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,107 @@
...
+         new_donation = input(f"Please enter a donation amount for {current_donor}: ")
+         donations[current_donor].append(float(new_donation))
+         print_email(current_donor, new_donation)
```
robalford: I think you could eliminate some of the repeated code here (always a good thing) by doing something like this:

```
def add_donation():
    donor = input("\nPlease enter a donor name: ")
    new_donation = input(f"Please enter a donation amount for {donor}: ")
    if donor not in donations:
        donations[donor] = [float(new_donation)]
    else:
        donations[donor].append(float(new_donation))
    print_email(donor, new_donation)
```

students/csdotson/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def write_letters():
+    # For all donors, generates a thank you letter, and writes it to disk as a text file
+    for donor in donations:
+        file_name = '_'.join(donor.split()) + ".txt"
+        with open(file_name, 'w') as f:
+            f.write(f"Dear {donor},\n\nThank you very much for your generosity! Your most recent gift of ${(donations[donor][-1])} will be put to great use. So far, you've donated a total of ${sum(donations[donor])}!\n\nSincerely,\nThe Team")
```
robalford: You can use triple quotes `"""` to keep long strings under Pep8 line character limit and they also let you preserve white space in your strings, so this could be written like this:

```
f.write(f"""Dear {donor},
            
            Thank you very much for your generosity! Your most recent gift of ${(donations[donor][-1])} will be 
            put to great use. So far, you've donated a total of ${sum(donations[donor])}!
                    
            Sincerely,
            The Team""")
```

PR #248, opened 2018-03-07
Submitting mailroom.py

wesmoy17:


students/WesleyM/Lesson3/mailroom.py
```diff
@@ -0,0 +1,47 @@
+donors = [
+  ["Alice", 20],
+  ["Bob", 100],
+  ["Charles", 30, 50, 10],
+  ["Denise", 10, 5],
+  ["Edward", 25, 20, 20, 20, 10],
+  ]
+
+def usein():
```
robalford: Python style guide uses snake case for functions and variable names and it's a good idea to avoid abbreviations/short hand for readability, so I'd probably call this function something like `user_input`

students/WesleyM/Lesson3/mailroom.py
```diff
@@ -0,0 +1,47 @@
...
+  action = input("Choose one of three actions: \n 1. Send a Thank You \n 2. Create a Report \n 3. Quit \n>")
+  while (action.lower() not in ["1", "2", "3", "send a thank you", "create a report", "quit"]):
```
robalford: You don't need parentheses around the `while` clause and you might want to assign this list to a variable outside the while block for readability so you could say something like `while action.lower() not in options`

students/WesleyM/Lesson3/mailroom.py
```diff
@@ -0,0 +1,47 @@
...
+    sendty()
+    usein()
+  if action.lower() in ['2', 'create a report']:
+    report()
+    usein()
+  if action.lower() in ['3', 'quit']:
+    return
+ 
+
+def sendty():
```
robalford: Same note as above here on naming functions

students/WesleyM/Lesson3/mailroom.py
```diff
@@ -0,0 +1,47 @@
...
+  donor = input("Enter a donor name or input 'List' for a list of donors\n>")
+  if donor.lower() == "list":
+    for names in donors:
+        print(names[0])
+  else:
+    donation = input(("Enter a donation amount for {} \n>").format(donor))
```
robalford: You don't need the parentheses around the `"Enter a donation amount for {} \n>"` string here

students/WesleyM/Lesson3/mailroom.py
```diff
@@ -0,0 +1,47 @@
...
+    for names in donors:
+      if donor == names[0]:
+        names.append(int(donation))
+        break
+    else:
+      donors.append([donor, int(donation)])
+    print(("Thank you {} for your donation of ${:.2f}").format(donor, int(donation)))
```
robalford: can lose the parentheses around the string here as well

PR #249, opened 2018-03-07
[DennisLee][Lesson 04][Trigrams] Ready for submission.

DennisLee-DennisLee:


students/DennisLee/lesson04/trigram.py
```diff
@@ -0,0 +1,349 @@
...
+        ui = input("\nType a menu number: ")
+        if ui == '1':  # Quit
+            print("\n\nGoodbye!")
+            break
+        elif ui == '2':  # Run the mutator with the specified filename
+            print(f'\n\nCurrent directory is {os.getcwd():s}')
+            file_name = input(
+                    '\n\nType the name of the file to mutate using trigrams: '
+                    ).strip()
+            print('\n\n', build_new_story(file_name, counter))
```
natasha-aleksandrova: For readability I would recommend not putting function calls into `print` statements. I wouldn't normally read `print` statements as one wouldn't expect important logic to be part of it, but some strings that get printed out.

My point is, it is easy to skip over and miss important logic piece 🙂

students/DennisLee/lesson04/trigram.py
```diff
@@ -0,0 +1,349 @@
...
+    the story may sometimes fail, eventually a full story will be built.
+
+    :source_file:  The file containing the original story text.
+
+    :counter:  A miscellaneous number for randomizing the output text.
+
+    :return:  The output text string, containing at least 200 words.
+    """
+    print('\n\nBuilding the story...')
+    if type(counter) != int:
```
natasha-aleksandrova: keep in mind that it is recommended to use `isinstance` for type checking

students/DennisLee/lesson04/trigram.py
```diff
@@ -0,0 +1,349 @@
...
+            # Get the text quotation state after the initial two words
+            quote_opened = quote_state(quote_opened, text[0])
+            quote_opened = quote_state(quote_opened, text[1])
+        else:  # Continue sentence by forming the next two-word key
+            two_words = get_next_key(two_words, next_word)
+
+        # Get a word to complete the trigram
+        next_word = pick_from_choices(word_pairs[two_words], quote_opened)
+        if next_word == None:
+            err = True
```
natasha-aleksandrova: could this have a `break` statement here instead?

PR #250, opened 2018-03-07
Lesson02

natrod22:
I've uploaded three different files to print a grid in three different ways.
I used PyCharm to develop code for Lesson02, so hopefully it is more in line with Pep8. 

I'm open to any and all comments so that I can make improvements.


students/NatalieRodriguez/Lesson02/grid_3_inputs.py
```diff
@@ -0,0 +1,25 @@
...
+    row(a)
+    for i in range(b):
+        for i in range(c):
+            column(a)
+        row(a)
+
+a=int(input("Columns: "))
+b=int(input("Rows: "))
+c=int(input("Insides: "))
+func(a,b)
```
natasha-aleksandrova: need a bit more spacing:
```
a = int(input("Columns: "))
b = int(input("Rows: "))
c = int(input("Insides: "))
func(a, b)
```

also choose more descriptive var names when appropriate 🙂 

for example
```
column_size = int(input("Columns: "))
row_size = int(input("Rows: "))
insides_size = int(input("Insides: "))
func(column_size, row_size)

def func(column_size, row_size)
```

this is a small piece of code, but imagine lots of code and you're trying to step through to understand the logic.. then you go "what was 'a' var again?"

PR #253, opened 2018-03-07
Submitting Fibonacci in Series Module for Feedback

traub01-devtrain:
Submitting Fibonacci in Series Module for Feedback.


students/JamesTraub/Lesson02/series.py
```diff
@@ -0,0 +1,30 @@
...
+    elif n == 1:
+        return 1
+    else:
+        for i in range(n):
+            # create Fibonacci series
+            small_fib = fibi[i]
+            range_plus1 = i + 1
+            big_fib = fibi[range_plus1]
+            next_fib = small_fib + big_fib
+            fibi.append(next_fib)
```
natasha-aleksandrova: I would suggest replacing all of the above with:
`fibi.append(fibi[i] + fibi[i+1])` which uses simple indexing; it is less readable with all the extra variables that are being used.

PR #256, opened 2018-03-08
Herela: Lesson 8

harvey-herela:


students/HarveyHerela/lesson08/Circle.py
```diff
@@ -0,0 +1,99 @@
+from math import pi
+from functools import total_ordering
+
+import random
+import time
+
+@total_ordering
+class Circle:
```
natasha-aleksandrova: class foo(object): is the 'new' way of declaring classes (since python 2.2).

so this should be `class Circle(object)`

students/HarveyHerela/lesson08/Circle.py
```diff
@@ -0,0 +1,99 @@
...
+    
+    def __init__(self, the_radius):
+        self._radius = the_radius
+        self._diameter = the_radius * 2
+            
+            
+    @classmethod
+    def from_diameter(cls, diameter):
+        self = cls(diameter / 2)
+        return self
```
natasha-aleksandrova: should just be `return cls(diameter / 2)`

students/HarveyHerela/lesson08/Circle.py
```diff
@@ -0,0 +1,99 @@
...
+        return "Circle with radius: {}".format(self.radius)
+        
+    
+    def __repr__(self):
+        return "Circle({})".format(self.radius)
+    
+    
+    def __add__(self, other):
+        bigger_circle = Circle(self.radius + other.radius)
+        return bigger_circle
```
natasha-aleksandrova: can be inline return as well

students/HarveyHerela/lesson08/Circle_Unittests.py
```diff
@@ -0,0 +1,102 @@
+from math import pi
+import random
+import unittest
+
+from Circle import Circle
+
+class Circle_Test(unittest.TestCase):
```
natasha-aleksandrova: class names have no underscores -> `CircleTest`

students/HarveyHerela/lesson08/Circle_Unittests.py
```diff
@@ -0,0 +1,102 @@
...
+            
+    
+    def test_sort(self):
+        circle_list = [Circle(random.randint(0, 10000)) for i in range(100)]
+        circle_list_s1 = sorted(circle_list)
+        circle_list_s2 = sorted(circle_list, key=Circle.sort_key)
+        self.assertEqual(circle_list_s1, circle_list_s2)
+        for i in range(99):
+            self.assertTrue(circle_list_s1[i] < circle_list_s1[i + 1])
+
```
natasha-aleksandrova: Couple notes on testing:
* in this case, I'm not sure `for i in range(100)` useful for testing Circle. Can't think of a benefit it would provide. Something like `Circle(-1)` or `Circle(0)` or `Circle('hi')` type of tests would be more useful.
* with above in mind, you can utilize `setUp` method that's part of the unittest. It sets up whatever you want before each test, since Circle is instantiated before each test we can move that there:
```
def setUp(self):  # unittest uses java style so methods are camel case
    self.radius = 10
    self.circle_test = Circle(self.radius) 

def test_constructor(self):
    self.assertEqual(self.circle_test.radius, self.radius)
```
* would also like to see tests that assert failures: ex: `c.area = 42` should result in attribute error

students/HarveyHerela/lesson08/Circle.py
```diff
@@ -0,0 +1,99 @@
...
+    Print the circle and get something nice
+    Be able to add two circles together
+    Be able to compare two circles to see which is bigger
+    Be able to compare to see if there are equal
+    (follows from above) be able to put them in a list and sort them
+    """
+    
+    def __init__(self, the_radius):
+        self._radius = the_radius
+        self._diameter = the_radius * 2
```
natasha-aleksandrova: take a look at this:
```
c = Circle(5)
print(c.radius)  # --> 5
c._diameter = 4
print(c.diameter) # --> 4
print(c.radius) # --> 5.. oops not good!
```
whenever I use setters/properties, I use double leading underscores -> `self.__diameter` this prevents the user from modifying vars directly, so they will always go via setters and this will ensure that proper logic happens.

students/HarveyHerela/lesson08/Circle.py
natasha-aleksandrova: I take this back :) in Python3 it's always new style class regardless how you define it. My comment only applies to python2.x (greater than 2.2)

harvey-herela: Since explicit is better than implicit, does that mean that, even though both are correct, it's better style to explicitly show the subclassing of object?

PR #258, opened 2018-03-08
FizzBuzz Code

natrod22:
This one was fun to figure out :)


students/NatalieRodriguez/Lesson02/Fizz_Buzz.py
```diff
@@ -0,0 +1,15 @@
+for i in range(0,101):
```
natasha-aleksandrova: note that zero is the default start for `range` so you can either do `range(1, 101)` to start at 1 or `range(101)` to start at zero

PR #261, opened 2018-03-08
List Lab Exercise

itsabhister:


students/itsabhi/Lesson3/list_lab.py
```diff
@@ -0,0 +1,88 @@
+#! /usr/bin/env Python3
+
+
+def series1():
+    my_list = ["Apples", "Pears", "Oranges", "Peaches"]
+    print(my_list)
+    add_fruit = input("Add another fruit: ")
+    my_list.append(add_fruit)
+    print(my_list)
+    (number) = input("Enter a number: ")
```
robalford: I wouldn't put parentheses around `number` here, since parentheses have meaning in Python this could lead to confusion

students/itsabhi/Lesson3/list_lab.py
```diff
@@ -0,0 +1,88 @@
...
+    print(my_list[int(number)-1])
+    my_list = ["Mangos"] + my_list
+    print(my_list)
+    my_list.insert(0, "Berries")
+    print(my_list)
+    for fruit in my_list:
+        if fruit.startswith("P"):
```
robalford: You can use `if fruit.lower().startswith('p'):` to catch lower case user input

students/itsabhi/Lesson3/list_lab.py
```diff
@@ -0,0 +1,88 @@
...
+    like_oranges = input("Do you like oranges: ")
+    like_peaches = input("Do you lile peaches: ")
+    while not (like_apples == "yes" or like_apples == "no"):
+        like_apples = input("Do you like apples, yes or no: ")
+    while not (like_pears == "yes" or like_pears == "no"):
+        like_pears = input("Do you like pears, yes or no: ")
+    while not (like_oranges == "yes" or like_oranges == "no"):
+        like_oranges = input("Do you like oranges, yes or no: ")
+    while not (like_peaches == "yes" or like_peaches == "no"):
+        like_peaches = input("Do you like peaches, yes or no: ")
```
robalford: There is a lot of repeated code in this section. Can you think of a way to make it more DRY by placing the prompts inside a loop?

students/itsabhi/Lesson3/list_lab.py
```diff
@@ -0,0 +1,88 @@
...
+            my_list.remove(my_list_copy[i])
+        i += 1
+    print(my_list)
+
+
+def series4():
+    my_list = ["Apples", "Pears", "Oranges", "Peaches"]
+    my_list_copy = my_list[:] #Shallow copy
+    reverse_my_list = []
+    for item in my_list_copy:
```
robalford: Nice catch on looping through a copy of the original list!

students/itsabhi/Lesson3/list_lab.py
itsabhister: good catch!

students/itsabhi/Lesson3/list_lab.py
itsabhister: Maybe iterate the "for" loop over a list? I agree there is room to optimize this piece.

PR #262, opened 2018-03-09
Lesson 3 List lab

preactive:
List lab submission and fixes for PR 169.

let me know what you think of the mod i made on line 24 here: https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/commit/9c3e61a931a70719e1e9118be68ad0a3f2cd1587#diff-b3e27a23e5fea67b49894d5c5caad2e8R24

and the testing asset fix for PR 169

### DIFF
--- a/students/preactive/Lesson2/series.py
+++ b/students/preactive/Lesson2/series.py
@@ -13,12 +13,33 @@ def fib_gen():
     while True:
         a, b = b, a + b
         yield a
-f = fib_gen()
-next(f)
-next(f)
-next(f)
-next(f)
-print(next(f))
+
+
+def fib_gen_teacher():
+    a,b = 0,1
+    while True:
+        a, b = b, a + b
+        yield a
+
+def fib_gen_two():
+    a, b, n = 0, 1, 0
+    while True:
+        n += 1
+        if n == 1:
+            yield 0
+        a, b = b, a + b
+        yield a
+
+def fib_gen_call():
+    f = fib_gen_teacher()
+    print(next(f))
+    print(next(f))
+    print(next(f))
+    print(next(f))
+    print(next(f))
+    print(next(f))
+
+fib_gen_call()
 
 
 def lucas(n):
@@ -54,3 +75,5 @@ def assert_testing():
         assert sum_series(lucas_index + 1, 2, 1) == lucas_test, \
             "Sum Series Lucas Failed"
 assert_testing()
+
+
--- a/students/preactive/Lesson2/series.py
+++ b/students/preactive/Lesson2/series.py
@@ -24,20 +24,21 @@ def fib_gen_teacher():
 def fib_gen_two():
     a, b, n = 0, 1, 0
     while True:
-        n += 1
-        if n == 1:
+        if n == 0:
+            n = True
             yield 0
         a, b = b, a + b
         yield a
 
 def fib_gen_call():
-    f = fib_gen_teacher()
-    print(next(f))
-    print(next(f))
-    print(next(f))
-    print(next(f))
-    print(next(f))
-    print(next(f))
+    fib_teach = fib_gen_teacher()
+    fib_mez = fib_gen_two()
+    print(str(next(fib_teach)) + " " + str(next(fib_mez)))
+    print(str(next(fib_teach)) + " " + str(next(fib_mez)))
+    print(str(next(fib_teach)) + " " + str(next(fib_mez)))
+    print(str(next(fib_teach)) + " " + str(next(fib_mez)))
+    print(str(next(fib_teach)) + " " + str(next(fib_mez)))
+    print(str(next(fib_teach)) + " " + str(next(fib_mez)))
 
 fib_gen_call()
 
--- a/students/preactive/Lesson3/list_lab.py
+++ b/students/preactive/Lesson3/list_lab.py
@@ -16,17 +16,126 @@ def series_one():
     # the fruit corresponding to that number (on a 1-is-first basis).
     # Remember that Python uses zero-based indexing,
     # so you will need to correct.
-    response_two_txt = "Pick a integer from 1 to %s:\n>" % str(len(fruits))
+    num_fruits = len(fruits)
+    response_two_txt = "Pick a integer from 1 to %s:\n>" % str(num_fruits)
     response_two = input(response_two_txt)
-    print(response_two + ":" + fruits[int(response_two) - 1])
-
-    if response_two not in range(1,5):
-        print("Not in range. Try again mouth breather.")
+    try:
+        response_two_int = int(response_two)
+        if response_two_int not in range(1,num_fruits + 1):
+            print("Not in range. Try again, mouth breather.")
+            series_one()
+        else:
+            print(response_two + ": " + str(fruits[response_two_int-1]))
+    except:
+        print("Not in ranges. Try again, mouth breather.")
         series_one()
-    else:
-        print(response_two + ":" + fruits[response_two-1])
 
+    # Add another fruit to the beginning of the
+    # list using “+” and display the list.
+    fruits = ["tomato"] + fruits
+    print(fruits)
+
+    # Add another fruit to the beginning of the list
+    # using insert() and display the list.
+    fruits.insert(len(fruits),"banana")
+    print(fruits)
+
+    # Display all the fruits that begin with “P”, using a for loop.
+    for fruit in fruits:
+        try:
+            if fruit[0:1].lower() == "p":
+                print(fruit)
+        except:
+            print("Something went wrong with lower method. "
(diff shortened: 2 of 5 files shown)


students/preactive/Lesson2/series.py
```diff
@@ -13,12 +13,34 @@ def fib_gen():
...
-print(next(f))
+
+
+def fib_gen_teacher():
+    a,b = 0,1
+    while True:
+        a, b = b, a + b
+        yield a
+
+def fib_gen_two():
```
robalford: It's honestly been a while since I did anything with generators (don't have many use cases in my web dev work). That's great that you're looking at this as an alternative approach to the problem, and I believe they'll by using fibonacci to teach generators in the second quarter so you're getting a massive head start if you plan to continue on ;) 

I think you could just do this to yield 0 for the first item in the sequence though (@natasha-aleksandrova maybe you want to chime in here since you guys were looking at this before): 

```
def fib_gen_teacher():
    yield 0
    a,b = 0,1
    while True:
        a, b = b, a + b
        yield a
```

students/preactive/Lesson3/list_lab.py
```diff
@@ -0,0 +1,142 @@
...
+    response_two_txt = "Pick a integer from 1 to %s:\n>" % str(num_fruits)
+    response_two = input(response_two_txt)
+    try:
+        response_two_int = int(response_two)
+        if response_two_int not in range(1, num_fruits + 1):
+            print("Not in range. Try again, mouth breather.")
+            series_one()
+        else:
+            print(response_two + ": " + str(fruits[response_two_int - 1]))
+    except:
```
robalford: It's a best practice to narrow in on a specific exception, so you don't end up allowing some unanticipated error to pass silently. Seems like you'd want to catch ValueError here and modify your error message, since you're already checking the input range inside your try block.

students/preactive/Lesson3/list_lab.py
```diff
@@ -0,0 +1,142 @@
...
+        series_one()
+
+    # Add another fruit to the beginning of the
+    # list using “+” and display the list.
+    fruits = ["tomato"] + fruits
+    print(fruits)
+
+    # Add another fruit to the beginning of the list
+    # using insert() and display the list.
+    fruits.insert(len(fruits), "banana")
```
robalford: You want to add 'banana' to the beginning of the list here

students/preactive/Lesson3/list_lab.py
```diff
@@ -0,0 +1,142 @@
...
+    print(fruits)
+
+    # Display all the fruits that begin with “P”, using a for loop.
+    for fruit in fruits:
+        try:
```
robalford: We'll be covering try/except in detail a bit later in the course but generally you want to use a try block when there's a situation where a specific exception is possible or likely to occur. This doesn't seem to be the case here.

students/preactive/Lesson3/list_lab.py
```diff
@@ -0,0 +1,142 @@
...
+            if fruit[0:1].lower() == "p":
```
robalford: This can just be `fruit[0]`

students/preactive/Lesson3/list_lab.py
```diff
@@ -0,0 +1,142 @@
...
+    print(series_one_stub)
+
+    # Remove the last fruit from the list.
+    series_one_stub.pop()
+
+    # Display the list.
+    print(series_one_stub)
+
+    # Ask the user for a fruit to delete, find it and delete it.
+    delete_fruit = input("enter a fruit to remove. CaSe SeNsItIvE:\n>")
```
robalford: You could use the `lower()` method here to make case insensitive

students/preactive/Lesson3/list_lab.py
```diff
@@ -0,0 +1,142 @@
...
+        print(series_one_stub)
+    else:
+        print("Try again, El Jefe!")
+        series_two()
+
+    # (Bonus: Multiply the list times two. Keep asking until
+    # a match is found. Once found, delete all occurrences.)
+    doubled_list = []
+    for fruit in series_one_stub:
+        doubled_list.append(fruit * 2)
```
robalford: You want to multiply the list times two here, not the individual items, so you end up with two of each item in the list and then delete all occurrences of each item as you loop through the list

students/preactive/Lesson3/list_lab.py
```diff
@@ -0,0 +1,142 @@
...
+    processable = True
+    while processable:
+        delete_fruit = input("enter a times two fruit item to remove."
+                             " CaSe SeNsItIvE:\n>")
+        if delete_fruit in doubled_list:
+            doubled_list = list(filter(lambda a: a != delete_fruit, doubled_list))
```
robalford: It's awesome that you're trying out some more advanced features of Python but I think you could achieve the same thing here with a simple while loop and the `list.remove()` method

students/preactive/Lesson3/list_lab.py
```diff
@@ -0,0 +1,142 @@
...
+            continue
+        while processable:
+            like_fruit = input("(yes|no) Do you like %s\n>" % fruit.lower())
+
+            # For each “no”, delete that fruit from the list.
+            # For any answer that is not “yes” or “no”, prompt the user to
+            # answer with one of those two values (a while loop is good here)
+            if "no" == like_fruit:
+                series_one_stub = list(
+                    filter(lambda a: a != fruit, series_one_stub))
```
robalford: If you loop through a copy of the original list you can remove items from the list while you loop and simplify a lot of this code

robalford:
@preactive list lab looks good! I made some suggestions for ways to simplify the code that you may want to consider. Do you want us to review slicing assignment in this PR as well, or will you be submitting that separately?

students/preactive/Lesson3/list_lab.py
preactive: will ValueError catch when someone tries to inject funny business like: i̷̡̫̠͖͌̾̑́̑̀ͨͦ͛̚͠͞n̶͙̰̭̳̖̘̹̻͚̜̾ͣ̔ͮͤ̉͂̐̌̋ͪ̀̌́͜p̸̛͍̯̯͖̲̹̜̦̼̲̥͌͌̇͗̆̾ͬ̏̀ͪ͆̅ͧ̍͗̍̉̄̓͘̕͢ͅu̽̇̋̋̊̃ͪ̂̓̾ͣ̆͂̈́́͏̢̰̫̭̟̣̞̗̰̰̼͇̤͉̯̹̣̠̗͠t̸̝͔̥̲͈̼͈̦̰̭̼̗͔̹̳̭̦͎͒͆ͦ̕

robalford: I think you might want to use a custom exception class for that ;)

natasha-aleksandrova:
I believe slicing is an update to existing submission.

students/preactive/Lesson2/series.py
preactive: Didnt know you could have multiple yields.  thanks!

students/preactive/Lesson3/list_lab.py
preactive: fixed

students/preactive/Lesson3/list_lab.py
preactive: ok went with ValueError.  Fixed

students/preactive/Lesson3/list_lab.py
preactive: I tried a few edge cases and i did not hit the execption.  Removed try block

students/preactive/Lesson3/list_lab.py
preactive: fixed

students/preactive/Lesson3/list_lab.py
preactive: I like to remind people of case, as the majority of the internet runs off of *nix of some kind.  The only places where I like to normalize is login IDs as this causes problems when people login.  that one time when register for a service and you use "Preactive" and then 3 years later you login with preactive and it says your not a user.  [triggered].  until some decent admin picks up the ticket and looks at the database for another user of the same case and unlocks it.

students/preactive/Lesson3/list_lab.py
preactive: Ahh that was not explicit in the instruction.  But I tested it and the logic still works to remove all instances.

students/preactive/Lesson3/list_lab.py
preactive: Indeed.  Sometimes I dont like lots of lines under an IF to make the blocks more readable for what actions are being taken at the cost of more complex and shorter logic.

PR #263, opened 2018-03-09
Lesson 02 submission for grid print exercise.

mmsunday1:


students/M_Sunday/lesson01/break_me.py
```diff
@@ -0,0 +1,18 @@
...
+        str = 'hello'
+        print(a * str)
+    elif type == 'SyntaxError':
+        #print(,)
+        pass
+    elif type == 'AttributeError':
+        text = 'cool'
+        text.atterror
+
+call_except(type = 'AttributeError')
```
natasha-aleksandrova: be careful about overriding built-in functions in python 🙂 

these usually have special syntax highlighting and if your editor doesn't highlight, github will. notice the highlight on `type` var, it's a built-in function.

PR #264, opened 2018-03-09
lesson02 fib, lucas, and sum series assignment

natrod22:
I'm not sure I did the assert statements correctly. It did not throw an exception, so I assumed they went through alright.

### DIFF
--- a/students/NatalieRodriguez/Lesson02/sequences.py
+++ b/students/NatalieRodriguez/Lesson02/sequences.py
@@ -1,0 +1,71 @@
+# Fibonacci Sequence
+# Natalie Rodriguez
+# 3/8/2018
+
+def fib_seq(n):
+#using recursion to calculate the Fibonnaci Sequence
+#have the user call the function to name the 'n'th term
+#of the sequence.
+
+    if n == 0:
+        return 0 #the zeroth term. These would not follow the formula below if called.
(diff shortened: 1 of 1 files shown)


students/NatalieRodriguez/Lesson02/sequences.py
```diff
@@ -0,0 +1,71 @@
+# Fibonacci Sequence
+# Natalie Rodriguez
+# 3/8/2018
+
+def fib_seq(n):
+#using recursion to calculate the Fibonnaci Sequence
+#have the user call the function to name the 'n'th term
+#of the sequence.
```
natasha-aleksandrova: your comments should be indented or use triple double quotes for doc strings:
```
def fib_seq():
    """Fib seq function
    here is the explanation
    """
```

students/NatalieRodriguez/Lesson02/sequences.py
```diff
@@ -0,0 +1,71 @@
...
+
+    if n == 0:
+        return 0 #the zeroth term. These would not follow the formula below if called.
+    elif n == 1: #the first term
+        return 1
+    else:
+        return (fib_seq(n - 1) + fib_seq(n - 2))
```
natasha-aleksandrova: you don't need the extra parens in return

PR #266, opened 2018-03-10
Lesson 6: Testing & refactoring mailroom

nDruP:
This one really confused me. 
Please let me know if I've refactored the correct parts of my code to make them more test-friendly.
If not, I definitely need some more insight to the decision making behind what to refactor.

### DIFF
--- a/students/nDruP/lesson06/test_mailroom4.py
+++ b/students/nDruP/lesson06/test_mailroom4.py
@@ -1,4 +1,6 @@
 from mailroom4 import *
+import datetime
+import os.path
 
 
 def test_main_menu():
@@ -7,11 +9,13 @@ def test_main_menu():
     assert main_menu("3") == create_letters
     assert main_menu("4") == sys.exit
 
+
 def test_check_not_exit():
     assert check_not_exit("Exit") == False
     assert check_not_exit("exit") == False
     assert check_not_exit("Bpb") == True
     assert check_not_exit("") == True
+
 
 def test_user_input():
     test_str = ["exit", "Bob", "12345", "cwd"]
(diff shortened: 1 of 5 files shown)


students/nDruP/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def test_thank_u_str():
+    assert thank_u_str("Andrew", 13) == (divider + f"Dearest Andrew,\n"
+            f"\tThank you for donation(s) of $13.00!\n"
+            "We will use your donation(s) to create real living Pokemon.\n"
+            "You now have our eternal loyalty. Use it wisely.\n"
+            "Sincerely,\n"
+            f"We're a Pyramid Scheme & so is Andrew"
+            + divider)
```
natasha-aleksandrova: you generally  want your expected part to be independent of actual code you're testing. (dependency on divider from mailroom)

also looks like you're using `pytest`, expected should be on the left side

students/nDruP/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,83 @@
...
+    curdate = (datetime.datetime.now()).strftime("%Y_%m_%d")
+    x = ("Emelio h", [1, 2, 3])
+    y = ("Garbanzo o", [4, 5, 6])
+    z = ("Preid", [7, 8, 9])
+    write_letter_to_dir(x[0], x[1])
+    write_letter_to_dir(y[0], y[1])
+    write_letter_to_dir(z[0], z[1])
+    assert os.path.isfile("Emelio_h_"+curdate+".txt")
+    assert os.path.isfile("Garbanzo_o_"+curdate+".txt")
+    assert os.path.isfile("Preid_"+curdate+".txt")
```
natasha-aleksandrova: not anything you need to know, but something to keep in mind

since the clock is ticking there is a chance that curdate will not match to curdate in your `write_letter_to_dir` function. for this we usually use something called mocking, where we can essentially "freeze" the time (set fake current time) and ensure that no matter what we always get matching current time.

natasha-aleksandrova:
@nDruP I think you did a really good job isolating code into units and testing those units. I agree that it can be hard making the decisions on how to make the code unit testable.

The easiest to unit tests are algorithms or library type of code. When you have interactive code like mailroom or other things with complex business logic that's when it gets tricky. Note that typically in developer's toolbox there are unit tests (testing small units of code), integrated (is more like a feature test where you put components together) and automated test (where you would test mailroom module as a whole). All of these types of tests have its place and together create confidence in your code and a full suite.

PR #267, opened 2018-03-10
Lesson8

JerHsi:


students/JerryH/Lesson08/circle.py
```diff
@@ -0,0 +1,57 @@
...
+        return self.radius * 2
+
+    @diameter.setter
+    def diameter(self, value):
+        self.radius = value / 2
+
+    @property
+    def area(self):
+        """ Read Only """
+        return self._area
```
natasha-aleksandrova: note that to make these attributes really read only you need to use double underscore -> `self.__area`

something like this will change the value of area:
```
c._area = 4
```

students/JerryH/Lesson08/circle.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @property
+    def from_diameter(self):
+        return self.diameter
+
+    #@from_diameter.setter
+    def from_diameter(self, diameter):
```
natasha-aleksandrova: (step 5) your `from_diameter` should be an alternate constructor, this means that it should be a class method:
```
@classmethod
    return cls(diameter/2)
```
above will return an instance of Circle class by passing in radius derived from diameter

students/JerryH/Lesson08/circle.py
```diff
@@ -0,0 +1,57 @@
...
+        return "Circle({})".format(self.radius)
+
+    def __str__(self):
+        return "Circle with radius: {}".format(self.radius)
+
+    def __add__(self, other):
+        return "Circle({})".format(self.radius + other.radius)
+
+    def __mul__(self, other):
+        return "Circle({})".format(self.radius * other)
```
natasha-aleksandrova: __add__ and __mul__ methods should do actual addition and multiplication, not return the string repr of it. (the assignment output on the website is produced by the `__repr__` function)

students/JerryH/Lesson08/test_circle.py
```diff
@@ -0,0 +1,65 @@
...
+    def test_return_radius(self):
+        c = Circle(6)
+        self.assertEqual(c.diameter, 12)
+        c.diameter = 4
+        self.assertEqual(c.radius, 2.0)
+
+    def test_area(self):
+        c = Circle(2)
+        self.assertAlmostEqual(c.area, 2**2*pi)
+        with self.assertRaises(AttributeError):
```
natasha-aleksandrova: 👍

PR #268, opened 2018-03-10
lesson 02 - fizz_buzz_exercise complete

mmsunday1:


students/M_Sunday/lesson02/fizz_buzz_exercise.py
```diff
@@ -0,0 +1,11 @@
+# lesson 02 - fizz_buzz_exercise
+
+for i in range(1, 101, 1):
```
robalford: The third 'step' parameter of the `range` function defaults to 1 so no need to pass that in

students/M_Sunday/lesson02/fizz_buzz_exercise.py
```diff
@@ -0,0 +1,11 @@
...
+    if (i % 5 == 0) and (i % 3 == 0):
```
robalford: You can remove the parentheses here

students/M_Sunday/lesson02/series.py
```diff
@@ -0,0 +1,163 @@
...
+        if n == 0:
+            val = 0
+        elif n == 1:
+            val = 1
+        else:
+            val = val_pp + val_p
+            val_pp = val_p
+            val_p = val
+
+    return val
```
robalford: This works but since we know that fibonacci(n) = fibonacci(n-2) + fibonacci(n-1), can you think of a simpler solution that utilizes recursion?

students/M_Sunday/lesson02/series.py
```diff
@@ -0,0 +1,163 @@
...
+verify(27, series(2, 2, 1), lucas(2))
+verify(28, series(3, 2, 1), lucas(3))
+verify(29, series(4, 2, 1), lucas(4))
+verify(30, series(5, 2, 1), lucas(5))
+verify(31, series(6, 2, 1), lucas(6))
+verify(32, series(7, 2, 1), lucas(7))
+verify(33, series(0, 57, 102), 57)
+verify(34, series(1, 57, 102), 102)
+verify(35, series(2, 57, 102), 159)
+verify(36, series(3, 57, 102), 261)
```
robalford: You can simplify your tests by using Python's built in `assert` statement, so these would look like:

```
    assert fibonacci(0) == 0
    assert fibonacci(1) == 1
    assert fibonacci(2) == 1
    assert fibonacci(3) == 2
    assert fibonacci(4) == 3
    assert fibonacci(5) == 5
    assert fibonacci(6) == 8
    assert fibonacci(7) == 13
```

students/M_Sunday/lesson03/slicing_lab_exercise.py
```diff
@@ -0,0 +1,59 @@
...
+
+def reversed1(seq):
+    # This function reverses the item order in a sequence
+    return seq[::-1]
+
+
+def third_swap(seq):
+    # This function orders the items in the sequence with the middle
+    # third of the items, the last third, and then the first third
+    stop = int(len(seq)/3)
```
robalford: You can use the floor division operator `//` here to dealing with numbers that aren't divisible by three

PR #270, opened 2018-03-10
lesson04 / kata14.py

22e88:


students/elmar_m/lesson04/kata14.py
```diff
@@ -0,0 +1,137 @@
...
+    x = x.replace('-', ' ')
+    x = x.replace('.', ' ')
+    x = x.replace(',', ' ')
+    x = x.replace('"', ' ')
+    x = x.replace("'", ' ')
+    x = x.replace(":", ' ')
+    x = x.replace('?', ' ')
+    x = x.replace('!', ' ')
+    x = x.lower()
+    return x
```
natasha-aleksandrova: you can make this a little more flexible:
```
chars_to_replace = ('-', '.', ',')  # add more chars
for char in chars_to_replace:
    x = x.replace(char, ' ')
return x.lower()
```

students/elmar_m/lesson04/kata14.py
```diff
@@ -0,0 +1,137 @@
...
+    '''
+    Remove line breaks from a file. 
+    Return the files content as a string respectively one long line.
+    
+        ARGS:
+    x: the file to operate on
+    ''' 
+    with open(x, 'r') as myfile:
+        l = ' '.join(l.strip() for l in myfile)
+        return l
```
natasha-aleksandrova: a bit more pythonic to return in place -> `return ' '.join(l.strip() for l in myfile)` (especially since the vars used are non-descriptive)

students/elmar_m/lesson04/kata14.py
```diff
@@ -0,0 +1,137 @@
...
+#######################################################################
+
+def main():
+    infile = './sherlock.txt'
+    mydict = collections.defaultdict(list)
+    oneline = make_oneline(infile)
+    cleaner_oneline = cleanup(oneline)
+    biglist = create_biglist(cleaner_oneline)
+
+    for slice in get_stuff_generator(biglist):
```
natasha-aleksandrova: be careful overwriting built-in functions (`slice`) ->  notice github  highlights it

students/elmar_m/lesson04/kata14.py
```diff
@@ -0,0 +1,137 @@
...
+
+        ARGS:
+    x: the dictionary to append to
+    y: the item. Item is expected to be a list of 3 elements. The  
+       first two are joined by whitespace and built the new key. 
+       The third element then builts the value of that key.
+    '''
+    k = ' '.join(y[0:2])
+    v = y[2]
+    x[k].append(str(v))    # put key and value in dictionary
```
natasha-aleksandrova: go ahead and use same var names `def append_to_dict(mydict, myslice)`

generally avoid one letter vars because it is very non-descriptive (in cases like comprehensions it is ok)

PR #278, opened 2018-03-11
08 - minor changes

JerHsi:


students/JerryH/Lesson08/circle.py
```diff
@@ -17,15 +17,11 @@ def diameter(self, value):
...
-    @property
-    def from_diameter(self):
-        return self.diameter
-
-    #@from_diameter.setter
-    def from_diameter(self, diameter):
-        self.diameter = diameter
+    @classmethod
+    def from_diameter(cls, diameter):
+        return cls.diameter
```
natasha-aleksandrova: @JerHsi so the goal here is to provide an alternate constructor, which would allow user to create an instance of a Circle from a diameter. The return of this classmethod should be identical to `Circle(<radius_val>)`

the class method should return an instance -> `return cls(diameter / 2)`

then you can use it like a normal circle instance:
```
c2 = Circle.from_diameter(10)
print(c2)
print(c2.radius)
```

PR #281, opened 2018-03-12
Lesson3 - Mailroom, Part1

DeFeNdog:


students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+    ]
+]
+
+
+def check_for_quit(action):
+    if action.lower() == 'quit':
+        donor_app()
+        return
+    else:
+        return
```
robalford: This function can be written a bit more concisely like this:
```
def check_for_quit(action):
    if action.lower() == 'quit':
        donor_app()
    return  
```

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+def format_donation(donation):
+    donation = float(donation)
+    donation = "{0:.2f}".format(donation)
+    return donation
+
+
+def get_donation():
+    donation = input('Donation amount: ')
+    check_for_quit(donation)
+    return donation
```
robalford: Remember to convert `donation` to a float before storing it in your `_donors` data structure. Right now I'm seeing the following error when trying to generate a report after a new donation is added:
```
  File "./test_student_assignments.py", line 81, in <lambda>
    sorted_donors.sort(key=lambda x: sum(x[1]))
TypeError: unsupported operand type(s) for +: 'float' and 'str'
```

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def add_donor(donor, donation):
+    _donors.append([donor, [float(donation)]])
+    return True
+
+
+def append_donor(donor, donation):
```
robalford: I would probably call this function something like `append_donation` to avoid confusion

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+    for d in _donors:
+        if donor.lower() == d[0].lower():
+            d[1].append(donation)
+            return True
```
robalford: You don't appear to be using the return value anywhere so I think line 71 can be omitted

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+    email_str = 'Dear {a},\nThank you for your generous ' \
+        + 'donation of ${b}.\nRegards\nUW Student'
+    print(email_str.format(a=donor_name, b=donation))
+    return
+
+
+def sort_donors(donors):
+    sorted_donors = list(donors)
+    sorted_donors.sort(key=lambda x: sum(x[1]))
+    sorted_donors.reverse()
```
robalford: `list.sort()` takes a `reverse` parameter that you can use to specify the sort order

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+    return sorted_donors
+
+
+def get_donor_summary(donors):
+    summary = []
+    for i, donor in enumerate(donors):
```
robalford: This can be `for donor in donors:` since you aren't using the index in your loop

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+        d = donor[1]  # all donations
+        s = sum(d)  # total donations
+        g = len(d)
+        m = float(s) / max(len(d), 1)  # mean
```
robalford: I would suggest using meaningful variable name like `average_donation` here for readability

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+        n = []  # new donor item
+        n.append(donor[0])
+        n.append('${0:.2f}'.format(s))
+        n.append(str(g))
```
robalford: It's ok to pass numerical values to `str.format()` so I don't believe you need to call `str` here

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+        n.append('${0:.2f}'.format(m))
+        summary.append(n)
```
robalford: I think you could simplify this block by getting rid of `n` and doing something like this:

```
summary.append([donor[0], '${0:.2f}'.format(s), str(g), '${0:.2f}'.format(m)])
```

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+        print(rf.format(*args))
+
+
+if __name__ == "__main__":
+
+    menu = {}
+    menu['1'] = 'Send a Thank You'
+    menu['2'] = 'Create a Report'
+    menu['3'] = 'quit'
+
```
robalford: It's more common in Python to see simple dicts defined as literals like:
```
menu = {
        '1': 'Send a Thank You',
        '2': 'Create a Report',
        '3': 'quit',
    }
```

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+            if selection == '1':
+                complete = False
+                action = ''
+                donation = None
+                while not complete:
+                    action = input("Type a name or 'list': ")
+                    check_for_quit(action)
+                    # donor IS found
+                    a = action.lower()  # line too long without this
+                    if any(x for x in _donors if x[0].lower() == a):
```
robalford: Since each donor in `_donors` should be unique, you should be able to loop through the donor list once, check for the donor name and either append the donation to an existing donor, or add a new one. I don't think you want to use `any()` here since you shouldn't have duplicate donors. Taking this approach should let you eliminate some of the duplicated code in your `if` and `else` blocks here.  You may also want to consider factoring your donor look up code into its own function.

students/carlos_novoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+                        break
+                    # list requested
+                    elif action == 'list':
+                        print(', '.join(d[0] for d in _donors))
+                        continue
+                    # donor NOT found
+                    else:
+                        donation = get_donation()
+                        donation = format_donation(donation)
+                        complete = add_donor(action, donation)
```
robalford: You don't appear to be using the `complete` variable anywhere so you can just call the `add_donor` function here without assigning it

PR #282, opened 2018-03-12
Lesson 3 Underway

stuartkershaw:


students/stuartkershaw/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,31 @@
+def exchange_first_last(seq):
+    seq[0], seq[-1] = seq[-1], seq[0]
+    return seq
```
natasha-aleksandrova: neat solution!

how would you solve using slicing only? (seq[-1], seq[0] is indexing, where slicing is seq[start: end: step])

students/stuartkershaw/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,31 @@
...
+    return seq[4:-4:2]
+
+def reverse_sequence(seq):
+    return seq[::-1]
+
+def return_specified_thirds(seq):
+    index = int(len(seq) / 3)
+    seg_end = seq[:index]
+    seg_start = seq[index:]
+    return seg_start + seg_end
```
natasha-aleksandrova: I'm assuming the assert fails right? (it looks like it's missing the mid third.)

PR #284, opened 2018-03-12
Lesson04 Trigram.py submission

csdotson:
Completed code for trigrams.py. It seems to work, but I definitely feel like there's some serious inefficiency in how it's put together :)


students/csdotson/lesson04/trigrams.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
+# Lesson 4 - Trigrams
+import random
+
+def run():
+    input_file = input("Please enter the name of a file (with extension):\n")
+    read_file(input_file)
```
natasha-aleksandrova: I would suggest a better flow for your main entry point.

right now looking here last logic is to read a file, doesn't really tell me anything else is going on.. and then each following function calls next function. it is hard to follow and you don't get a full view at what the code is doing.

what you could do:
```
text = read_file(input_file)
story = make_story(text) 
print("here is your story: ", story)
```


```
def make_story(text):
    text_hash = create_hash(text)
    ....
    return fix_punctuation_and_display(story)
```

students/csdotson/lesson04/trigrams.py
```diff
@@ -0,0 +1,77 @@
...
+            word_array[index+1] = word_array[index+1].capitalize()
+    story = ' '.join(word_array)
+
+    print("\nAnd, here's your story:\n\n", story)
+
+
+ending_punctuation = [".", "?", "!"]
+other_punctuation = [",", ";", ":"]
+punctuation_remove = ["(", ")", "[", "]", '"', "#", "'"]
+replace_with_space = ["\n", "--"]
```
natasha-aleksandrova: I would put the at the top of the file, and even make them constants (ENDING_PUNCTUATION)

students/csdotson/lesson04/trigrams.py
```diff
@@ -0,0 +1,77 @@
...
+    fix_punctuation_and_display(story)
+
+
+def fix_punctuation_and_display(story):
+    # Capitalize beginning of story, remove spacing around punctuation, capitalize after ending punctuation
+    story = story.capitalize()
+    for entry in other_punctuation:
+        story = story.replace(f' {entry}', entry)
+    for entry in ending_punctuation:
+        story = story.replace(f' {entry}', entry)
```
natasha-aleksandrova: curious why the two for loops? they are doing the same thing

if you wanted to keep var names for clarity you could do -> 
`for entry in other_punctuation + ending_punctuation`

students/csdotson/lesson04/trigrams.py
```diff
@@ -0,0 +1,77 @@
...
+        text_hash[' '.join(pattern)].append(text[index+2])
+    make_story(text, text_hash)
+
+
+def make_story(text, text_hash):
+    # Generate story from word hash
+    random_starting_point = random.randint(0,len(text)-1)
+    story = ' '.join(text[random_starting_point:random_starting_point+2])
+    while True:
+        next_key = ' '.join(story.split(' ')[-2:])
```
natasha-aleksandrova: split() splits on empty space by default

students/csdotson/lesson04/trigrams.py
```diff
@@ -0,0 +1,77 @@
...
+
+def create_hash(text):
+    # Create hash needed to produce trigrams
+    text_hash = {}
+    for index in range(len(text) - 1):
+        key = ' '.join(text[index:index+2])
+        text_hash[key] = []
+    for index in range(len(text)-2):
+        pattern = text[index:index+2]
+        text_hash[' '.join(pattern)].append(text[index+2])
```
natasha-aleksandrova: I'm not sure why you need the two for loops here, should be able to create the look up with just one.

does this have something to do with `if text_hash[next_key] == []`?

if you're trying to safely to exit then you can `if not text_hash.get(next_key) or`

students/csdotson/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,107 @@
...
+    "1": send_thank_you,
+    "2": create_report,
+    "3": write_letters,
+    "q": quit,
+}
+
+thank_you_menu = {
+    "1": add_donation,
+    "2": list_donors,
+    "q": quit,
```
natasha-aleksandrova: be careful using `quit` for var name; it's a built-in function (see github highlight differently)

students/csdotson/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def create_report():
+    # Print formatted report showing donors and donation stats
+    header = '{:20}|{:^15}|{:^15}|{:>15}'.format("Donor Name", "Total Given", "Num Gifts", "Average Gift")
+    print()
+    print(header)
+    print("-" * len(header))
+
+    for donor in donations:
```
natasha-aleksandrova: more common pattern is to use `items` if you're going to use key and values:
`for donor, donor_data in donations.items()`

students/csdotson/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,107 @@
...
+    if response not in donations.keys():
+        new_donor = response
+        new_donation = input(f"Please enter a donation amount for {new_donor}: ")
+        donations[new_donor] = [float(new_donation)]
+        print_email(new_donor, new_donation)
+    else:
+         current_donor = response
+         new_donation = input(f"Please enter a donation amount for {current_donor}: ")
+         donations[current_donor].append(float(new_donation))
+         print_email(current_donor, new_donation)
```
natasha-aleksandrova: you can take the print_email out of if/else block since its duplicated

also, no need to rename `response` var, you can call it `donor_entry`.. so basically to simplify:
```
donor_name_entry = input("Please enter donor name")
donation = float(input(f"Please enter a donation amount for {donor_name_entry}: "))
if donor_entry in donations:  # no need to call keys()
    donations[donor_name_entry] = [donation]
else:
    donations[donor_name_entry].append(donation)
print_email(donor_name_entry, donation)
```

you can even make it simpler if you use defaultdict

natasha-aleksandrova:
please be sure to submit both assignments in the course website.

PR #286, opened 2018-03-13
Question

JerHsi:
@natasha-aleksandrova Not sure if I am on the right track.  Also, what's the reason I can't get my main() or prompt() to run when this script is run?  I keep getting "send_thank_you()" prompt instead.

### DIFF
--- a/students/JerryH/Lesson09/mailroom.py
+++ b/students/JerryH/Lesson09/mailroom.py
@@ -2,14 +2,10 @@
 
 class Donor:
 
-    # Class Variables
-    #num_of_donors = 0
-
     def __init__(self, first, last, donations = None):
         self.first = first
         self.last = last
         self.donations = donations
-        # num_of_donors += 1
 
     @property
     def full_name(self):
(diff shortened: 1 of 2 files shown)


students/JerryH/Lesson09/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+db = DonorBook([d1, d2, d3, d4, d5, d6])
+
+QUIT_OPT = '4'
+
+selection_map = {
+    "1": db.send_thank_you(),
+    "2": db.create_report(),
+    "3": db.send_letters(),
+    "4": db.quit_program()
+	}
```
natasha-aleksandrova: the reason it goes straight to thank you is because of this code here.

when it creates the selection_map, it executes each of your functions, send thank you will be the first one and it gets stuck here.


remember that we don't want to call the function in the selection_map, we want to map to its object and execute it later once we made the selection.

So it should be -> `"1": db.send_thank_you,` not `"1": db.send_thank_you(),`

students/JerryH/Lesson09/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+    def create_report(self):
+        pass
+
+
+    def send_letters(self):
+        pass
+
+
+    def quit_program(self):
+    	print("Thanks for using my script! Bye!")
```
natasha-aleksandrova: you are on the right track here!

You `Donor` class is good, it should represent a single donor object just like you're doing.


Your `DonorBook` is also on the right track, but as mentioned in the course assignment:
"There should be no input function in the class that holds the data!"

So keep things directly related to your Donor DB (`DonorBook`) in the class, like creating generating thank you letter text.. but keep the user interaction code like `input` out of the class.

natasha-aleksandrova:
Let me know if you need some more feedback from me here! (or questions)

PR #287, opened 2018-03-13
submit mailroom.py

tricaonguyen:
Hi Natasha,
I made some changes that you suggested in merged #220.

### DIFF
--- a/students/tri_nguyen/lesson05/mailroom.py
+++ b/students/tri_nguyen/lesson05/mailroom.py
@@ -1,0 +1,201 @@
+#!/usr/bin/env python3
+
+# ------------------------------------------- #
+# Student:   Tri Nguyen
+# Instructor: Natasha Aleksandrova
+# Date:  09-Mar-2018
+# ------------------------------------------- #
+
+# Import stuff
+from operator import itemgetter
+import time
+
+# Get local time
+local_time = time.localtime()
+
+# Declare variables and data structure
+
+donor_dict = { #  [0] == total donation, [1] == number of gifts, [2] == average
+
+    'William Gates, III': [653784.49, 2, 326892.24],
+    'Mark Zuckerberg': [16396.10, 3, 5465.37],
+    'Jeff Bezos': [877.33, 1, 877.33],
+    'Paul Allen':[708.42, 3, 236.14],
+    'Tri Nguyen': [100.00, 1, 100.00]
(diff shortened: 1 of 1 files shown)


students/tri_nguyen/lesson05/mailroom.py
```diff
@@ -0,0 +1,201 @@
...
+# Declare variables and data structure
+
+donor_dict = { #  [0] == total donation, [1] == number of gifts, [2] == average
+
+    'William Gates, III': [653784.49, 2, 326892.24],
+    'Mark Zuckerberg': [16396.10, 3, 5465.37],
+    'Jeff Bezos': [877.33, 1, 877.33],
+    'Paul Allen':[708.42, 3, 236.14],
+    'Tri Nguyen': [100.00, 1, 100.00]
+    }
```
natasha-aleksandrova: for a little more clarity it would be a good idea to use a dict for the value as well, like so:
`{"Jeff Bezos": {"total": 877.33, "num_of_gifts": 3, "avg": 877.33}, }`
your code is now well documented and easy to understand.

students/tri_nguyen/lesson05/mailroom.py
```diff
@@ -0,0 +1,201 @@
...
+    ''')
+
+
+def display_donor_list():
+    ''' this function displays list of donors
+        from donor_dict '''
+
+    donor_name = enumerate([donor for donor in donor_dict])
+    for idx, name in donor_name:
+        print(idx + 1, '-', name)
```
natasha-aleksandrova: -> `for idx, name in enumerate(donor_dict):`

dictionaries iterate over keys by default

students/tri_nguyen/lesson05/mailroom.py
```diff
@@ -0,0 +1,201 @@
...
+
+    if ',' in name:
+        name = name.replace(',', '')
+        name = name.replace(' ', '_')
+        name = name + '.txt'
+        return name
+    else:
+        name = name.replace(' ', '_')
+        name = name + '.txt'
+        return name
```
natasha-aleksandrova: I think you can just have all of this in one block:
```
def format_filename(name):
    name = name.replace(',', '')
    name = name.replace(' ', '_')
    return name + '.txt'
```
replace will not do anything if `,` is not in name

PR #288, opened 2018-03-13
Lesson 03 - List Lab

stuartkershaw:


students/stuartkershaw/lesson03/slicing_lab.py
```diff
@@ -1,5 +1,5 @@
 def exchange_first_last(seq):
-    seq[0], seq[-1] = seq[-1], seq[0]
+    seq[-1], seq[0] = seq[::len(seq) - 1]
```
natasha-aleksandrova: here is the solution with slicing only : `seq[-1:] + seq[1:-1] + seq[:1]`

students/stuartkershaw/lesson03/list_lab.py
```diff
@@ -0,0 +1,89 @@
...
+print(fruit)
+
+fruit = fruit * 2
+
+print(fruit)
+
+delete_fruits = input('Please enter a fruit to delete: ')
+
+while not delete_fruits in fruit:
+    delete_fruits = input('That fruit isn\'t available! Please enter another:  ')
```
natasha-aleksandrova: good practice to use vars that reflect plural items vs a single item

`fruit` should be `fruits`
`delete_fruits` should be `delete_fruit` or `fruit_to_delete`

PR #289, opened 2018-03-13
add mailroom.py

iamAndrewKim:


students/AndrewKim/lesson03/mailroom.py
```diff
@@ -0,0 +1,52 @@
...
+    
+    
+while True:
+    choice = input("Please choose from a menu of 3 actions: 'Send a Thank You', 'Create a Report' or 'quit': ")
+    if choice == "Send a Thank You":
+        thankYou()
+    elif choice == "Create a Report":
+        report()
+    elif choice == "quit":
+        break
```
natasha-aleksandrova: for user's experience and easier code - it is better to use numbers for menu selection 

it is a bit much to ask users to type out case sensitive strings

students/AndrewKim/lesson03/mailroom.py
```diff
@@ -0,0 +1,52 @@
+
+donors = ["Andrew Kim", "Jamie Park", "Caroline Yeung", "Billy Yan"]
+totalAmt = [5.0, 4.0, 3.0, 2.0]
+numGifts = [1, 1, 1, 1]
+
+def thankYou():
```
natasha-aleksandrova: var names and function names should be lower case with underscores (snake_case)

PR #291, opened 2018-03-13
[DennisLee][L5][Mailroom P3] Ready for submission.

DennisLee-DennisLee:


students/DennisLee/lesson05/mailroom.py
```diff
@@ -0,0 +1,212 @@
...
+    Display the menu of choices for donor management.
+
+    :return:  None.
+    """
+    # create a dictionary of menu items, menu text, and menu caller functions
+    options = ("Send a thank you", "Create a report", 
+               "Send letters to everyone", "Quit")
+    funcs = (send_thank_you, create_a_report, send_all_letters, exit_screen)
+    choices = {str(x): {'option': y, 'function': z}
+               for x, y, z in zip(range(1, len(options) + 1), options, funcs)}
```
robalford: You have a really great handle on the syntax and capabilities of the various comprehension patterns in Python, and you're digging into some of the cool functional programming tools as well. What can be tricky is figuring out when to use comprehensions versus standard loops in your code. There are both performance and readability considerations to take into account. 

In some of the cases here, I would argue that simply defining your data structures as dict literals rather than building from existing tuples in your comprehensions would be a better choice for readability. It's also not common to see comprehensions that are not assigned to values, as in your print comp below. @natasha-aleksandrova do you have any insight on that?

That said, I'm not seeing a lot of practical use cases for comprehensions in your current mailroom version, so it's great that you're experimenting with them in these ways. I'd just offer that in the 'real world' it's always important to consider whether there's a simpler, cleaner or more readable solution that doesn't significantly impact the performance of your program.

natasha-aleksandrova: you're absolutely right Rob, comprehensions return a value and using them in generating print statements isn't really the pattern.

DennisLee-DennisLee: Yes Rob, I didn't many opportunities for comprehensions, so I did things that went against my intuition, like converting 2-line **for ...: print(...)** loops into one-liners and building up the menu option structures using comprehensions instead of defining them like that directly. I'm glad you recognized that. :)

PR #293, opened 2018-03-14
Dustin_L - Lesson07 - HTML Render

rdlaner:
Lesson 7, HTML render assignment complete.

Any feedback welcome. I'm interested in what you think about the NOTEs on lines 140 and 160 and adhering to the Liskov Substitution Principle.


students/Dustin_L/lesson07/html_render.py
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env python3
+"""HTML Render Module
+
+This module contains all of the functions for the HTML Render module
+"""
+
+
+class Element(object):
```
robalford: In Python3 it's not necessary to explicitly subclass from `object`. Base classes are conventionally defined like this:
```
class Element:
```

students/Dustin_L/lesson07/html_render.py
```diff
@@ -0,0 +1,180 @@
...
+    """Element class for rendering an HTML element"""
+    tag = ''
```
robalford: It's not necessary to define an empty attribute in Python in order to establish a shared interface for subclasses, though I can see how the instructions for the assignment lead you that way here. But I think the `tag` attribute could be omitted and defined in each subclass.

students/Dustin_L/lesson07/html_render.py
```diff
@@ -0,0 +1,180 @@
...
+class TitleElement(OneLineElement):
+    """Title type Element"""
+    tag = 'title'
+
+
+class SelfClosingElement(Element):
+    """Self-Closing type Element"""
+    def __init__(self, content=None, **attrs):
+        super().__init__(content=None, **attrs)
+
```
robalford: It's not necessary to override `__init__` here or include the getter and setter below, since `content` defaults to `None` in the base class, I believe you can remove these methods/properties without impacting your program.

students/Dustin_L/lesson07/html_render.py
```diff
@@ -0,0 +1,180 @@
...
+# NOTE: Shouldn't this sublclass maintain the same init params as Element
+#       in order to adhere to the Liskov Substitution principle?
+class AnchorElement(OneLineElement):
+    """Anchor link type element"""
+    tag = 'a'
+
+    def __init__(self, link, content):
+        super().__init__(content, href=link)
+
```
robalford: I'm not quite sure if this gets to your question and maybe @natasha-aleksandrova has some insight into the application of LSP in dynamically typed languages and whether that applies here but a common pattern in Python is to use *args and **kwargs when overriding and/or extending the `__init__ ` method in a subclass, so this could be rewritten as:
```
    def __init__(self, link, *args, **attrs):
        attrs['href'] = link
        super().__init__(*args, **attrs)
```

PR #295, opened 2018-03-14
Updated lesson03 labs

githubtater:


students/Roy_Tate/lesson03/slice_lab.py
```diff
@@ -1,21 +1,7 @@
...
 def exchange_first_last(sequence):
-    first = sequence[0]
-    print('first: ', first)
-    last = sequence[-1]
-    print('last: ', last)
-    middle = sequence[1:len(sequence) -1]
-    print(middle)
-    new_sequence = last + middle + first
+    new_sequence = sequence[-1] + sequence[1:len(sequence) - 1] + sequence[0]
     return new_sequence
```
natasha-aleksandrova: here is the pure slicing solution: `seq[-1:] + seq[1:-1] + seq[:1]`

PR #296, opened 2018-03-14
lesson05 - fixes based on feedback

ghassan24601:


students/ghassan/lesson05/new_mailroom.py
```diff
@@ -25,12 +25,12 @@ def send_thankyou():
...
+        donors[donor_name].append(float(donation_amount))
         print('Thank you {} for your generous donation of {}'.format(
             donor_name, sum(donors[donor_name])))
     except KeyError:
-        donors[donor_name] = []
-        donation_amount = input('Please enter a donation amount: ')
-        donors[donor_name].append(float(donation_amount))
+        donors[donor_name] = [float(donation_amount)]
         print('Thank you {} for your generous donation of {}'.format(
             donor_name, sum(donors[donor_name])))
```
natasha-aleksandrova: to take this further.. your KeyError should really only be around key access and you have dupe print statement. so this could look like:
```
donation_amount = input('Please enter a donation amount: ')
try:
    donors[donor_name].append(float(donation_amount))
except KeyError:
    donors[donor_name] = [float(donation_amount)]
print('Thank you {} for your generous donation of {}'.format(
             donor_name, sum(donors[donor_name])))
```

above gives additional simplicity and isolates KeyError possibility to the actual potential problem area.

PR #298, opened 2018-03-14
first commit of Kata exercise

JeroenSweerts:
Hi, 
Please review my code for the Kata Fourteen assignment.
Not sure why I keep getting a 'b' in front of each word.

Thanks.

Jeroen

### DIFF
--- a/students/JeroenSweerts/Lesson4/kata.py
+++ b/students/JeroenSweerts/Lesson4/kata.py
@@ -1,0 +1,40 @@
+import os
+import random
+#get path of current work directory
+os.chdir(os.path.dirname(__file__))
+cwd = os.getcwd()
+
+#open the text file and read it into a list of words
+f = open(cwd + '\sherlock.txt','rb')
+text = f.read()
+words = text.split()
+words = [str(i) for i in words]
+f.close()
+
+#create a dictionary of all 2 word combinations as the key and the following word as value
+words_dict = {}
+for i in range(len(words)-3):
+    try:
+        words_dict[words[i] + '_' + words[i+1]].append(words[i+2])
+    except:
+        words_dict[words[i] + '_' + words[i+1]] = [words[i+2]]
+
+#chose arbitrary word pair as a starting point
+start_i = random.randint(0,len(words)-3)
+word1 = words[start_i]
+word2 = words[start_i+1]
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
the `b` in front of a string is for bytes:
```
>>> s = b'hi'
>>> s
b'hi'
>>> type(s)
<class 'bytes'>
>>> s = 'hi'
>>> type(s)
<class 'str'>


>>> with open('hi.txt', 'r') as f:
...     out = f.read()
...
>>> type(out)
<class 'str'>
>>> with open('hi.txt', 'rb') as f:
...     out = f.read()
...
>>> type(out)
<class 'bytes'>
>>>
```

students/JeroenSweerts/Lesson4/kata.py
```diff
@@ -0,0 +1,40 @@
...
+#get path of current work directory
+os.chdir(os.path.dirname(__file__))
+cwd = os.getcwd()
+
+#open the text file and read it into a list of words
+f = open(cwd + '\sherlock.txt','rb')
+text = f.read()
+words = text.split()
+words = [str(i) for i in words]
+f.close()
```
natasha-aleksandrova: it is preferred to use `with` block when working with files
http://www.pythonforbeginners.com/files/with-statement-in-python

so this would look like:
```
with open(os.path.join(cwd, filename), 'r'):
    text = f.read()
words = text.split()

```
notice in above only read() is under `with` block and the use of `os.path.join` you want to avoid manually creating filepaths.

students/JeroenSweerts/Lesson4/kata.py
```diff
@@ -0,0 +1,40 @@
+import os
+import random
+#get path of current work directory
+os.chdir(os.path.dirname(__file__))
+cwd = os.getcwd()
```
natasha-aleksandrova: pep8: you want two empty lines after imports

students/JeroenSweerts/Lesson4/kata.py
```diff
@@ -0,0 +1,40 @@
...
+
+#open the text file and read it into a list of words
```
natasha-aleksandrova: pep8: need one space after `#`

PR #299, opened 2018-03-15
Exception handling in mailroom3

MikeS-1:
couldn't find a sensible place to add list comprehensions

### DIFF
--- a/students/MikeShand/lesson05/mailroom3.py
+++ b/students/MikeShand/lesson05/mailroom3.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+import sys
+from collections import defaultdict
+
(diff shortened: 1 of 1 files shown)


students/MikeShand/lesson05/mailroom3.py
```diff
@@ -0,0 +1,81 @@
...
+
+    elif new_donor.lower() == 'q':
+        return
+
+    else:
+        try:
+            amount = float(input('Please enter the donation amount:'))
+            donors[new_donor].append(amount)
+            print('Thank you {} for your generous donation of ${:.2f}'.format(new_donor, amount))
+        except ValueError:
```
robalford: You generally want to catch the exception as close as you can to where it occurs so I'd move the `append` and `print` statements out of the `try` block

PR #300, opened 2018-03-15
Submitting Mailroom2.py

wesmoy17:


students/WesleyM/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,55 @@
+donors = [
+  {'name': 'Alice Adams Ron', 'donation': [20]} ,
+  {'name': 'Bob Be-Lake', 'donation': [100]},
+  {'name': 'Charles Cruz', 'donation': [30, 50, 10]},
+  {'name': 'Denise Dnice', 'donation': [10, 5]},
+  {'name': 'Edward Eduardo', 'donation': [25, 20, 20, 20, 10]}
+  ]
+
```
robalford: I would suggest using a dict as your main data structure here rather than a list of dicts

students/WesleyM/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,55 @@
...
+  {'name': 'Edward Eduardo', 'donation': [25, 20, 20, 20, 10]}
+  ]
+ 
+
+def send_ty():
+  donor = input("Enter a donor name or input 'List' for a list of donors\n>")
+  if donor.lower() == "list":
+    for names in donors:
+        print(names['name'])
+    use_input()
```
robalford: I think it's a bit confusing to go back to the original prompt here and would consider placing the `donor` prompt into a while loop so you can enter a name directly after viewing the list donors

students/WesleyM/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,55 @@
...
+  else:
+    donation = float(input(("Enter a donation amount for {} \n>").format(donor)))
```
robalford: Don't need the parentheses around the string here

students/WesleyM/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,55 @@
...
+        break
+    else:
+      donors.append({'name': donor, 'donation': [donation]})
+    print("Thank you {} for your donation of ${:.2f}".format(donor, donation))
+  use_input()
+  
+def letters():
+  for name in donors:
+    print('Printing Letter to {a}'.format(a = name['name']))
+    f = open('{a}.txt'.format(a = name['name']), 'w')
```
robalford: We'll learn more about the mechanics of this later in the course, but it's good idea to use `with` when working with files because it will automatically close them for you when you exit the with block, e.g.:
```
with open(the_filename, 'w') as outfile:
    outfile.write(something)
```

students/WesleyM/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,55 @@
...
+    f.write('Dear Ms./Mrs./Mr./Dr. {x}, \n We are thankful for your donation of ${y}. Your donation will be used for (insert harmful activity here). We hope you donate again soon!'.format(x = name['name'], y = sum(name['donation'])))
+    f.close()
```
robalford: Remember you can use triple quotes to preserve returns and indents in your strings and stay under the Pep8 line character limit.

students/WesleyM/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,55 @@
...
+    total = sum(rep['donation'])
+    num = len(rep['donation'])
+    print("|{:<20}|{:>15.2f}|{:>15}|{:>15.2f}|".format(rep['name'], total, num, total/num))
+  use_input()
+
+choices = {1: send_ty, 2: report, 3: letters}
+
+def use_input():
+  action = int(input("Choose one of four actions: \n 1. Send a Thank You \n 2. Create a Report \n 3. Send Letter to Everyone \n 4. Quit \n>"))
+  while (action not in [1, 2, 3, 4]):
```
robalford: You don't need the parentheses here for your `while` statement. I would also suggest adding a fourth exit function to your `choices` dict so this could become `while action not in choices` and eliminate the additional conditional expressions below.

PR #301, opened 2018-03-15
slices assignment

natrod22:


students/NatalieRodriguez/Lesson03/slices.py
```diff
@@ -0,0 +1,74 @@
...
+dog_list = ['Virgil', 'River', 'Kibson', 'Dexter', 'China', 'Lucy', 'Percy']
+bird_list = ['Bluejay', 'Sparrow', 'House Finch', 'Cardinal', 'Nuthatch', 'Tufted Titmouse']
+number_list = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
+
+#________________________________________________________________________
+#Exchange the first item in the sequence with the last item in the sequence using slices.
+def first_last_exchanged(dog_list):
+    print("Original Sequence:",dog_list)
+    dog_list[0], dog_list[-1] = dog_list[-1], dog_list[0]
+    return dog_list
```
natasha-aleksandrova: how can you solve it using slicing only?

`[-1]` / `[0]` are indexing
`[start:stop:step]` is slicing (hint: its a one liner)

students/NatalieRodriguez/Lesson03/slices.py
```diff
@@ -0,0 +1,74 @@
...
+print(first_last_exchanged(dog_list))
+
+
+#_______________________________________________________________________
+#Print the sequence with every other item removed
+
+def every_other_removed(dog_list):
+    print("Original Sequence:", dog_list)
+    del dog_list[1::2]
+    return dog_list
```
natasha-aleksandrova: can be simplified to one liner: `return dog_list[::2]`

students/NatalieRodriguez/Lesson03/slices.py
```diff
@@ -0,0 +1,74 @@
...
+
+print(reverse_list(number_list))
+
+#________________________________________________________________________
+#middle third, last third, first third.
+
+def thirds(bird_list):
+    print("Original Sequence:", bird_list)
+    bird_list[:2], bird_list[2:4], bird_list[4:] = bird_list[2:4], bird_list[4:], bird_list[:2]
+    return bird_list
```
natasha-aleksandrova: do not assign back to sequence, that's called a side-effect. you should return a new sequence .

natasha-aleksandrova: this function will break if you pass in a tuple or different sized sequences. (hint: you should calculate what a third of a sequence is first).

students/NatalieRodriguez/Lesson03/slices.py
```diff
@@ -0,0 +1,74 @@
...
+dog_list = ['Virgil', 'River', 'Kibson', 'Dexter', 'China', 'Lucy', 'Percy']
+bird_list = ['Bluejay', 'Sparrow', 'House Finch', 'Cardinal', 'Nuthatch', 'Tufted Titmouse']
+number_list = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
+
+
+assert first_last_exchanged(dog_list) == ['Percy', 'River', 'Kibson', 'Dexter', 'China', 'Lucy', 'Virgil']
+assert every_other_removed(dog_list) == ['Percy', 'Kibson', 'China', 'Virgil']
+assert part_three(number_list) == [4, 6, 8, 10]
+assert reverse_list(number_list) == [15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0]
+assert thirds(bird_list) == ['House Finch', 'Cardinal', 'Nuthatch', 'Tufted Titmouse', 'Bluejay', 'Sparrow']
```
natasha-aleksandrova: here are some overall comment:
* don't define your sequences twice, at the top and at the bottom here. define at the bottom only where you are doing your tests.
* your asserts should ideally pass in each type of sequence (tuple, string, list) which means that your code should work for all of those types. that is fairly simple with slicing and returning new sequence
* following on above statement, your input args in your functions should be called something like `seq` and not dog_list etc.

PR #302, opened 2018-03-16
Lesson9

JerHsi:


students/JerryH/Lesson09/mailroom.py
```diff
@@ -0,0 +1,156 @@
...
+        self.last = last
+        self.donations = donations
+
+    @property
+    def full_name(self):
+        return "{} {}".format(self.first, self.last)
+
+    @property
+    def donor_history(self):
+        return sum(self.donations)
```
natasha-aleksandrova: maybe call it `def total_donations`?

students/JerryH/Lesson09/mailroom.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def add_donation(self, amount):
+        return self.donations.append(amount)
+
+
+class DonorBook:
+
+    def __init__(self, donors = []):
```
natasha-aleksandrova: better to make default be `None` -> `def __init__(self, donors=None)`
it is dangerous to have mutable sequence as a default param, and you're already checking for `is None`

students/JerryH/Lesson09/mailroom.py
```diff
@@ -0,0 +1,156 @@
...
+    def add_donor(self, donor):
+        # if donor not in self.donors:
+        self.donors.append(donor)
+
+    def get_all_donor_names(self):
+        donor_name_list = []
+        for donor in self.donors:
+            #print("-->" + donor.full_name)
+            donor_name_list.append(donor.full_name)
+        return donor_name_list
```
natasha-aleksandrova: great place to use comprehension!
`return [donor.full_name for donor in self.donors]`

students/JerryH/Lesson09/mailroom.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def list_all_donor_names_sorted(self):
+        all_names = ""
+        for each_donor_name in sorted(self.get_all_donor_names()):
+            all_names += "{}\n".format(each_donor_name)
+        return(all_names)
```
natasha-aleksandrova: i think you could do 
`return "\n".join(sorted( self.get_all_donor_names()))`

students/JerryH/Lesson09/mailroom.py
```diff
@@ -0,0 +1,156 @@
...
+                first, last = donor_name.split(" ")
+                self.add_donor(Donor(first, last, [donation]))
+            except ValueError:
+                print("Please enter both of your \"First Name\" and \"Last Name\"")
+        else:
+            for donor in self.donors:
+                if donor.full_name == donor_name:
+                    donor.add_donation(donation)
+
+        print("Thank You Email:  Thanks for the donation!\n\n")
```
natasha-aleksandrova: it is best to extract some of this code outside of the class, specifically one dealing with user interaction

PR #303, opened 2018-03-16
lesson_05 submission

slickw1lly:


students/khtruong/lesson_05/mailroom_v3.py
```diff
@@ -0,0 +1,146 @@
...
+
+def amount_input():
+    """Return prompt asking for donation amount."""
+    while True:
+        try:
+            amount = float(input('Enter donation amount! > '))
+        except ValueError:
+            print('\nPlease enter dollar amount and NOT text!')
+        else:
+            break
```
natasha-aleksandrova: 👍

students/khtruong/lesson_05/mailroom_v3.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def send_thankyou_email():
+    """Return a menu selection to send thank you email to donor."""
+    menu_selection(thankyou_prompt, thankyou_dict)
+
+
+def list_donors():
+    """Return all donors name."""
+    [print(donor) for donor in data_dict.keys()]
```
natasha-aleksandrova: this isn't really what list comprehensions are intended to be used for.

comprehensions have a return value, and using it as a shortcut to just print a value is wasteful.

here is an alternative way it could be used with prints:
https://stackoverflow.com/questions/37084246/printing-using-list-comprehension

students/khtruong/lesson_05/mailroom_v3.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def summarize_donation():
+    '''Return total given, number of gifts, and average gifts for each donor'''
+    summarized_list = [{'donor_name': donor_key,
+                        'total_given': sum(data_dict.get(donor_key)),
+                        'num_gifts': len(data_dict.get(donor_key)),
+                        'avg_gifts': (sum(data_dict.get(donor_key)) /
+                                      max(len(data_dict.get(donor_key)), 1))
+                        } for donor_key in data_dict.keys()]
```
natasha-aleksandrova: nice job on utilizing list comprehension. here are some styling suggestions:

```
summarized_list = [{"donor_name": k, "total_given": sum(v), "num_gifts": len(v), "avg_gifts": sum(v) / len(v)} for k, v
                   in data_dict.items()]
```

use `items` if you need both key and value. it's ok to use shorter variables in comprehensions

PR #306, opened 2018-03-16
Working mailroom.py

hortw1991:


students/thorn/lesson03/mailroom.py
```diff
@@ -9,9 +9,9 @@
 # Global that holds donor history and the amounts donated.
 DONOR_HISTORY = [
     ['Tom Horn',        [599.23, 1000.00]],
-    ['Theo Hartwell',   [1000.00, 900.88, 23.6]],
-    ['James Jones',     [8723.22, 27167.22, 91817.66]],
-    ['Sterling Shepard',[90012.32, 2312.24]],
+    ['Theo Hartwell',   [0.01, 0.01, 0.1]],
+    ['Bailey Kimmitt',  [8723.22, 27167.22, 91817.66]],
+    ['Paul Hubbell',    [90012.32, 2312.24]],
     ['David Beckham',   [1817266.11, 123123.66, 111335.112]]
```
natasha-aleksandrova: all caps indicates a constant, in our case donor data is not a constant so it should be defined as all lower case.

students/thorn/lesson03/mailroom.py
```diff
@@ -26,33 +26,69 @@ def send_thanks():
...
+        1 - Send Thanks\n\
+        2 - Create Donor Report\n\
+        3 - Quit\n")
+        if choice == "1":
+            send_thanks()
+        if choice == "2":
+            create_report()
+        if choice == "3":
+            print("Quitting.")
+            break
```
natasha-aleksandrova: nice readable entry point here

if / elif would be more appropriate here vs several if statements as only one of them will be true

students/thorn/lesson03/mailroom.py
```diff
@@ -26,33 +26,69 @@ def send_thanks():
...
+        quit()
+
+    if name_request.lower() == "return":
+        return
+
+    # List request - only donors returned.
+    if name_request.lower() == 'list':
+        for donors in DONOR_HISTORY:
+            print(donors[0])
+        return
```
natasha-aleksandrova: if / elif here as well

it is more syntactically correct to have `for donor in DONOR_HISTORY` as you are getting a single donor when you're iterating

( i would really call it `for donor in donors`)

students/thorn/lesson03/mailroom.py
```diff
@@ -26,33 +26,69 @@ def send_thanks():
...
+
+    name_request = name_request.title()
+    donation_amount = float(input(("Please enter a donation amount for {}: ".format(name_request.title()))))
```
natasha-aleksandrova: no need to call `.title()` again here

students/thorn/lesson03/mailroom.py
```diff
@@ -26,33 +26,69 @@ def send_thanks():
...
+    # Add to existing donor.  i is the donor location in the nested list.
+    for i, donors in enumerate(DONOR_HISTORY):
+        if donors[0].title() == name_request.title():
+            DONOR_HISTORY[i][1].append(donation_amount)
+            # Break to stop it from creating i numbers of the person.
+            break
+        # Create new donor.
+        else:
+            DONOR_HISTORY.append([name_request.title(), [donation_amount]])
+            break
```
natasha-aleksandrova: you don't need enumerate, here is a simplified version:
```
for donor in donors:
    if donor[0].title() == name_request:  # name_request already has title applied
        donor[1].append(donation_amount)
```

your `else` has a bug. run your program, select send thanks then enter David Beckham and donation amount, now select create report - notice that david beckham is listed twice. Can you spot why?

students/thorn/lesson03/mailroom.py
```diff
@@ -26,33 +26,69 @@ def send_thanks():
...
+    - $ Total - 27 spaces in, right justified
     """
+    print("Donor:                    |    $ Total     |   Donations   |   $ Average   |")
+    print("-"*76)    
+    for item in DONOR_HISTORY:
+        amt_total = float(sum(item[1]))
+        num_total = int(len(item[1]))
+        # Thousand separator as default. Careful with the space if we get some big donors.
+        print("{:<26}|${:>15,.2f}|{:>15}|{:>15,.2f}".format(item[0], amt_total, num_total, amt_total/num_total))
+    return
```
natasha-aleksandrova: you don't need the return here

students/thorn/lesson03/mailroom.py
hortw1991: Thanks for all the suggestions!  I changed this to add the donation amount to an existing donor if its found (in your suggested syntax), print the message and return.  Then outside of that it takes care of adding a new donor if an existing one was not found.  

It seems to be working correctly - but was this the solution you had in mind?  Thanks again!

    for donor in donors:
        if donor[0].title() == name_request:
            donor[1].append(donation_amount)
            print(f"Thank you {name_request.title()} for your donation of ${donation_amount:.2f}.")
            return

    donors.append([name_request, [donation_amount]])

PR #307, opened 2018-03-17
submitted first 'working' version of mailroom3.py (lesson05)

johnpharmd:


students/johnpharmd/lesson05/mailroom3.py
```diff
@@ -22,7 +20,6 @@
 
 # Processing
 def send_ty():
-    # global donors_amts
```
natasha-aleksandrova: I would recommend re-visiting this function (`send_ty`)

here are some main points:
* list comprehensions are used to collect data (it has return value) so using it for print statements or to see if exception is thrown isn't really the right pattern
* it took me a second to understand the reason for `ZeroDivisionError`. Simple is usually better, my understanding is that you're checking to see if there are any numbers in the name? use `response.isalpha()` instead.
* u should avoid things like lines 27 and 42. if you're having to write duplicate code it is probably needing a loop there. You can do while True and exit when the input you got is good
* there are too many nested if / else blocks which makes it hard to read to understand the logic. 
* remove commented out code, it makes it harder to read

students/johnpharmd/lesson05/mailroom3.py
```diff
@@ -183,28 +197,28 @@ def get_report():
...
+    # add comprehension here?
     for donor in donors_amts:
-        d1 = donors_amts[donor][1]
-        d2 = donors_amts[donor][2]
+        d1 = donors_amts[donor]['donations']
+        d2 = donors_amts[donor]['num_of_donations']
         print('{:<15}{}{:>10}{:>12}{}{:>11}'.format(donor, '  $',
               d1, d2, '  $', d1 // d2))
     print()
     program_run()
```
natasha-aleksandrova: your `program_run` needs to have a while True loop, then you wouldn't need to call this from every function.

PR #308, opened 2018-03-17
Add lesson05 exception exercise

22e88:


students/elmar_m/lesson05/except_exercise.py
```diff
@@ -0,0 +1,56 @@
...
+# parameters. (pun intended)
+
+langs = ['java', 'c', 'python']
+
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    more_joke = more_fun(langs[1])
+finally:
+    more_joke = more_fun(langs[2])
```
robalford: You want to use an `else` block for your call to `more_fun(langs[2])` since the instructions say to call the last `more_fun` function if there are no exceptions and the `finally` clause will run whether there are any exceptions or not

PR #309, opened 2018-03-17
Lesson 03 Complete

stuartkershaw:


students/stuartkershaw/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,91 @@
...
+
+seq_in = (2, 123.4567, 10000, 12345.67)
+
+file_out = ''.join(['file_', format(seq_in[0], '03')])
+float_dec_out = format(seq_in[1], '.2f')
+sci_out = format(seq_in[2], '.2e')
+sci_sig_out = format(seq_in[3], '.2e')
+
+out_msg = '{}: {}, {}, {}'.format(file_out, float_dec_out, sci_out,
+                                  sci_sig_out)
```
robalford: This could be written a bit more concisely and with a single call to `format()` like this:
```
print("file_{:03d} : {:10.2f}, {:.2e}, {:.3g}".format(2, 123.4567, 10000, 12345.67))
```

students/stuartkershaw/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,91 @@
...
+
+# Task Two
+
+file_out_alt = f"file_{seq_in[0]:03}"
+float_dec_out_alt = f"{seq_in[1]:.2f}"
+sci_out_alt = f"{seq_in[2]:.2e}"
+sci_sig_out_alt = f"{seq_in[3]:.2e}"
+
+out_msg_alt = f"{file_out_alt}: {float_dec_out_alt}, {sci_out_alt}" \
+               "{sci_sig_out_alt}"
```
robalford: Remember to add the `f` to the second line of the string or this will print out `{sci_sig_out_alt}`

students/stuartkershaw/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,91 @@
...
+print(out_msg_alt)
+
+# Task Three
+
+
+def formatter(tuple_in):
+    t = tuple_in
+    fstring = []
+    for i in t:
+        fstring.append('{:d}')
```
robalford: This could be accomplished without the loop by doing something like this:
```
['{:d}'] * len(t)
```

students/stuartkershaw/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,91 @@
...
+    return out_msg + ', '.join(fstring).format(*t)
+
+
+print(formatter((2, 3, 5, 7, 9)))
+
+# Task Four
+
+
+def formatter_four(tuple_in):
+    t = tuple_in
```
robalford: There's no need to reassign your function arguments like this, you can just use `tuple_in` inside the function block, which is also a more readable, meaningful variable name than `t`.

students/stuartkershaw/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,91 @@
...
+    collect = []
+    collect.append(format(t[3], '02'))
+    collect.append(str(t[4]))
+    collect.append(str(t[2]))
+    collect.append(format(t[0], '02'))
+    collect.append(str(t[1]))
+    return ' '.join(collect)
```
robalford: This function could be simplified by passing the rearranged items straight into `str.format()`

students/stuartkershaw/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,91 @@
...
+]
+
+set_table(test_list)
+
+
+# Task Six (Extra)
+
+nums = (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+
+print(('{:5d}' * len(nums)).format(*nums))
```
robalford: Nice!

stuartkershaw:
Thanks for the feedback @robalford - I've submitted the mailroom assignment to wrap up lesson 3.

students/stuartkershaw/lesson03/mailroom.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
+
+donors = [
+    'Anne Hathaway', 'Geronimo Jackson', 'Dingo Dogson',
+    'Leo Kottke', 'Flask McCreole', 'Piglet Norquist',
+    [[4000, 6250], [150, 400, 1250], [150, 300, 100],
+     [75, 125], [1000], [20, 15, 30]]
+]
+
```
robalford: I would suggest combining each donor and donation into a single list here for easier lookups and additions to the overall data structure. This should allow you to simplify some of the code in your 'set_donation`, `get_donor_name` and `show_donor_table` functions. 

Something like this (kind of like rows in a database):
```
donors = [
    ['Anne Hathaway', [4000, 6250]],
    ['Geronimo Jackson', [4000, 6250]],
]
```

students/stuartkershaw/lesson03/mailroom.py
```diff
@@ -0,0 +1,100 @@
...
+def get_selection():
+    options = 'Please select a, b, or c:\n\
+a) send a thank you\nb) create a report\nc) quit\n'
+    while True:
+        try:
+            selection = input(options)
+            while selection not in ['a', 'b', 'c']:
+                selection = input(options)
+            return selection
+        except ValueError:
```
robalford: I may be missing something but I don't see where a `ValueError` would be raised in the `try` block here. Seems like any 'bad' input would be filtered out in your `while` loop check

students/stuartkershaw/lesson03/mailroom.py
```diff
@@ -0,0 +1,100 @@
...
+    i = index
+    dl = donor_list
+    while True:
+        try:
+            donation = int(input('Please enter a donation amount: '))
+            while donation < 1:
+                donation = int(input('Please provide a whole \
+number greater than zero: '))
+            dl[-1][i].append(donation)
+        except ValueError:
```
robalford: Nice - this is a great place for a `try/except` that catches `ValueError`

PR #310, opened 2018-03-18
Lesson 3: Mailroom Exercise

itsabhister:


students/itsabhi/Lesson3/mailroom.py
```diff
@@ -0,0 +1,73 @@
+#! /usr/bin/env python3
+
+
+def sending_thanks(d):
+    donor_hist = d
```
robalford: There's no need to reassign the `d` argument here, but I would use `donor_hist` as your parameter name as it's more meaningful/readable

students/itsabhi/Lesson3/mailroom.py
```diff
@@ -0,0 +1,73 @@
...
+    unique_donors = []
+    for entry in donor_hist:
+        name = entry[0]
+        if name not in unique_donors:
+            unique_donors.append(name)
```
robalford: Can you think of a way to user the original donors list to check for unique donors without needing to build a second `unique_donors` list here?

students/itsabhi/Lesson3/mailroom.py
```diff
@@ -0,0 +1,73 @@
...
+
+    donor = input("Type the Full Name of the donor here: ")
```
robalford: Remember to give your user the 'list' option in your prompt

students/itsabhi/Lesson3/mailroom.py
```diff
@@ -0,0 +1,73 @@
...
+    if donor == "list":
+        print(unique_donors)
+        donor = input("Type the Full Name of the donor here: ")
+    if donor not in unique_donors:
```
robalford: I would consider using a `while` loop for your user input here so you don't have to repeat the prompts in your code

students/itsabhi/Lesson3/mailroom.py
```diff
@@ -0,0 +1,73 @@
...
+        donor_hist.append([donor])
+        amount = input("Enter Donation value: ")
+        donor_hist[-1].append([amount])
+#        print(donor_hist)
```
robalford: Remember to convert the `amount` to a numerical type so you can use it in your report calculations

students/itsabhi/Lesson3/mailroom.py
```diff
@@ -0,0 +1,73 @@
...
+
+    else:                       #donor appears in list of donors
+        for entry in donor_hist:
+            if donor == entry[0]:
+                amount = input("Enter Donation value: ")
+                entry[1].append(int(amount))
+                continue
+    #       print(donor_hist)
+    print("Hello {}, Just wanted to drop a note of thanks for your recent donation of ${}.".format(donor, amount))
+    return 1
```
robalford: You don't appear to be assigning or using these return values anywhere. See my note below on using `while True` to initialize an infinite loop for your user prompt

students/itsabhi/Lesson3/mailroom.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def second_integer(list):
+    return list[1]
```
robalford: It's a good idea to avoid using Python built-ins like `list` as variable names. One convention is to use a trailing underscore like `list_`, or something specific to your program like `donation_list`

students/itsabhi/Lesson3/mailroom.py
```diff
@@ -0,0 +1,73 @@
...
+def create_report(donors):
+    donor_hist = donors
+    donor_summary = []
+    print("{:20s} | {:10s} | {:10s} | {:10s}".format("Donor Name", "Total Given", "Num Gifts", "Average Gift"))
+    for entry in donor_hist:
+        count = len(entry[1])
+        donor = entry[0]
+        total = sum(entry[1])
+        average = total/count
+#        print("{:20s} $ {:<10d}   {:<10d} $ {:<.2f}".format(donor, total, count, average))
```
robalford: It's a best practice to remove all unused/commented code from your files. If you track your changes incrementally  in git, then you can always go back and look at a previous version

students/itsabhi/Lesson3/mailroom.py
```diff
@@ -0,0 +1,73 @@
...
+    for item in donor_summary:
+        print("{:20s} $ {:>10d}   {:>10d} $ {:>.2f}".format(item[0], item[1], item[2], item[3]))
+    return 2
+
+
+def main():
+    donors = [['Bill Gates', [100000, 60000, 200000]], ['Jeff Bezos', [20000, 40000]], ['Paul Allen', [50000]],
+              ['Mark Zuckerberg', [1000000]], ['Howard Schultz', [25000]]]
+    option = 0
+    while option != 3:
```
robalford: You can use `while True:` to initialize an infinite loop in Python and then you don't need to `option` variables, or numerical return values from the functions

itsabhister:
Thanks for the detailed feedback, Rob. I am incorporating some of these suggestions in the next Lesson4 exercise.

PR #311, opened 2018-03-18
submitting trigrams.py

wesmoy17:


students/WesleyM/Lesson4/trigrams.py
```diff
@@ -0,0 +1,42 @@
+import re
+import random
+
+def open_file(string):
+    '''Opens, reads and closes a .txt file'''
```
robalford: Python convention is to use triple double quotes `"""` for docstrings

students/WesleyM/Lesson4/trigrams.py
```diff
@@ -0,0 +1,42 @@
...
+    f = open(string)
+    secret_data = f.read()
+    f.close()
+    return secret_data
+
+def ngrams(input, n):
```
robalford: Cool idea to make it more flexible like this so you can experiment with different word group lengths and see what kind of results you get

students/WesleyM/Lesson4/trigrams.py
```diff
@@ -0,0 +1,42 @@
...
+    '''Takes text and makes n-grams'''
+    input = input.split(' ')
```
robalford: It's a good idea to avoid using built-in names like `input` and `dict` for variables. One convention in Python is to use a trailing underscore like `input_`, or use a more descriptive name like `trigram_dict`

students/WesleyM/Lesson4/trigrams.py
```diff
@@ -0,0 +1,42 @@
...
+def main():
+    dict = {}
+    text = (re.sub('[^A-Za-z]+', ' ', "{}".format(open_file('sherlock_small.txt')))).lower()
+    trigram = ngrams(text,3)
+
+    for i in range(len(trigram)):
+        if dict.get(trigram[i][0] + ' ' + trigram[i][1], False) == False:
+            dict[trigram[i][0] + ' ' + trigram[i][1]] = [trigram[i][2]]
+        else:
+            dict[trigram[i][0] + ' ' + trigram[i][1]].append(trigram[i][2])
```
robalford: Check out `dict.setdefault()` or `defaultdict` from the `collections` module of the Python standard library for a way to simplify these kinds of look ups and assignments

robalford:
@wesmoy17 very nice work here. I made a few minor notes, and you should check out Natasha's [solution](https://github.com/natasha-aleksandrova/Self_Paced-Online/blob/master/solutions/lesson4/trigrams.py) for a few ideas on how to simplify things, but this is ready to merge.

robalford:
@wesmoy17 I don't see the assignment submission on the course website, so be sure to submit it there as well so we can grade. thanks!

students/WesleyM/Lesson5/except_exercise.py
```diff
@@ -0,0 +1,55 @@
...
+
+langs = ['java', 'c', 'python']
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    next_joke = more_fun(langs[1])
+else:
+    more_fun(langs[2])
+
+last_fun()
```
robalford: You can use `finally` to call this in the try/except block regardless of whether there were any exceptions

students/WesleyM/Lesson5/mailroom3.py
```diff
@@ -0,0 +1,77 @@
...
+  print('Closing Program')
+
+choices = {1: send_ty, 2: report, 3: letters, 4: close_program}
+
+def use_input():
+  try:
+    action = int(input("Choose one of four actions: \n 1. Send a Thank You \n 2. Create a Report \n 3. Send Letter to Everyone \n 4. Quit \n>"))
+  except ValueError:
+    action = input("Choose one of four actions: \n 1. Send a Thank You \n 2. Create a Report \n 3. Send Letter to Everyone \n 4. Quit \n>")
+  else:
```
robalford: The try/except block needs to be inside of your while loop here. Otherwise, if a user enters bad input on the first attempt, they can never exit the `except` block. You can see this if you enter, for example, 'one' followed by '1'. I think this could also be simplified by using `KeyError` for the dict lookup.

PR #316, opened 2018-03-19
lesson06 - test_new_mailroom

ghassan24601:


students/ghassan/lesson06/test_new_mailroom.py
```diff
@@ -0,0 +1,25 @@
...
+import unittest
+from new_mailroom import main_menu, send_thankyou, donors
+
+
+class TestMailroom(unittest.TestCase):
+    def test_main_menu(self):
+        ''' this tests if the function is actually
+        returning the user's input '''
+        users_choice = main_menu()
+        self.assertEqual(users_choice, users_choice)
```
robalford: In this case, `user_choice` will always equal `users_choice` so this isn't actually testing anything.

students/ghassan/lesson06/test_new_mailroom.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def test_send_thankyou_new_donor(self):
```
robalford: The name of this function should give you a clue that your `send_thankyou` function could use some refactoring. Since each unit test is supposed to test only one thing, it's important that you try to have each your functions do only one thing. So you might look at factoring out the 'add donor' code into its own function that can be called by `send_thankyou`. This would make each separate function easier to test on their own, and make it easier to identify any potential bugs that might arise.

students/ghassan/lesson06/test_new_mailroom.py
```diff
@@ -0,0 +1,25 @@
...
+        ''' tests whether a new donor was actually added
+        to the donors list'''
+        # I did a list comprehention here to get
+        # a fresh copy of the dictionary keys as of
+        # a reference or a pointer to the same dictionary
+        old_donors = [donor for donor in donors.keys()]
+        send_thankyou()
```
robalford: Testing in this way would require you to enter user input via the command line during your tests which defeats the purpose of automated testing. Can you think of a way to refactor your `send_thankyou` function to make it more testable? I would suggest isolating letter generation into a separate function that returns rather than prints the letter so that it can be tested. I would suggest having your `create_report` function return a value that can be printed, making it more testable as well.

students/ghassan/lesson06/test_new_mailroom.py
```diff
@@ -0,0 +1,25 @@
...
+        self.assertGreater([
+            new_donor for new_donor in donors.keys()],
+            old_donors)
+
```
robalford: You need 
```
if __name__ == '__main__':
    unittest.main()
```
at the end of the file to run the tests from the command line

robalford:
@ghassan24601 I think you need to spend some more time on unit testing before moving on from this assignment. Your code needs to be refactored to make it more testable (see my notes in the comments). Try to return values from your functions that can be tested, and aim to write functions that do only one thing, making them more testable, readable and easier to debug. 

I couldn't find too many opportunities for writing unit tests with the code written as is, but here's an example for you to reference of a test for your `save_report()` function (need to import the function as well as the `os` module for this to work:
```
    def test_save_report(self):
        save_report()

        self.assertTrue(os.path.isfile('Jessie.txt'))
        self.assertTrue(os.path.isfile('Ford.txt'))
        self.assertTrue(os.path.isfile('James.txt'))
        # check that the file contains content
        with open('Jessie.txt') as f:
            size = len(f.read())
        self.assertGreater(size, 0)
```
We are going to grade this as incomplete for now, but please resubmit the assignment when it is finished. You may also consider using [pytest](https://docs.pytest.org/en/latest/) rather than `unittest` , as it allows you to write your tests as functions rather than classes to keep things simpler and less verbose. This is the tool that is used in the video lecture on 'Test Driven Development' from the course materials.

ghassan24601:
Thanks for the feedback and the example. I was also having a hard time finding test cases. 
I will try to do refactoring and introduce more test cases.

natasha-aleksandrova:
going to merge this in for now and we can review again when you're ready.

PR #321, opened 2018-03-20
assignment7

kuhnbt7:


students/kuhnbt/lesson7/html_render.py
```diff
@@ -0,0 +1,174 @@
...
+        else:
+            self.content = []
+        self.text = None
+        if keywords and type(keywords) != dict:
+            raise TypeError('Keywords must be in dictionary format')
+        self.keywords = keywords
+
+    def append(self, new_content):
+        """Add content (str or Element) to an element"""
+        if type(new_content) == str or isinstance(new_content, Element):
```
natasha-aleksandrova: you should be able to do: `if isinstance(new_content, (str, Element))`

students/kuhnbt/lesson7/html_render.py
```diff
@@ -0,0 +1,174 @@
...
+assignment about submitting attributes to the constructor for an Element
+ class. I understand the point about how passing a reserved word like class
+would cause an error, but the assigment description seemed to suggest 
+using **kwargs to collect user-defined attributes, which I don't think
+would solve the problem because a user could still pass a reserved word. 
+Then in the run_html_render script the code passes strings of attributes 
+which doesn't match what was talked about in the assignment description.
+Anyway, I decided to require the user to pass a dict of strings which 
+seemed to me like the simplest solution. Let me know if I totally 
+missed the point!"""
```
natasha-aleksandrova: (Please let me know if I misunderstood your question)
in step 4
```
Extend the Element class to accept a set of attributes as keywords to the constructor, e.g. run_html_render.py

Element("some text content", id="TheList", style="line-height:200%")
```
it is talking about adding html style attributes as keyword arguments -> 
`def __init__(self, content=None, **kwargs)`
kwargs is essentially a dictionary anyway so what you have will work just need to change the signature

PR #322, opened 2018-03-20
Submitting series.py

githubtater:
I was clicking through past assignments and I hadn't submitted this yet.

### DIFF
--- a/students/Roy_Tate/lesson03/mailroom.py
+++ b/students/Roy_Tate/lesson03/mailroom.py
@@ -1,0 +1,109 @@
+# author/student: Roy Tate (githubtater)
+
+def input_donation_amount():
+    # helper function. Use to receive contribution amounts.
+    donation_amount = input('Enter the donation amount: ')
+    return float(donation_amount)
+
+
+def send_thank_you(donor_list):
+    # Obtain the name of the donor, the contribution amount, add the donation to the list, and generate
+    # a thank you email.
+    while True:
+        name = input("Enter the full name of the recipient (Type 'list' to view donors. Type '1' for Main Menu):\n")
+        if name.lower() == 'list':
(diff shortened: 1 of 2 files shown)


githubtater:
combining pull for lesson02 and lesson03 assignments.

students/Roy_Tate/lesson03/mailroom.py
```diff
@@ -0,0 +1,110 @@
...
+        initial_menu()
+
+    elif name in donor_names():
+        donation_amount = float(input('EXISTING: Enter the donation amount: '))
+        add_donation(name, donation_amount)
+        print(email_str.format(name.title(), float(donation_amount)))
+    else:  # name not found in list
+        donation_amount = input('NEW: Enter the donation amount: ')
+        donor_list.append([name, float(donation_amount)])
+        print(email_str.format(name.title(), float(donation_amount)))
```
natasha-aleksandrova: lines 55 and 59 are identical so you can move it out to line 60

students/Roy_Tate/lesson03/mailroom.py
```diff
@@ -0,0 +1,110 @@
...
+def print_list():
+    print('{:*^50}'.format('DONOR LIST'))
+    for donor in donor_names():
+        print(donor)
+
+
+def add_donation(donor_name, new_donation):
+    for index, donor in enumerate(donor_list):
+        if donor[0] == donor_name:
+            donor_list[index].append(new_donation)
```
natasha-aleksandrova: since you get a reference to the donor item in the list, you don't need to append by index, basically following will work as well:
```
for donor in donor_list:
     if donor[0] == donor_name:
        donor.append(new_donation)
```
I would also call this function `add_new_donor` or combine to have functionality to add new donor and to add to existing donor

PR #324, opened 2018-03-20
Lesson5 - Exceptions Exercise / Lesson4 - Trigrams fixes

csdotson:
Whoops... I must've messed up my feature branching a bit here. The intent was for these to be on different branches. Sorry for grouping these together. What's here:

1) Completed Lesson5 Exception Exercise - included both exercise and test files. Let me know if you have any comments/suggestions

2) Update of Lesson4 Trigrams - Just wanted to incorporate a few of the suggested fixes from Natasha. Maybe worth a quick glance if you have a moment.


natasha-aleksandrova:
that's awesome that you were trying to use feature branching 🙂 

let us know if you need some help or have questions on feature branching, it's pretty much a must in the industry.

students/csdotson/lesson05/mailroom_3.py
```diff
@@ -0,0 +1,115 @@
...
+def send_thank_you():
+    # Create sub_prompt for creating thank you's and adding new donors/donations
+    prompt(thank_you_prompt, thank_you_menu)
+
+
+def add_donation():
+    # Append donation to existing donor, or add donor/donation to existing list
+    try:
+        donor_name = input("\nPlease enter a donor name: ")
+        new_donation = input(f"Please enter a donation amount for {donor_name}: ")
```
robalford: I would suggest doing the float conversion here, so it's only happening once in the try block. You generally want to include only the code that might raise the exception in the try block if you can, and catch the exception as closely to that point as you can, so I'd move the dict look up and additions out since there's nothing in those operations that would raise a `ValueError`.

PR #325, opened 2018-03-20
Lesson5 - Mailroom pt3

csdotson:
Adding comprehensions and exception handling of user input. 

**Notes:**

- I only found one place where I felt a comprehension made sense. There are other loops where I could've used a comprehension, but I thought readability might be compromised. Let me know if there were other places where I should've considered a comprehension.


csdotson:
Btw, I attempted to use feature branching here. Let me know if it looks like I did it correctly - not sure if it looks any different on your end whether I'm opening a PR from a feature branch vs a PR from master.

natasha-aleksandrova:
screen shot 2018-03-20 at 8 22 57 am

natasha-aleksandrova:
I think your branching looks right! Looks no different to us to really, we just different source for the branch (your custome name instead of `master`)

natasha-aleksandrova:
accident merge

students/csdotson/lesson05/mailroom_3.py
```diff
@@ -0,0 +1,115 @@
...
+        avg_gift = total_given / num_gifts
+        print('{:21}{:>15.2f}{:>16}{:>16.2f}'.format(donor, total_given, num_gifts, avg_gift))
+    print()
+
+
+def list_donors():
+    # Formatted print of donor names
+    print("\nList of donors:")
+    [print(donor) for donor in donations]
+    print()
```
natasha-aleksandrova: (see my post in slack on this)

this is better as a for loop, or have your function return a string that can be printed (which can be utilized once u get to unit testing):
```
def list_donors():
    return "\n".join(list(donations))
```

students/csdotson/lesson05/mailroom_3.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python3
+# Lesson 4 - Mailroom... part 2
+
+### Function Definitions ###
+def prompt(prompt, menu):
+    # Create menu based on arguments
+    while True:
+        try:
+            response = input(prompt)
```
natasha-aleksandrova: this line should technically be above try

PR #326, opened 2018-03-20
Lesson 4 Mailroom2 exercise

itsabhister:


students/itsabhi/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,69 @@
+#! /usr/bin/env python3
+
+def sending_thanksv2(donor_hist):
+
+    donor = input("Type the Full Name of the donor here or type List to display current donors : ")
```
robalford: I would suggest placing your user input prompts in while loops so you don't have to repeat them in your conditional code blocks below

students/itsabhi/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,69 @@
...
+    if donor == "list":
+                for k in donor_hist:
```
robalford: This block looks over-indented

students/itsabhi/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,69 @@
...
+                    print(k)
+                donor = input("Type the Full Name of the donor here: ")
+    if donor in donor_hist.keys():
```
robalford: The `in` operator checks for existence of keys in dict so no need to call `keys()` here

students/itsabhi/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,69 @@
...
+        amount = input("Enter Donation value: ")
+        donation_hist = donor_hist.get(donor)
+        donor_hist[donor] = (donation_hist + [int(amount)])
```
robalford: I would use `append()` here since `amount` is a single value

students/itsabhi/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,69 @@
...
+    return True
+
+
+def quitting(donors):
+    print("Quitting...")
+    return False
+
+def main():
+    donors_dict = {'Bill Gates': [100000, 60000, 200000], 'Jeff Bezos': [20000, 40000], 'Paul Allen': [5000], 'Mark Zuckerberg': [1000000], 'Howard Schultz': [25000]}
+
```
robalford: This is a nice format for dictionaries, to keep them under Pep8 line character length and keep each entry on its own line:
```
    donors_dict = {  
        'Bill Gates': [100000, 60000, 200000],
        'Jeff Bezos': [20000, 40000],
        'Paul Allen': [5000],
        'Mark Zuckerberg': [1000000],
        'Howard Schultz': [25000]
    }
```

PR #328, opened 2018-03-20
Lesson03 list_lab.py Ready for Review

idcrickmore:
Please review list_lab.py

Thank you,

Ian David


natasha-aleksandrova:
@idcrickmore I also see string format lab here as well but not submitting on the course website, is it not ready for review?

students/idcrickmore/lesson03/list_lab.py
```diff
@@ -1,37 +1,138 @@
...
-    fruit = ["Apples", "Pears", "Oranges", "Peaches"]
+    print("Number {} is {}".format(num_response, fruit[num_response-1]))
+
     # add two more fruits
-    two_more_fruit = ["Persimmon", "Avocado"]
-    
+    fruit_response2 = [input
+                       ("What's a second kind of fruit you'd like to add? > ")]
+    fruit_response3 = [input
+                       ("What's a third kind of fruit you'd like to add? > ")]
```
natasha-aleksandrova: if you're not going to need fruit_response3 as a list then don't create list so that later it is simply `fruit.insert(0, fruit_response3)`

students/idcrickmore/lesson03/list_lab.py
```diff
@@ -1,37 +1,138 @@
...
+    print("start with P")
+
+
+def series_2():
+    print("Task 2")
+    # This function references the global 'fruit' variable
+    # But only modifies a local version, 'fruit_2'
+    global fruit
+    fruit_2 = fruit[:]
+    print(fruit_2)
```
natasha-aleksandrova: you shouldn't need to use `global` here

students/idcrickmore/lesson03/list_lab.py
```diff
@@ -1,37 +1,138 @@
...
+
+    # remove the last item from the fruit list
+    fruit_2 = fruit_2[0:-1]
```
natasha-aleksandrova: since 0 is default it can just be `fruit_2[:-1]`

students/idcrickmore/lesson03/list_lab.py
```diff
@@ -1,37 +1,138 @@
...
+        # ... and remove it from the list
+        for x in fruit_2:
+            if x == delete_response:
+                fruit_2.remove(delete_response)
+                fruit_deleted = True
+        # if user input is bunk ask again
+        if fruit_deleted is False:
+            print("{} is not a fruit in the list. Try again."
+                  .format(delete_response))
+            print(fruit_2)
```
natasha-aleksandrova: could you utilize `break` and `else` clause here instead of keeping track of a bool var?

natasha-aleksandrova:
ah! you would need to use feature branching (advanced concept) or have 1 PR for both assignments.

go ahead and submit the same PR url in the str format lab on the website.

idcrickmore:
Done. Thank you!

natasha-aleksandrova:
@idcrickmore now I see mailroom here 🙂 

is that something you want reviewed as well?

natasha-aleksandrova:
note: str format is good to go

students/idcrickmore/lesson03/list_lab.py
idcrickmore: These are great comments, thank you. 

Just so I understand the grading/workflow process, should I make these changes to the script and make another push to github?

IDC

natasha-aleksandrova: great question. grading wise, you will get a full grade for completing the assignment. 
when we leave suggestions/recommendations it is for your learning benefit, and yes you could do updates and commit into an open PR or create a new one (depending if one is still open or not).

For this particular PR - I would like to merge it as soon as possible as it is getting too big to manage. Ideally PRs should be small containing single feature.

idcrickmore:
Ok, understood. I have uploaded and pushed all files for lesson 3 (slicing,
list, string and mailroom). Please go ahead and review them as completed
work.

Thank you,

IDC

natasha-aleksandrova:
sounds good 👍

students/idcrickmore/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+            report()
+        elif choice[0] == '3' or choice[0] == 'q':
+            print("quiting")
+            quit()
+        else:
+            # if the user input does not match one of the 3 options,
+            # call the menu function again
+            menu()
+
+    options()
```
natasha-aleksandrova: this would be a pretty good candidate for `while True` loop

students/idcrickmore/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def thankyou():
+    global donors
```
natasha-aleksandrova: you shouldn't need to use globals here

students/idcrickmore/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+    print("\n------------------ THANK YOU ------------------\n")
+    print("- type 'list' to to see a complete list of donors -")
+    print("- type 'menu' at any time to return to the menu -\n")
+
+    # simple function to list all the donors
+    # calls the 'thankyou' function at the end to ask for user input
+    def list():
```
natasha-aleksandrova: be careful about overwriting built-in objects, here you just overwrote `list` (notice github highlighted it different color)

students/idcrickmore/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+        menu()
+    if name_check == 'list':
+        list()
+    else:
+        name_loc = 0
+
+        for name in donors:
+            if name_check == name[0].lower():
+                break
+            name_loc += 1
```
natasha-aleksandrova: you could use `for`/`else` instead of counter variable and append in the for loop once you find the existing donor

students/idcrickmore/lesson03/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+    for name in donors:
+        sum_don = sum(donors[name_loc][1:])
+        num_don = len(donors[name_loc])-1
+        ave_don = sum_don / num_don
+        donors_report += [[donors[name_loc][0], sum_don, num_don, ave_don]]
+        name_loc += 1
+
+    # sorts the donors report by the second column "donation[1] using 'lambda'
+    # and reverses the order from largest to smallest
+    donors_report.sort(key=lambda donation: donation[1], reverse=True)
```
natasha-aleksandrova: check out `itemgetter` for using with sort/sorted functions (`from operator import itemgetter`)

idcrickmore:
Natasha and/or Rob,

I've run into a problem with part 7 of lesson 7. I've created new classes
for Ul and Li (subclasses of Element)  and a new class for H (subclass of
OneLineTag). When I run 'run_html_render.py' an html file is successfully
created, but the final Li instance is broken into three (3) lines when it
should just be one.



I think the problem is somewhere in the render() function of the Element
Class but I'm not sure. Do you have any suggestions or ideas why this is
happening and how I can fix it?

my lesson07 folder on github can be found here:
https://github.com/idcrickmore/Self_Paced-Online/tree/master/students/idcrickmore/lesson07

Thank you

Ian David

idcrickmore:
In the following screenshot from the lesson 8 lecture on Properties, how
does the program know to use the 'get_health' function at line 4, and then
know to use the 'set_health' function on line 5? From what I can tell, both
functions and they are accessed by calling the exact same
'property_fighter_guy.health' property, without distinguishing between
which function is being called.



Thanks,

IDC

PR #329, opened 2018-03-20
Activity 3

ChrisKenyon89:


students/Chris_Kenyon/Lesson_3/list_lab.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+# Series 3
+print("---------------------------------")
+series_3 = ['Apples', 'Pears', 'Oranges', 'Peaches']
+
+# While loop to check for proper input and
+# remove fruit from list based on user tastes
+gross_fruit = []
+for f in series_3:
```
robalford: If you loop through a copy like this: `for f in series_3[:]`, you can remove fruits from the original list inside your `for` loop without needing to build and loop through the `gross_fruit` list

students/Chris_Kenyon/Lesson_3/Mailroom_Pt1.py
```diff
@@ -0,0 +1,144 @@
...
+            continue
+        return option
+
+
+def send_thanks():
+    """ Send Thanks """
+    page_break()
+    while True:
+        list_names = [item[0] for item in donor_chart]
+        amount = []
```
robalford: Doesn't look like you need to assign this here, since you're reassigning to the input value below

students/Chris_Kenyon/Lesson_3/Mailroom_Pt1.py
```diff
@@ -0,0 +1,144 @@
...
+        try:
+            print("To whom would you like to say thank you?\n"
+                  "(type \"list\" for a full list of names or"
+                  "\"exit\" to return to the menu)")
+            name = input("--->")
+        except ValueError:
```
robalford: I don't think `ValueError` is the right exception to catch here, as `input` will happily convert values like '1.5' into strings.

students/Chris_Kenyon/Lesson_3/Mailroom_Pt1.py
```diff
@@ -0,0 +1,144 @@
...
+                donor_chart[list_names.index(name)][1].append(amount)
+        elif name.lower() == 'exit':
+            break
+        else:
+            addname = input("The name you selected is not in the list,"
+                            "would you like to add it(y/n)? ")
+            if addname[0].lower() == 'y':
+                amount = getAmount()
+                print(amount)
+                if amount == 'exit':
```
robalford: Could this check for 'exit' be moved into the `getAmount()` function, to eliminate repetition and nested `if` statement here?

students/Chris_Kenyon/Lesson_3/Mailroom_Pt1.py
```diff
@@ -0,0 +1,144 @@
...
+                    donor_chart.append([name, [amount]])
+            elif addname.lower() == 'exit':
+                break
+            else:
+                print("\nName was not added, try again\n")
+                continue
+        print("\nDear {} \nThank you for your generous donation of ${:.2f}!!\n"
+              "Now all of the kittens will get "
+              "to eat this year".format(name, amount))
+        break
```
robalford: There are a lot of nested conditional logic and loops in your `send_thanks()` function which makes it a bit difficult to read and debug. Can you think of a way to refactor to make it 'flatter'?

students/Chris_Kenyon/Lesson_3/Mailroom_Pt1.py
```diff
@@ -0,0 +1,144 @@
...
+    return
+
+
+def getAmount():
```
robalford: Python convention is to use `snake_case()` for variable and function names

students/Chris_Kenyon/Lesson_3/Mailroom_Pt1.py
```diff
@@ -0,0 +1,144 @@
...
+    while True:
+        try:
+            amount = input("How much did they donate: ")
+            if str(amount).lower() == 'exit':
+                return amount.lower()
+            else:
+                return float(amount)
+        except:
```
robalford: It's always better to narrow in on a specific exception, rather than catching all exceptions with a bare `except` statement, because sometimes there can be errors that you don't anticipate and you can don't want those to pass silently. This would be a good place to catch a `ValueError`, as that will be raised by a failed `float()` conversion for non-numeric types.

students/Chris_Kenyon/Lesson_3/Mailroom_Pt1.py
ChrisKenyon89: I'm not familiar with many Error catches which would you suggest?

students/Chris_Kenyon/Lesson_3/Mailroom_Pt1.py
ChrisKenyon89: is there a way to return a break when assigning a variable with a function
like 
While True
    amount = get_amount()
what can get_amount() return to break the while loop

PR #331, opened 2018-03-21
trigram

ZhuojingXie:


students/ZhuojingXie/lesson04/trigram.py
```diff
@@ -0,0 +1,54 @@
...
+            #word i can add in.
+            break
+
+    print( ' '.join(new_text_list))
+
+
+def main():
+    a = read_file()
+    b = my_key(a)
+    new_text(b)
```
natasha-aleksandrova: I would generally avoid using one letter meaningless vars as it is unclear what is happening when reading the code

PR #332, opened 2018-03-21
submt lesson6 files

tricaonguyen:
Natasha,
I made changes per your advice in #287  to mailroom.py and used dict for values as well.

I did:
+    donor_name = enumerate([donor for donor in donor_dict])
+    for idx, name in donor_name:
+        print(idx + 1, '-', name)
because I wanted to use the index number to list donor names.  For example:
1 - William Gates, III
2 - Mark Zuckerberg
3 - Jeff Bezos
4 - Paul Allen
5 - Tri Nguyen

Is there a better way to do it without enumerating?

With regard to test_mailroom.py, I am not sure how to test functions that print. Do I use mock? Could you provide an example of how to test my create_report() function using mock?

Looking forward to your advice and tips.

Thanks,

Tri

### DIFF
--- a/students/tri_nguyen/lesson06/mailroom.py
+++ b/students/tri_nguyen/lesson06/mailroom.py
@@ -1,0 +1,225 @@
+#!/usr/bin/env python3
+
+# ------------------------------------------- #
+# Student:   Tri Nguyen
+# Instructor: Natasha Aleksandrova
+# Date:  16-Mar-2018
+# ------------------------------------------- #
+
+# Import stuff
+from operator import itemgetter
+import time
+
+# Get local time
+local_time = time.localtime()
+
+# Declare variables and data structure
+
+donor_dict = {
+
+    'William Gates, III': {'total': 653784.49, 'num_gifts': 2, 'avg': 326892.24},
+    'Mark Zuckerberg': {'total': 16396.10, 'num_gifts': 3, 'avg': 5465.37},
+    'Jeff Bezos': {'total': 877.33, 'num_gifts': 1, 'avg': 877.33},
+    'Paul Allen':{'total': 708.42, 'num_gifts': 3, 'avg': 236.14},
+    'Tri Nguyen': {'total': 100.00, 'num_gifts': 1, 'avg': 100.00}
+    }
+
+
+# Declare functions
+
+
+def display_menu():
+    ''' display a menu of 3 actions '''
+
+    menu = ''' Select one of the actions below:
+
+    1. Send a Thank You
+    2. Create a Report
+    3. Send letters to everyone
+    4. Quit
+
+    '''
+    return menu
+
+
+def display_donor_list():
+    ''' display list of donors
+        from donor_dict '''
+
+    # enumerate the list to get the index number
+    donor_name = enumerate([donor for donor in donor_dict])
+    for idx, name in donor_name:
+        print(idx + 1, '-', name)
+
+
+def convert_to_float(num):
+    ''' convert numeric values from string to float '''
+
+    try:
+        num = float(num)
+    except ValueError:
+        print('value entered is not a number!!')
+    else:
+        return num
+
+
+def generate_email_template():
+    ''' send an email to one donor that has just donated
+        this function works with send_thanks() function '''
+
+    email_template = '''
+                Dear {0},
+
+                Thank you for giving us ${1:,.2f}. Your money will be put to good use.
+                Please come back and donate more.
+
+                Best Regards,
+
+                John Doe
+            '''
+
+    return email_template
+
+
+def send_thanks():
+    ''' send thanks to a donor for their generous donation '''
+
+    while True:
+
+        full_name = input('Enter a Full Name (or type "list" to get the list of donors): ').title()
+
+        if full_name.lower() == 'list':
+            display_donor_list()
+            continue
+
+        elif donor_dict.get(full_name):
+            donation_amount = convert_to_float(input('Enter donation amount: '))
+            # Update values in list
+            donor_dict[full_name]['total'] += donation_amount
+            donor_dict[full_name]['num_gifts'] += 1
+            donor_dict[full_name]['avg'] = donor_dict[full_name]['total'] / donor_dict[full_name]['num_gifts']
+
+        else:
+            donation_amount = convert_to_float(input('Enter donation amount: '))
+            donor_dict[full_name] = dict(total=donation_amount, num_gifts=1, avg=donation_amount)
+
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
Hi there,

once you get all your keys out of dictionary (donor names) then you treat it just like a normal list, with a normal list you really have two options: using `enumerate` or counter variable. `enumerate` is a good choice here:
```
for idx, name in enumerate(list(donor_dict)):

```
note that you don't need a for loop to iterate over keys, by default dicts iterate over keys.
-----
as far as unit testing, `mock` module is a good choice in general for cases like that... but I wouldn't   use here to assert print statements. Consider the report function that prints the report, what you can do is change it to return the report string instead of doing the print in there:
```
def get_report():
    report_str = ""
    # some stuff here to create the str
    return str
```
now your unittest can assert that correct report string is generated and not worry about print statements.

students/tri_nguyen/lesson06/mailroom.py
```diff
@@ -0,0 +1,225 @@
...
+def print_header():
+    ''' print the header of donor report '''
+
+    header = ['Donor Name', 'Total Given', 'Num Gifts', 'Average Gift']
+    template = '{0:<26}|{1:^13}|{2:^11}|{3:^14}'
+
+    formatted_columns = template.format(*header)
+    border = '{}'.format('-' * len(formatted_columns))
+
+    return (formatted_columns, border)
```
robalford: Don't need the parentheses here

students/tri_nguyen/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+    1. Send a Thank You
+    2. Create a Report
+    3. Send letters to everyone
+    4. Quit
+
+    '''
+
+        expected_result = mailroom.display_menu()
+        self.assertEqual(expected_result, known_menu)
+        # are assertIsNotNone and assertIsInstance necessary??
```
robalford: It's good to be thinking this way when writing tests, and checking for existence and type are useful things to test for in some cases, however since the function you're testing here is only assigning a string to a variable, I don't think you need these assertions

students/tri_nguyen/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+    def test_convert_to_float(self):
+        ''' test convert_to_float function of mailroom.py '''
+
+        self.assertEqual(mailroom.convert_to_float('100.00'), 100.0)
+        self.assertEqual(mailroom.convert_to_float('12.234'), 12.234)
+        self.assertEqual(mailroom.convert_to_float('145.12'), 145.12)
+        self.assertEqual(mailroom.convert_to_float('14.123456'), 14.123456)
+
```
robalford: I would suggest testing for bad input here as well

students/tri_nguyen/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+                Best Regards,
+
+                John Doe
+            '''
+
+        expected_result = mailroom.generate_email_template()
+        self.assertEqual(expected_result, known_temp)
+        # are assertIsNotNone and assertIsInstance necessary??
+        self.assertIsNotNone(expected_result)
```
robalford: Same as above, this is just testing the string assignment which is bordering on testing Python's built-in functionality. But you could make your `generate_email_template` function more testable by including `full_name and `donation_amount` parameters and then testing the formatted string output with different input values.

students/tri_nguyen/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+    def test_format_filename(self):
+        ''' test format_filename function of mailroom.py '''
+
+        self.assertEqual(mailroom.format_filename('tri nguyen'), 'tri_nguyen.txt')
+        self.assertEqual(mailroom.format_filename('william gates, iii'), 'william_gates_iii.txt')
+        self.assertEqual(mailroom.format_filename('one'), 'one.txt')
+
+        with self.assertRaises(TypeError):
+            mailroom.format_filename(82)
+
```
robalford: Nice! That's what I mean by testing for bad input. This test looks great.

students/tri_nguyen/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+        ''' test print_header function of mailroom.py '''
+
+        known_header = ['Donor Name', 'Total Given', 'Num Gifts', 'Average Gift']
+        known_template = '{0:<26}|{1:^13}|{2:^11}|{3:^14}'
+        formatted_columns = known_template.format(*known_header)
+        border = '{}'.format('-' * len(formatted_columns))
+        
+        self.assertEqual(mailroom.print_header()[0], formatted_columns)
+        self.assertEqual(mailroom.print_header()[1], border)
+
```
robalford: Can you think of a way to test your calculated values from the `donor_dict` end that end up in the report here as well?

students/tri_nguyen/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+        known_dict = {
+            'William Gates, III': {'total': 653784.49, 'num_gifts': 2, 'avg': 326892.24},
+            'Mark Zuckerberg': {'total': 16396.10, 'num_gifts': 3, 'avg': 5465.37},
+            'Jeff Bezos': {'total': 877.33, 'num_gifts': 1, 'avg': 877.33},
+            'Paul Allen':{'total': 708.42, 'num_gifts': 3, 'avg': 236.14},
+            'Tri Nguyen': {'total': 100.00, 'num_gifts': 1, 'avg': 100.00}}
+
+        known_list = [(name, known_dict[name]['total'], known_dict[name]['num_gifts'], known_dict[name]['avg']) for name in known_dict]
+
```
robalford: This is tricky to test well, but because this basically repeats the code verbatim from, your `generate_donor_list` function, it isn't really testing anything. I think the best you could do here would be something simple like:
```
self.assertIn('Jeff Bezos', mailroom.generate_donor_list())
```

robalford:
@tricaonguyen you're off to a great start here with your unit tests for mailroom, and you've done a nice job writing functions that return values. I do see some areas for improvement and have made some notes in the PR for your review. I would also suggest adding another test for `send_all_letters()` that tests that the files get created on the file system with the correct names. I'd encourage you to add the revisions from my comments to the next version/assignment, but this PR is ready to merge.

PR #334, opened 2018-03-21
Lesson10

JerHsi:
@natasha-aleksandrova I think I am not understanding the #3 requirement for this assignment.  Could you explain a little more what it means by "double contributions under $100"?  Also created a new branch for this.  Thanks


natasha-aleksandrova:
so you would run projections on your current DB and in that example would produce projections report if each donation under 100 was doubled (multiplier of 2)

natasha-aleksandrova:
closing this for now

PR #337, opened 2018-03-22
added testing program

alexLaws:
I pulled together a number of tests.  Several of the programs rely on user input to run correctly but I think I tested the main functions.


students/alexLaws/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+
+import mailroom
+
+
+def test_1():
+    '''test the note writing function with a current donation'''
```
robalford: It's a best practice/Python convention to give your unit tests specific, detailed names like `test_generate_note_current_donation()`

PR #338, opened 2018-03-22
[DennisLee][L6][Mailroom P4] Ready for submission.

DennisLee-DennisLee:


students/DennisLee/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,393 @@
...
+        '': {'function': m.exit_screen},
+        'quit': {'function': m.exit_screen},
+        'list': {'function': m.print_donor_list}
+}
+
+
+def test_manage_donors_1():
+    # An I/O error occurs because the function prompts for menu choice
+    with pytest.raises(IOError):
+        m.manage_donors()
```
robalford: Rather than catching these `IOError`s like this for your prompts, I would encourage you to look into the concept of mocking user input data for your tests. Python's `unittest` testing framework uses classes and has a little steeper learning curve than `pytest` but you seem to have a good handle on the basics of unit testing and I think you'd find `unittest` very useful. The `unittest.mock` module allows you to essentially fake user input data, third party API calls etc. in your tests so that you can test the parts of your program that depend on these types of external inputs and resources.

students/DennisLee/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,393 @@
...
+
+# Simulate and test function calls that occur in the code 
+# after the input() function that causes the I/O error
+# There aren't menu options for 0', '5', 'a', & '', so they return False
+# '1' and '3' lead to user prompts, so catch those
+# '2' and '4' go to functions w/o user prompts, so they return True
+def test_call_menu_function_1():
+    assert m.call_menu_function(
+            choices, '0', m.respond_to_bad_main_menu_choice) == False
```
robalford: Generally in Python/Pep8, you want to use the `is` operator for comparison to `False`, or in most cases just do `assert not function_that_returns_a_boolean()`.

students/DennisLee/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+
+def test_respond_to_bad_main_menu_choice():
+    assert m.respond_to_bad_main_menu_choice('any') == None
+
+
+def test_exit_screen_1():
+    assert m.exit_screen() == None
+
```
robalford: Use the `is` operator for `None` comparisons as well for Python/Pep8 convention

students/DennisLee/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,393 @@
...
+    donation_choices.pop('list')
+    assert m.call_menu_function(donation_choices, 'fdasfd', m.add_donation,
+            donor_name='C. N. Enome') == False
+
+
+# For the `add_donation` call, only positive numbers are added
+# successfully to the donor history list and are returned back
+def test_add_donation_1():
+    assert m.add_donation(50, 'C. N. Enome') == 50
+    del m.donor_history['C. N. Enome']  # restore dict for next tests
```
robalford: The `unittest` framework includes a `setUp()` method and some other helper methods for shared test data to help you out with these kinds of situations

students/DennisLee/lesson06/test_mailroom.py
DennisLee-DennisLee: Thanks. I had tried reassigning the input name to my own input-mocking function, but that didn't seem to work during testing. Also, the course content is a little confusing with regard to what testing tool to use; it explicitly mentions `pytest`, but then some of the required reading material showed how to use `unittest`. Since `unittest` involved classes, which haven't been covered extensively in the coursework yet, I assumed we were supposed to stick with `pytest`.

students/DennisLee/lesson06/test_mailroom.py
DennisLee-DennisLee: If a function can return `None`, wouldn't `not None` evaluate to `True`? That's why I haven't been using the simpler expression.

students/DennisLee/lesson06/test_mailroom.py
robalford: @DennisLee-DennisLee I totally agree and was already planning on bringing this to the attention of the course authors. We've had students use both `pytest` and `unittest` for the mailroom assignment and it's fine to go with whichever test runner/framework you prefer. Some people like `pytest` for its simplicity and more Pythonic style, but I feel like for both mocks and set up data, you might find the `unittest` API quite useful.

students/DennisLee/lesson06/test_mailroom.py
robalford: There are definitely some cases where it makes sense to go against the conventions and you'll see boolean comparisons that use the `==` operator in plenty of Python code. I just wanted to mention this, as a general Python style consideration. 

Here's a stack overflow post that speaks to the topic and links to the relevant Pep8 documentation if you want to learn more: https://softwareengineering.stackexchange.com/a/322394

PR #341, opened 2018-03-23
Mailroom refactored for readability and testing

MikeS-1:
Mailroom_test included for assertions and use with pytest


students/MikeShand/lesson06/mailroom4_test.py
```diff
@@ -0,0 +1,31 @@
...
+    assert "Andy" in mail_list
+    assert "Bill" in mail_list
+    assert "Chuck" in mail_list
+
+
+def test_donor_report():
+    report = mailroom.donor_report()
+    assert 'Andy' in report
+    assert 'Bill' in report
+    assert 'Chuck' in report
```
natasha-aleksandrova: to be a  more useful test, I would assert full report string which would test that report renders as expected vs just verifying that those names were included.

students/MikeShand/lesson06/mailroom4_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def test_batch_file():
+    mailroom.batch_file()
+    assert os.path.isfile("Andy.txt")
+    assert os.path.isfile("Bill.txt")
+    assert os.path.isfile("Chuck.txt")
```
natasha-aleksandrova: another thing that could be tested is the content of one of the files

natasha-aleksandrova:
note: needs to be graded on edx

PR #342, opened 2018-03-23
Herela: Lesson 7: HTML rendering

harvey-herela:


students/HarveyHerela/lesson07/html_render.py
```diff
@@ -0,0 +1,128 @@
+class Element:
+    """Base class for all HTML tags."""
+    def __init__(self, tag_name='element', content=None, **kwargs):
+        if content is None:
```
robalford: I would make `tag_name` a class attribute, rather than instance attribute since it should be the same value for all instances of your `Element` class

students/HarveyHerela/lesson07/html_render.py
```diff
@@ -0,0 +1,128 @@
...
+        self.tag_open = "<{tag}{tag_style}>".format(tag=tag_name, tag_style=tag_style)
+        self.tag_close = "</{tag}>".format(tag=tag_name)
+
+
+    def append(self, content):
+        self.content.append(content)
+
+
+    def render(self, file_out, cur_ind = "", one_line_tag=False):
+        file_out.write(cur_ind + self.tag_open)
```
robalford: I would include the logic for `OneLineTag` in the `render` method of that subclass, so you don't need the boolean parameter here in the base class

students/HarveyHerela/lesson07/html_render.py
```diff
@@ -0,0 +1,128 @@
...
+        if not one_line_tag:
+            file_out.write('\n')
+
+        for content in self.content:
+            if type(content) == str:
+                file_out.write(content)
```
robalford: See the section 'Notes on Ducking Typing' in the assignment instructions for a couple strategies for handling both xml and plain text and apply the same indentation logic to each. It looks like your plaint text content is not currently indented to the same level as the xml attributes

students/HarveyHerela/lesson07/html_render.py
```diff
@@ -0,0 +1,128 @@
...
+
+        if not one_line_tag:
+            file_out.write('\n' + cur_ind)
+        file_out.write(self.tag_close)
+
+
+class Html(Element):
+    def __init__(self, content=None, **kwargs):
+        Element.__init__(self, tag_name="html", content=content, **kwargs)
+
```
robalford: See my note above on making `tag_name` a class attribute, shared between all instances of these classes -- then you won't need to call `__init__()` on each `Element` subclass,  you can just override the `tag_name` attribute

students/HarveyHerela/lesson07/html_render.py
```diff
@@ -0,0 +1,128 @@
...
+        self.tag_open = self.tag_open[:-1] + "/>"
+
+
+    def append(self, content):
+        raise TypeError
+
+
+    def render(self, file_out, cur_ind = ""):
+        file_out.write(cur_ind + self.tag_open)
+
```
robalford: Although Python will let you get away with it, you want to be very careful about changing the call signature when overriding methods in a subclass. This could be avoided here by moving the `OneLineTag` render logic into that subclass's `render` method as mentioned above and doing away with the `one_line_tag` parameter

students/HarveyHerela/lesson07/html_render_test.py
```diff
@@ -0,0 +1,99 @@
...
+
+import html_render
+
+class HtmlRenderTest(unittest.TestCase):
+
+    def test_basic_tags(self):
+        tags = ('Body', 'Head', 'Ul', 'Li')
+        for t in tags:
+            expected = '<{tag}>\n\n</{tag}>'.format(tag=t.lower())
+            page = eval('html_render.{tag}()'.format(tag=t))
```
robalford: Not a concern here obviously, but you generally want to avoid using `eval` in production code for reasons detailed in this post by a prominent Python dev/thought leader: 

https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html

PR #344, opened 2018-03-24
list lab completion

mmsunday1:


students/M_Sunday/lesson03/Mailroom_Part_1.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/python
+from operator import itemgetter
+
+# Lesson 03, Mailroom Part 1 Assignment
+
+
+# ------- Global Assignments -------
+donor_data = [['John', 'Marks', 5000.34, 2500, 1267],
```
robalford: I would suggest storing donations separately from names in their own sub-list to simplify your calculations and look ups. So the first item in your `donor_data` list would look something like this:
```
['John Marks', [5000.34, 2500, 1267]
```

students/M_Sunday/lesson03/Mailroom_Part_1.py
```diff
@@ -0,0 +1,102 @@
...
+    print(header)
+    print(buffer)
+    for item in sorted(data, key=itemgetter(-2), reverse=True):
+        print("{:<18} $  {:>9.2f}  {:>10}  $   {:>9.2f}".format(item[0] + ' ' + item[1],
+                                                                item[-2], item[-3],
+                                                                item[-1]))
+
+
+def compose_email(donor_name, donation_amount):
+    greetingemail = "\nDear {},".format(donor_name)
```
robalford: Python convention is to use `snake_case` for variable and function names (words separated by underscores)

students/M_Sunday/lesson03/Mailroom_Part_1.py
```diff
@@ -0,0 +1,102 @@
...
+    print(greetingemail, bodyemail)
+
+
+def add_donation(donor_name, i):
+    donation_amount = input("\nEnter the donation amount for {} ".format(donor_name))
+    donor_data[i].append(float(donation_amount))
+    compose_email(donor_name, donation_amount)
+
+
+def thank_you(data=donor_data):
```
robalford: I don't think you need the `data` parameter here, since there's only one global `donor_data` variable that you'll be working with

students/M_Sunday/lesson03/Mailroom_Part_1.py
```diff
@@ -0,0 +1,102 @@
...
+    donor_name = input("\nPlease enter the donor's full name: ")
+    if donor_name == 'list':
+        list_display(donor_data)
+    else:
+        donor_firstname = donor_name.split(' ')[0]
```
robalford: `split()` will split on white space by default so no need to pass in the empty string

students/M_Sunday/lesson03/Mailroom_Part_1.py
```diff
@@ -0,0 +1,102 @@
...
+        donor_lastname = donor_name.split(' ')[1]
+        for i, item in enumerate(data):
+            if (item[0] == donor_firstname) and (item[1] == donor_lastname):
```
robalford: You don't the need parentheses for the `if/and` clauses here

students/M_Sunday/lesson03/Mailroom_Part_1.py
```diff
@@ -0,0 +1,102 @@
...
+                add_donation(donor_name, i)
+                break
+        else:
+            donor_data.append(donor_name.split(' '))
+            add_donation(donor_name, -1)
+
+
+def calc_data(data=donor_data):
```
robalford: This function only works as intended the first time it is called. On subsequent calls, it adds your calculated values into the new calculations as though they were donations. If you select `2: Create a Report
` a few times in a row on the command line, you'll see how the donation numbers grow each time a new report is generated. There's really no need to append these calculations to your data structure, I would just append the new donations that get added in the `thank_you` function and then do your calculations on the existing donations each time you need to generate a  new report in your `list_details` function.

students/M_Sunday/lesson03/list_lab.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/python
+
+# lesson 03 - list lab exercise
+
+
+def list_contents(list_, id_=1):
```
robalford: It doesn't look like you're ever passing in a different value for the `id_` parameter, so it can be removed from this function.

students/M_Sunday/lesson03/list_lab.py
```diff
@@ -0,0 +1,92 @@
...
+    print("\nSeries List Includes the Following Fruits:")
+    for j, item_ in enumerate(list_):
+        if id_ == 1:
+            print("  - {}".format(item_))
+
```
robalford: There's no need to call `enumerate()` in this loop since you aren't referencing the index of the list in the loop, so I believe this function can be rewritten as:
```
def list_contents(list_):  
    print("\nSeries List Includes the Following Fruits:")
    for item_ in list_:
        print("  - {}".format(item_))
```

students/M_Sunday/lesson03/list_lab.py
```diff
@@ -0,0 +1,92 @@
...
+user_val = input("\nAdd Another Fruit to Series List: ")
+
+list_1.insert(0, user_val)
+
+list_contents(list_1)
+
+print("\nAll the Fruits that Begin with the Letter P Are: ")
+
+for item in list_1:
+    if item[0] == 'P':
```
robalford: You can do `if item[0].lower() == 'p':` to catch lower case user input

students/M_Sunday/lesson03/list_lab.py
```diff
@@ -0,0 +1,92 @@
...
+list_1.remove(user_val)
+
+list_contents(list_1)
+
+# Ends Series 2
+
+# Series 3
+
+list_new = list_1[:]
+for item in list_1:
```
robalford: Nice catch looping through a copy of the original list!

students/M_Sunday/lesson03/list_lab.py
```diff
@@ -0,0 +1,92 @@
...
+    user_val = input("\nDo you like {}? ".format(item))
+    while (user_val != 'No') and (user_val != 'Yes'):
```
robalford: The parentheses are not necessary here for the `while .. and` statement

PR #345, opened 2018-03-24
Lesson05: mailroom.py 

22e88:
Added exception handling to mailroom.py. Additionally i reviewed the code for possibilities to use comprehensions, but i didn't see any without adding some kind of new functionality (e.g. show last two donations of every donor or so...). 
As i think that refactoring my code quite much is necessary anyway to be able to implement decent unittesting, this (use of comprehensions) may be done additionally in one of the following exercises.
Furthermore i added a small change to already assigned and accepted except_exercise.py to reflect Rob's comment on it (thx Rob :).

### DIFF
--- a/students/elmar_m/lesson05/mailroom.py
+++ b/students/elmar_m/lesson05/mailroom.py
@@ -5,11 +5,7 @@ file: mailroom.py
 elmar_m / 22e88@mailbox.org
 Lesson05: Mailroom Exercise Part 3
 '''
-
-# import time
-# import collections
 import time, collections, sys
-
 
 main_p = '\n> Main menu. Options: t == thankyou | r == report | q == quit program\n'
 sub_p = '\n>> Sub menu "thankyou". Options: l == list current donors | a == add donation / donor | m == send mail to every donor | x == exit submenu\n'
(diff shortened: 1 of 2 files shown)


students/elmar_m/lesson05/mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def report():
+    '''
+    Print a formatted list of current donors and donations
+    '''
+    # get the highest number of digits to create formatstring accordingly:
+    sumlist = []
+    for i in donors.values():
+        sumlist.append(sum(i))
```
natasha-aleksandrova: this for loop could be converted to a list comprehension

students/elmar_m/lesson05/mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+    '''
+    try:
+        while True:
+            response = input(p)
+            if d[response]() == 'exiting':
+                break
+    except KeyError:
+        print('sorry, unknown option:', response)
+        print(p)
+        menu(p, d)
```
natasha-aleksandrova: you should put your KeyError around line 172, and inside of the except just the print statement so that the while True can continue to run

PR #346, opened 2018-03-24
Modified functions to use slicing

shibinmathew:
Modified functions to use slicing and used linter to fix Pep-8 violations

### DIFF
--- a/students/shibin_mathew/lesson3/a_new_file.py
+++ b/students/shibin_mathew/lesson3/a_new_file.py
@@ -18,43 +18,44 @@ def exchange_first_last(seq):
 
 def every_other_removed(seq):
 	#Every other item removed
-	tup_b=()
-	str_length = len(seq)
-	new_str = ''
-	if isinstance(seq, str): #if sequence is a string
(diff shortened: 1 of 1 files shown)


students/shibin_mathew/lesson3/a_new_file.py
```diff
@@ -18,43 +18,44 @@ def exchange_first_last(seq):
...
-			if i % 2 == 0: #even value
-				new_str+=item #concatenate every other values
-		return new_str #return new string
-	else:
-		for i, item in enumerate(seq):
-			if i%2 ==0:
-				tup_b = tup_b+(item,)
-		return tup_b
+
+	return seq[:-1:2]
```
natasha-aleksandrova: note that you can just use `seq[::2]`

students/shibin_mathew/lesson3/a_new_file.py
```diff
@@ -18,43 +18,44 @@ def exchange_first_last(seq):
```
natasha-aleksandrova: `seq[-1:] + seq[1:-1] + seq[:1]` <- the answer to exchange first last method which will work for list, string and a tuple

PR #350, opened 2018-03-24
Lesson 03 Slicing Lab Exercise

traub01-devtrain:
Adding Lesson 03 Slicing Lab Exercise


students/JamesTraub/lesson03/slicing.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+# remove_every_other(tuple_seq)
+
+
+def remove_firstLast4_every_other(seq):
+   """ Create a function with the first 4 and 
+   the last 4 items removed, and then every other item in between. """
+   first_last_removed = seq[4:-4]
+   a_new_seq3 = first_last_removed[1::2]
```
natasha-aleksandrova: note that this can be combined into one line: -> `seq[4:-4:2]`

students/JamesTraub/lesson03/slicing.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+# reversed_elements(tuple_seq)
+
+
+def middle_last_first_third(seq):
+    """ Create a function with the middle third, then last third, 
+    then the first third in the new order. """
+    a_new_seq5 = seq[int(len(seq) / 3):int((len(seq) / 3) * 2)] + seq[-int(len(seq) / 3):] + seq[:int(len(seq) / 3)]
+    print(a_new_seq5)
```
natasha-aleksandrova: for readability it would be better to extract `int(len(seq) / 3)` into a var

PR #352, opened 2018-03-25
Lesson 08 - Emmanuel Twum - Circles

etwum:


students/etwum/lesson08/circle_class.py
```diff
@@ -0,0 +1,52 @@
+from math import pi
+
+
+class Circle():
```
robalford: You don't need to include parentheses when defining a class in Python

students/etwum/lesson08/circle_class.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self,radius):
```
robalford: Remember to include spaces between parameters for Pep8

students/etwum/lesson08/circle_class.py
```diff
@@ -0,0 +1,52 @@
...
+        self.radius = radius
+        self.__area = pi * radius**2
```
robalford: I would move this operation to the area `@property`

students/etwum/lesson08/circle_class.py
```diff
@@ -0,0 +1,52 @@
...
+    def diameter(self):
+        return self.radius * 2
+
+    @diameter.setter
+    def diameter(self, radius):
+        self.radius = radius/2
+
+    @property
+    def area(self):
+        return self.__area
```
robalford: As noted above, I would move the operation here:
```
    @property
    def area(self):
        return pi * self.radius**2
```
And because properties defined in this way are read-only, there's no need for the leading double underscore on the variable name. `circle.area = 6` will throw an `AttributeError` if you don't define a setter for area

students/etwum/lesson08/circle_class.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @classmethod
+    def get_diameter(cls,diameter):
+        return cls(diameter/2)
+
+    def __str__(self):
+        return "Circle with radius: {x:.1f}".format(x=self.radius)
```
robalford: I would just pass in the arg to str.format() instead of assigning to x here

students/etwum/lesson08/circle_class.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __repr__(self):
+        return 'Circle({x})'.format(x=self.radius)
+
+    def __add__(self, other):
+        total_radius = self.radius + other.radius
+        return total_radius
```
robalford: These operations should return new circle objects, rather than radius values

PR #353, opened 2018-03-25
Lesson 04, 05 (exceptions)

stuartkershaw:


students/stuartkershaw/lesson04/dict_lab.py
```diff
@@ -0,0 +1,72 @@
...
+
+d['fruit'] = 'Mango'
+
+print(d)
+
+for key in d:
+    print(key)
+
+for key, value in d.items():
+    print(value)
```
robalford: You can use `d.values()` here

robalford:
@stuartkershaw I've reviewed most of this PR at this point, and just wanted to say that this is really excellent work. You've got mailroom factored nicely into small focused functions and your code is clean and readable throughout. You don't need to include the generated files in your pull requests, as we will run your code if we have any concerns to make sure everything works. We will grade the mailroom and exceptions assignments but I still need to get a closer look at trigrams so let's leave the PR open for now while I finish that up.

robalford:
@stuartkershaw I don't seen your Lesson 4 mailroom assignment on the course website, so be sure to submit it there as well.

stuartkershaw:
Thanks for the feedback @robalford - I gave on the trigrams assignment after just a bit of effort, so I'm not expecting full credit there :) I'll return to it eventually, but will probably keep moving along with the other stuff for now. I thought I'd submitted lesson 4 mailroom, but will check again now and submit again. Thanks!

stuartkershaw:
@robalford Ah, it looks like lesson 4 is submitted, but I'm still underway with lesson 5 mailroom. Should have that submitted tomorrow AM.

students/stuartkershaw/lesson04/trigrams/trigrams.py
```diff
@@ -0,0 +1,23 @@
...
+
+for i, el in enumerate(collection):
+    if i < len(collection) - 3:
+        first = el
+        second = collection[i + 1]
+        key = (first, second)
+        if key not in trigram:
+            trigram[key] = [collection[i + 2]]
+        else:
+            trigram[key].append(collection[i + 2])
```
robalford: Check out `defaultdict` from the `collections` module in the Python standard library, or you can use `dict.setdefault()` to either return an existing key or set a new one to a default value if it doesn't exist

robalford:
@stuartkershaw sounds good, let's go ahead and close out this PR then and you can resubmit trigrams whenever you have a chance to come back to it. A made one minor note there just now for you to incorporate. @natasha-aleksandrova this is ready to merge.

PR #356, opened 2018-03-26
lesson 2 added

pythonKel:
Hello, I have added the lesson 2 files for Fizzbuzz and Series. I'm having trouble getting my head wrapped around the variable grid assign, and being behind, I decided to just pass it by and move on.
Thanks,
Kelly


students/KellyM/Lesson2/lesson 2 - series.py
```diff
@@ -0,0 +1,48 @@
+def fib(n):
+    """ This function calculates a Fibonacci Series, which is a series that
+        starts with 0 and 1, then calculates the next number by adding the
+        previous two numbers."""
+    if n == 1:
```
robalford: This should produce the series starting with 0 index (like you did for Lucas below) so that `fib(0) == 0, fib(1) == 1` etc.

students/KellyM/Lesson2/lesson 2 - series.py
```diff
@@ -0,0 +1,48 @@
...
+        return 1
+    elif n==2:
```
robalford: The Python Pep8 style guide prefers whitespace around operators. We recommend that you check your code against an online Pep8 linter, or configure your text editor to catch Pep8 violations as you type. There are instructions in the course materials for how to set this up for SublimeText, and the PyCharm IDE will provide this functionality by default.

students/KellyM/Lesson2/lesson 2 - series.py
```diff
@@ -0,0 +1,48 @@
...
+        return 1
+    else:
+        return (fib(n-1) + fib(n-2))
```
robalford: Nice work identifying the recursive solution to this problem! FYI the outer parentheses are not necessary in the return statement

students/KellyM/Lesson2/lesson 2 - series.py
```diff
@@ -0,0 +1,48 @@
...
+def sum_series(n, o=0, p=1):
+    """This function calculates either the Fibonacci Series or the Lucas
+        Numbers, depending on if additional parameters are provided. If no
+        optional parameters, the Lucas Numbers will be calculatted,
+        otherwise, the Fibonacci Series will be caluslated."""
+    if o or p is None:
+        return fib(n)
+    else:
+        return lucas(n)
+
```
robalford: The goal here is to create a series that defaults to fibonacci but can take any two numbers as the
first two values to produce a new series of numbers. You've got the right idea in setting the keyword arguments to 0 and 1 to produce the fibonacci series, but your function should work in such a way that the lucas series
would be produced by calling with `o=2` and `p=1` as the keyword arguments, so that `sum_series(7, 2, 1) == 29`. Other series could then be produced by substituting new values for the optional keyword arguments

robalford:
@pythonKel remember to submit your assignments on the course website as well along with a link to your pull request. I don't currently see submissions from you for either of these assignments

pythonKel:
Natasha / Rob,

Thanks for the welcome and the feedback. Will take another look at it this
week.

I wasn't aware we were to post to the course, attach a link and use github.
I thought it was strictly github.

Will check out the Pep8 guide too.

Cheers,

Kelly

natasha-aleksandrova:
@pythonKel  no worries, at the very bottom of each assignment is the submission area, you will want to submit a PR link that looks like: `https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/356`

you will also need to upload files there as well (uw policy for records keeping)

PR #357, opened 2018-03-27
Completed Lesson 2 Exercises

TrkCat:
Completed Lesson 2 exercises (and initial break_me.py lesson 1 exercise).

### DIFF
--- a/students/TrkCat/Lesson02/drawGrid.py
+++ b/students/TrkCat/Lesson02/drawGrid.py
@@ -1,0 +1,34 @@
+def gridPart1():
+    """Print a specific center divided 9 x 9 grid"""
+    for line in range(21):
+        if not line % 10:
+            print('+' + ' -' * 4 + ' + ' + '- ' * 4 + '+')
+        elif line % 2:
+            print(' ' * 21)
+        else:
+            print('|' + ' ' * 9 + '|' + ' ' * 9 + '|')
+            
+            
+def gridPart2(n):
+    """Print a center divided n x n grid"""
+    n = (n // 2) * 2 + 1
+    gridSize = 2 * n + 2
+    for line in range(gridSize + 1):
+        if not line % (n + 1):
+            print('+','- ' * (n // 2) + '+','- ' * (n // 2) + '+')
+        elif line % 2:
+            print(' ' * gridSize)
+        else:
+            print('|' + ' ' * n + '|' + ' ' * n + '|')    
+    
+    
+def gridPart3(b,n):
+    """Print an n x n grid repeated b x b times"""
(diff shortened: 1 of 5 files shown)


students/TrkCat/Lesson02/drawGrid.py
```diff
@@ -0,0 +1,34 @@
+def gridPart1():
```
natasha-aleksandrova: per pep8 (python style guide): function names and var names should be snake_case -> `grid_part1()`
same for modules (file names) -> `draw_grid.py`

students/TrkCat/Lesson02/fizzbuzz.py
```diff
@@ -0,0 +1,10 @@
+for num in range(1:101):
+    str = ('')
```
natasha-aleksandrova: `str` is a built-in python function which you are overwriting

github highlights built-ins for you (so do some IDEs), so make sure to use alternate var names

students/TrkCat/Lesson02/fizzbuzz.py
```diff
@@ -0,0 +1,10 @@
+for num in range(1:101):
```
natasha-aleksandrova: `range(1:101)` is an invalid syntax, should be `range(1, 101)`

students/TrkCat/Lesson02/fizzbuzz.py
natasha-aleksandrova: also, the parens are redundant

natasha-aleksandrova:
@TrkCat make sure to submit github PR link on the course website so that the assignments can get graded.

students/TrkCat/Lesson02/fizzbuzz.py
TrkCat: Thank you! Definitely don't want to be in the habit of overwriting built-in functions, I'll update the name.

students/TrkCat/Lesson02/drawGrid.py
TrkCat: Thanks! It'll take me awhile to break my camel case habits.

students/TrkCat/Lesson02/fizzbuzz.py
TrkCat: Could have sworn I checked this code through, but looks like I made an (incorrect) update afterwards. Thanks much, will update.

students/TrkCat/Lesson02/drawGrid.py
natasha-aleksandrova: understandable, it's challenging to switch between different languages with different style guidelines 🙂

PR #358, opened 2018-03-27
Lesson 8 assignment

kuhnbt7:


students/kuhnbt/lesson8/circle_class.py
```diff
@@ -0,0 +1,93 @@
...
+    def area(self):
+        return self._radius**2 * pi
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        if diameter <= 0:
+            raise ValueError('Diameter must be positive')
+        self = cls(diameter / 2)
+        self._diameter = diameter
+        return self
```
natasha-aleksandrova: it is typically just this -> `return cls(diameter / 2)`

diameter is already set in the `__init__` function so need need to set it again

students/kuhnbt/lesson8/test_circle.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class CircleTest(unittest.TestCase):
+    def setUp(self):
+        self.c = cc.Circle(4)
+
+    def test_radius(self):
+        self.assertEqual(self.c.radius, 4)
+        with self.assertRaises(ValueError):
+            f = cc.Circle(-2)
```
natasha-aleksandrova: this would typically be a separate test (error case one)

PR #359, opened 2018-03-27
submitting mailroom2

githubtater:


students/Roy_Tate/lesson04/dict_lab.py
```diff
@@ -0,0 +1,23 @@
...
+# Dictionaries 1 #
+
+print(user_dict)
+del user_dict['cake']
+print(user_dict)
+user_dict['fruit'] = 'Mango'
+print(user_dict)
+
+print(user_dict.keys())
+print('\nCake in user_dict? ' + str(('cake' in user_dict.keys())))
```
robalford: The `in` operator checks for key membership in dict, so this can be `'cake' in user_dict`

students/Roy_Tate/lesson04/dict_lab.py
```diff
@@ -0,0 +1,23 @@
...
+print('Mango in user_dict?  ' + str(('Mango' in user_dict['fruit'])))
```
robalford: This would be `'Mango' in d.values()` to check if 'Mango' is a value in the dict

students/Roy_Tate/lesson04/mailroom2.py
```diff
@@ -0,0 +1,184 @@
+# author/student: Roy Tate (githubtater)
+
+from collections import Counter
+
+import sys
+import os
+import os.path
+
+# Create the donor_dict and manually add a few entries
+donor_dict = {}
```
robalford: It's more conventional in Python to define a dict using literal syntax:
```
donor_dict = {
    'Bill': [1100, 2100, 3100],
    'Mark': [1000, 300000.23, 2000],
}
```
The syntax you use below is generally used for adding keys/values to an existing dict.

students/Roy_Tate/lesson04/mailroom2.py
```diff
@@ -0,0 +1,184 @@
...
+donor_dict['Bill'] = [1100, 2100, 3100]
+donor_dict['Mark'] = [1000, 300000.23, 2000]
+donor_dict['Jeff'] = [50, 450.355]
+donor_dict['Roy'] = [4500, 7500, 11221, 30232, 323]
+
+# This is the standard template to be used when sending thank you emails. Ideally this would be saved
+# externally as a template file.
+email_str_ty = '''\n
```
robalford: Python will retain and preserve new lines in triple quoted strings, so there's no need to use the `\n` character here

students/Roy_Tate/lesson04/mailroom2.py
```diff
@@ -0,0 +1,184 @@
...
+
+        Sincerely,
+
+        The good guys at the best organization'''
+
+
+# Prints the list of keys from donor_dict (in this case, the donor names)
+def print_list():
+    print('{:*^50}'.format('DONOR LIST'))
+    for donor in donor_dict.keys():
```
robalford: No need to call `keys()` here

students/Roy_Tate/lesson04/mailroom2.py
```diff
@@ -0,0 +1,184 @@
...
+# Helper function to add donations to appropriate donor's account
+def add_donation(name, new_donation):
+    if name in donor_dict.keys():
+        donor_dict[name].append(new_donation)
+    else:
+        donor_dict[name] = [new_donation]
+
+
+# Send formatted email to donor
+def print_email(email_str, donor_name, donation_amount):
```
robalford: The `email_str` parameter is not being used in your function, but the `email_str_ty` variable in your function block is referencing the variable defined in the global namespace so the function still works

students/Roy_Tate/lesson04/mailroom2.py
```diff
@@ -0,0 +1,184 @@
...
+    # Obtain the name of the donor, the contribution amount, add the donation to the list, and generate
+    # a thank you email.
+
+    name = input("Enter the full name of the recipient (Type 'list' to view donors. Type '1' for Main Menu):\n")
+    # thank_you_args = {
+    #                'list': print_list,
+    #                '1': main,
+    # }
+    # thank_you_prompt = "Enter the full name of the recipient (Type 'list' to view donors. " \
+    #                    "Type '1' for Main Menu):\n"
```
robalford: It's a good idea to remove commented code from your files before committing. Since git stores a history of your entire project (as long as you commit your changes), there's never really a need to keep old dead code lying around

students/Roy_Tate/lesson04/mailroom2.py
```diff
@@ -0,0 +1,184 @@
...
+        all_str += donor_str_fmt.format(donors[0], donors[1], donors[2], donors[3]) + '\n'
+    print(all_str)
+
+
+# take a prompt as a string and a list of options/arguments as a dict
+def menu_selection(prompt, arg_dict):
+    while True:
+        try:
+            response = input(prompt)
+            arg_dict[response]() == ''
```
robalford: The comparison to `''` doesn't appear to serve any purpose here

students/Roy_Tate/lesson04/mailroom2.py
```diff
@@ -0,0 +1,184 @@
...
+        except:
```
robalford: You've got the right idea here with the try/except block and we'll get into exceptions in more detail in lesson 5. However, you always want to catch a specific exception rather use a bare except statement like this. The problem with a bare except is that it will catch any exception that's raised, rather than just one (or more) that you expect to occur, and this can lead to bugs when errors pass through your code 'silently'. 

There's currently a bug like this in your program where if I try to exit from the command line by entering '4' to select the quit option, I see the following message and return to the original menu selection prompt:
```
Exiting the mailroom2.
Invalid Response. Try again.
```
What's happening here is that when your `die()` function calls `sys.exit()`, that function tries to exit the interpreter by raising the `SystemExit()` exception. Exceptions in Python will 'bubble up' the function call stack if they're not handled, so when this exception reaches your `except` statement in the `menu_selection()` function, it is caught and 'handled' by printing the error message and returning to the main prompt in your while loop.

It seems like the exception that you'd want to catch here is `KeyError`, because that will be raised if the user enters a value that is not a key in your `arg_dict`. You also want to try and only include code in your `try` block that might actually raise the exception, so I would suggest moving the call to input out of your try block and just including the call to the `arg_dict` function in there. Check out the course materials on exception handling in Lesson 5 for more details on best practices, you'll be required to use exception handling in mailroom for that assignment so you're getting a head start.

PR #360, opened 2018-03-27
lesson03 submission

prgrover:


students/prgrover/lesson03/list_lab.py
```diff
@@ -0,0 +1,109 @@
...
+
+fruit_list.pop()
+
+print ("Here is our updated list of fruits after removing the last item:",fruit_list, "\n")
+
+delete_fruit = input("Please enter a fruit you would like to have deleted: ")
+
+for fruit in fruit_list:
+    if fruit == delete_fruit:
+        fruit_list.remove(fruit)
```
robalford: You could also use a while loop here: 
```
while fruit in fruit_list:
    fruit_list.remove(fruit)
```

students/prgrover/lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+    * If the user types ‘list’, show them a list of the donor names and re-prompt
+    * If the user types a name not in the list, add that name to the data structure and use it.
+    * If the user types a name in the list, use it.
+    * Once a name has been selected, prompt for a donation amount.
+    * Turn the amount into a number – it is OK at this point for the program to crash if someone types a bogus amount.
+    * Once an amount has been given, add that amount to the donation history of the selected user.
+    * Finally, use string formatting to compose an email thanking the donor for their generous donation. Print the email to the terminal and return to the original prompt.
+
+    Args: None
+    """
```
robalford: Check out PEP 257 for Python multi-line doc string conventions: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

students/prgrover/lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+        return
+
+    while donor_name == "list":
+        for donor in donor_list:
+            print (donor[0])
+        donor_name = input("Please enter the full name of a donor. You can also enter list to see all current donors or e to exit: ")
+
+    for donor in donor_list:
+        if donor_name == donor[0]:
+            donor_found = 0
```
robalford: I think it would make more sense to assign `donor_name = True` here. It's confusing to use 0, since 0 is the numerical equivalent of the `False` boolean

students/prgrover/lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+        if donation_amount == "e":
+            return
+        donor[1].append(float(donation_amount))
+        print(email.format(donor_name, donation_amount))
+    else:
+        donation_amount = input("Please enter the donation amount or e to exit: ")
+        if donation_amount == "e":
+            return
+        donor_list.append([donor_name, [float(donation_amount)]])
+        print(email.format(donor_name, donation_amount))
```
robalford: I would suggest using a `while True` loop in the section to set up a loop that will go on until an exit condition (or `break` statement) is encountered. That way you can collect your user input and eliminate some of the repeated prompts and conditionals.

students/prgrover/lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+        print ("{:30s}  ${:11.2f}   {:9d}  ${:12.2f}".format(donor_name, total_given, num_gifts, average_gift))
+
+
+def user_option():
+    """
+    Prompt the user (you) to choose from a menu of 3 actions: “Send a Thank You”, “Create a Report” or “quit”)
+
+    Args: None
+    """
+    quit = 1
```
robalford: It's a good idea to avoid naming your variables after Python built-in names and functions like `quit` and I would use `while True` or some other more readable name like `while program_running` for your loop below.

students/prgrover/lesson03/slicing.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Args:
+    n:  Required. A sequence.
+    """
+
+    #Assign the value for 1/3 the length of n to x for easier readability. 
+    x = int(len(n) / 3)
+
+
+    return n[x:2*x] + n[2*x:] + n[:x]
```
robalford: you could use negative slices here e.g. `n[x:-x]` for the first section

students/prgrover/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,93 @@
...
+to take an arbitrary number of values.
+
+Args:
+seq: Tuple with an arbitraty number of values. Required.
+"""
+task_3 = (3, 10, 23, 32, 42)
+
+def formatter(seq):
+    l = len(seq)
+    return (("The {} numbers are: " + ", ".join(["{}"] * l)).format(l, *seq))
```
robalford: Don't need both sets of parentheses here

students/prgrover/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,93 @@
...
+
+And for an extra task, given a tuple with 10 consecutive numbers, can you work how to quickly print the tuple in columns that are 5 charaters wide? It’s easily done on one short line!
+"""
+cars = [('Honda', '2000', 999.00), ('Toyota', '2018', 30000.23), ('Acura', '2008', 9995.99), ('Infinity', '2015', 23000.00)]
+
+for car in cars:
+    print("{:10} {:10} ${:9.2f}".format(*car))
+
+task_6 = (0,1,2,3,4,5,6,7,8,9)
+print (("{:5d}" * 10).format(*task_6))
```
robalford: Remove whitespace before the `(` in the function call here. I'd recommend running your code through a Pep8 linter or configuring your editor to catch pep8 violations as you type to learn the Python style rules

PR #361, opened 2018-03-27
Pushing finished trigram.py

hortw1991:


students/thorn/lesson04/trigram.py
```diff
@@ -0,0 +1,118 @@
...
+        digrams = create_digrams(text_list)
+        trigrams = create_trigrams(text_list)
+        files[choice](trigrams)
+
+    elif choice == 2:
+        intext = os.path.join(os.getcwd(), 'sherlock.txt')
+        text_list = read_text(intext)
+        digrams = create_digrams(text_list)
+        trigrams = create_trigrams(text_list)
+        files[choice](trigrams)
```
natasha-aleksandrova: nice touch!

what you want to do here is set your file name in your if / else and then move the rest of the code because it is duplicated:
```
if choice == 1:
    filename = "small.text"
elif choice == 2:
    filename = "large.text"

intext = os.path.join(os.getcwd(), filename)
....

File: students/thorn/lesson04/trigram.py [hunk 2]
```diff
@@ -0,0 +1,118 @@
...
+    Ex: ([0],[0+1],[0+2])
+    """
+    trigrams = []
+    for i in range(len(text_list)-2):
+        trigrams.append((text_list[i], text_list[i+1], text_list[i+2]))
+
+    # Create set and shuffle for randomness.
+    random.shuffle(trigrams)
+
+    trigram_set = set(trigrams)
```
natasha-aleksandrova: the trigrams created here are not exactly as per instructions on the course website:
this text -> "I wish I may I wish I might"
should produce a map where you have a key of two consecutive words as a key and values which are words that are possible to follow the key:
```
"I wish" => ["I", "I"]
"wish I" => ["may", "might"]
"may I"  => ["wish"]
"I may"  => ["I"]
```

I don't see that happening here

students/thorn/lesson04/trigram.py
natasha-aleksandrova: I'm also not sure why there are two separate functions, it seems that it should only differ in the filename

students/thorn/lesson04/trigram.py
hortw1991: Oooh, I see now.  I completely misunderstood the concept and thought that I needed to just use each word once based on the trigram value versus ALL possible words that follow the key.

I'll fix this up and resubmit.  I will admit that this was one where the output was conceptually difficult for me to grasp, so no promises :)

natasha-aleksandrova: 🙂 lets us know if you need some help clarifying anything

essentially you should be creating a "map" of two words to a list of all possible words that follow it (hint: dictionary)

PR #364, opened 2018-03-28
Lesson 05 Mailroom

stuartkershaw:
I was able to refactor with improved exception handling. I had a harder time finding a good use for comprehensions though... perhaps it's just the way I've organized things. I was able to do some condensing around line 135 with the zip function, but I'm not sure I would use this in production. Perhaps it would make better sense in a situation where my variable names were different from the key names of the dict I'm building up.

### DIFF
--- a/students/stuartkershaw/lesson05/mailroom.py
+++ b/students/stuartkershaw/lesson05/mailroom.py
@@ -101,12 +101,16 @@ def set_donation(donor):
             get_selection()
 
 
+def list_donors():
+    for name in donors:
+        print(name)
+
+
 def get_donor_name():
     instruction = 'Please enter a full name or type \'list\' to see donors:\n'
     name_input = input(instruction)
     if name_input == 'list':
-        for name in donors:
-            print(name)
+        list_donors()
         get_donor_name()
     elif name_input in donors:
         set_donation(name_input)
@@ -118,12 +122,9 @@ def get_donor_name():
 def generate_rollup():
     for donor in donors:
         cur_donor = donors[donor]
-        if 'rollup' not in cur_donor:
-            cur_donor['rollup'] = {}
-        cur_rollup = cur_donor['rollup']
-        cur_rollup['number'] = len(cur_donor['donations'])
-        cur_rollup['total'] = sum(cur_donor['donations'])
-        cur_rollup['average'] = float(
+        number = len(cur_donor['donations'])
+        total = sum(cur_donor['donations'])
+        average = float(
             format(
                 sum(
                     cur_donor['donations']) / len(
@@ -131,6 +132,7 @@ def generate_rollup():
                     ), '.2f'
                 )
             )
+        cur_donor['rollup'] = dict(zip(('number','total','average'), (number, total, average)))
 
 
 def show_donor_table():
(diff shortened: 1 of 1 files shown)


students/stuartkershaw/lesson05/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+    while True:
+        try:
+            donation = int(input('Please enter a donation amount: '))
+            assert donation > 0
+        except AssertionError:
+            print('__AssertionError__ Please provide a whole number greater than zero.')
+            set_donation(donor)
+        except ValueError:
+            print('__ValueError__ Please provide a whole number.')
+            set_donation(donor)
```
natasha-aleksandrova: since you are in a `while True` loop you don't need to call `set_donation(donor)` as it will continue the loop

if you want to print exception name (I wouldn't, since it's a user program):
```
except ValueError as e:
    print(e.__class__.__name__)
```

students/stuartkershaw/lesson05/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+                apply_selection(selection)
+        except ValueError:
+            input(options)
+
+
+def set_donation(donor):
+    while True:
+        try:
+            donation = int(input('Please enter a donation amount: '))
+            assert donation > 0
```
natasha-aleksandrova: you don't usually use asserts in production code, you can raise ValueError here since you're already catching that.

students/stuartkershaw/lesson05/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+              '2) create a report\n'\
+              '3) send letters to everyone\n'\
+              '4) quit\n'
+    while True:
+        try:
+            selection = input(options)
+            if selection in ['1', '2', '3', '4']:
+                apply_selection(selection)
+        except ValueError:
+            input(options)
```
natasha-aleksandrova: in what case would the ValueError be raised?

if it does happen, the user is not captured, is that intended?

recommendation: 
* use `arg_dict[selection]()` and catch `KeyError` in `apply_selection` function and print a message to user then let the loop continue

students/stuartkershaw/lesson05/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+        except ValueError:
+            print('__ValueError__ Please provide a whole number.')
+            set_donation(donor)
+        else:
+            if 'donations' in donors[donor]:
+                donors[donor]['donations'].append(donation)
+            else:
+                donors[donor]['donations'] = [donation]
+            print(compose_email(donor))
+            get_selection()
```
natasha-aleksandrova: I don't think you need the call to `get_selection()` since the loop should continue

students/stuartkershaw/lesson05/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def list_donors():
+    for name in donors:
+        print(name)
+
+
+def get_donor_name():
```
natasha-aleksandrova: I would call this function `send_thank_you` or something along those lines as it appears to get some information (get_*) but has a side effect to make changes.

PR #366, opened 2018-03-29
submitting mailroom4.py and mailroom4_test.py for instructor review

wesmoy17:
Changed mailroom3.py into mailroom4.py by:
Separating user input selection into its own function away from main file
Included an if statement in the main file to rerun user input instead of after each menu function

Mailroom4_test.py checks:
Closing the test program prints the closing message
Typing list in option one prints out the expected list of donors
A mock attempt of the main menu adding a person to the donor list and quitting produces the same print out
Printing out the report with the expected result
Typing a correct and incorrect value into the input function produces the expected result
Executing the letters function prints out the expected result and the message inside each file is as expected

### DIFF
--- a/students/WesleyM/Lesson6/mailroom4.py
+++ b/students/WesleyM/Lesson6/mailroom4.py
@@ -1,0 +1,80 @@
+donors = {
+  'Alice Adams Ron': [20] ,
+  'Bob Be-Lake': [100],
+  'Charles Cruz': [30, 50, 10],
+  'Denise Dnice': [10, 5],
+  'Edward Eduardo': [25, 20, 20, 20, 10]
+  }
+ 
+
+def send_ty(donors):
+  donor = input("Enter a donor name or input 'List' for a list of donors\n>")
+  repeat = 0
+  if donor.lower() == "list":
+    for names in donors:
+        print(names)
+  else:
+    while repeat == 0:
+      try:
+        donation = float(input("Enter a donation amount for {} \n>".format(donor)))
+      except ValueError:
+        print("Enter donations numerically")
+      else:
+        for names, donations in donors.items():
+          if donor == names:
+            donors[names].append(donation)
+            break
+        else:
+          donors[donor] = [donation]
+        print("Thank you {} for your donation of ${:.2f}".format(donor, donation))
+        repeat = input("Would you like to add another donation for {}? Type 'yes' or 'no' \n>".format(donor))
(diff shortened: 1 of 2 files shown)


students/WesleyM/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,80 @@
...
+
+def report(donors):
+    print("Donor Name           | Total Given   | Num Gifts     | Average Gift  |\n" + "-"*70)
+    for name, donations in donors.items():
+        total = sum(donations)
+        num = len(donations)
+        print("|{:<20}|{:>15.2f}|{:>15}|{:>15.2f}|".format(name, total, num, total/num))
+
+def close_program(donors):
+    print('\nClosing Program\n')
```
natasha-aleksandrova: this should not have input var (in the case of this program just use the global donors data structure directly)

students/WesleyM/Lesson6/mailroom4_test.py
```diff
@@ -0,0 +1,88 @@
...
+def clear_screen():
+    sys.stdout = sys.__stdout__
+
+class TestMailRoom4(unittest.TestCase):
+
+    def test_close_program(self):
+        mailroom4.close_program(donors) == '\nClosing Program\n'
+
+    @patch('mailroom4.input', side_effect = ["list"])
+    def test_send_ty(self, mock_input):
```
natasha-aleksandrova: better to name `test_send_ty_list_option`

students/WesleyM/Lesson6/mailroom4_test.py
```diff
@@ -0,0 +1,88 @@
...
+        capture_print = print_screen()
+        mailroom4.report(donors)
+        clear_screen()
+        table_contents = 'Donor Name           | Total Given   | Num Gifts     | Average Gift  |\n'
+        table_contents += '----------------------------------------------------------------------\n'
+        for name, donations in donors.items():
+            total = sum(donations)
+            num = len(donations)
+            table_contents += "|{:<20}|{:>15.2f}|{:>15}|{:>15.2f}|\n".format(name, total, num, total/num)
+        self.assertEqual(capture_print.getvalue(), table_contents)
```
natasha-aleksandrova: your expected would be better with explicit string rather than copied code from the function you're testing.

natasha-aleksandrova:
@wesmoy17 please be sure to submit a PR link on the course website, thanks!

PR #368, opened 2018-03-29
first commit

JeroenSweerts:
Hi,

Please review my code for the Mailroom Part 4 assignment.

Thanks.

Best regards,

Jeroen

### DIFF
--- a/students/JeroenSweerts/Lesson6/mailroom4.py
+++ b/students/JeroenSweerts/Lesson6/mailroom4.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+
+donations = {}
+donations['papa'] = [100, 5, 15]
+donations['mama'] = [12, 200, 2, 66]
+donations['bompa'] = [1000]
+donations['bobonne'] = [500, 500]
+donations['onbekende'] = [1000000]
+
+
+def thankyou():
+    """If the user types ‘list’, show them a list of the donor names and re-prompt
+    If the user types a name not in the list, add that name to the data structure
+    and use it.
+    If the user types a name in the list, use it.
+    Once a name has been selected, prompt for a donation amount.
+    Turn the amount into a number – it is OK at this point for the program to
+    crash if someone types a bogus amount.
+    Once an amount has been given, add that amount to the donation history of
+    the selected user.
+    Finally, use string formatting to compose an email thanking the donor for their
+    generous donation. Print the email to the terminal and return to the original
+    prompt."""
+    while True:
+        name = input('Please enter a name or type "list" to see a list of names > ')
+        if name == 'list':
+            print(donations.keys())
+        else:
+            break
+    if name not in donations.keys():
+        donations[name] = []
+    try:
+        donation = input('Please enter a donation amount > ')
+        donations[name].append(int(donation))
+    except ValueError:
+        while not donation.isdigit():
+            print("You didn't enter a valid amount.")
+            donation = input('Please enter a donation amount > ')
+        donations[name].append(int(donation))
+    d = {'name':name, 'donation':donation}
+    print("Dear {name} , thank you very much for your generous donation of {donation}. Looking forward receiving even more money next time.".format(**d))
+    print(donations.keys())
+    print(donations.values())
+
+
+def createfile(name):
+    '''create .txt file containing thank you letter for each person who made a donation'''
+    d = {'name':name, 'donation':donations[name][-1]}
+    outfile = open(name+'.txt', 'w')
+    outfile.write("Dear {name} , \n\n".format(**d))
+    outfile.write("\t Thank you very much for your generous donation of ${donation}. \n\n".format(**d))
+    outfile.write("\t It will be put to very good use. \n\n")
+    outfile.write("\t\t Sincerely, \n")
+    outfile.write("\t\t   -The Team \n")
+    outfile.close()
+
+
+def thankyoueveryone():
+    [createfile(name) for name in donations.keys()]
+
+
+def report():
+    """ If the user (you) selected “Create a Report”, print a list of your donors,
+     sorted by total historical donation amount.
+     Include Donor Name, total donated, number of donations and average donation
+     amount as values in each row. You do not need to print out all their
+     donations, just the summary info.
+     Using string formatting, format the output rows as nicely as possible.
+     The end result should be tabular (values in each column should align with
+     those above and below)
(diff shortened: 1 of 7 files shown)


students/JeroenSweerts/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python3
+
+donations = {}
+donations['papa'] = [100, 5, 15]
+donations['mama'] = [12, 200, 2, 66]
+donations['bompa'] = [1000]
+donations['bobonne'] = [500, 500]
+donations['onbekende'] = [1000000]
```
robalford: The convention in Python is to use dict literal syntax when defining a new dict with some initial data:
```
donations = {
    'papa': [100, 5, 15],
    'mama': [12, 200, 2, 66],
}
```
The bracket syntax you're using is generally used for adding new data to an existing dict.

students/JeroenSweerts/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,105 @@
...
+    while True:
+        name = input('Please enter a name or type "list" to see a list of names > ')
+        if name == 'list':
+            print(donations.keys())
+        else:
+            break
+    if name not in donations.keys():
+        donations[name] = []
+    try:
+        donation = input('Please enter a donation amount > ')
```
robalford: If possible, you should only include the code that might raise the expected exception inside a try block. In this case, it's the call to `int()` that might raise a `ValueError` so I'd move the `donation` assignment outside of the `try` block

students/JeroenSweerts/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,105 @@
...
+        donations[name].append(int(donation))
+    except ValueError:
+        while not donation.isdigit():
+            print("You didn't enter a valid amount.")
+            donation = input('Please enter a donation amount > ')
+        donations[name].append(int(donation))
+    d = {'name':name, 'donation':donation}
```
robalford: Add space after colon in dict assignment for Pep8

students/JeroenSweerts/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,105 @@
...
+
+def createfile(name):
+    '''create .txt file containing thank you letter for each person who made a donation'''
+    d = {'name':name, 'donation':donations[name][-1]}
+    outfile = open(name+'.txt', 'w')
+    outfile.write("Dear {name} , \n\n".format(**d))
+    outfile.write("\t Thank you very much for your generous donation of ${donation}. \n\n".format(**d))
+    outfile.write("\t It will be put to very good use. \n\n")
+    outfile.write("\t\t Sincerely, \n")
+    outfile.write("\t\t   -The Team \n")
```
robalford: I would move the letter formatting to a separate function that returns the formatted letter and then write the whole string to the file here

students/JeroenSweerts/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,105 @@
...
+    outfile.close()
+
+
+def thankyoueveryone():
+    [createfile(name) for name in donations.keys()]
```
robalford: List comprehensions are intended to have return values, so this would be better formatted as a standard for loop

students/JeroenSweerts/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,105 @@
...
+     The end result should be tabular (values in each column should align with
+     those above and below)
+     After printing this report, return to the original prompt.
+     At any point, the user should be able to quit their current task and return
+     to the original prompt.
+     From the original prompt, the user should be able to quit the script cleanly."""
+    print(f'{"Donor Name":20s} {"|  Total Given":20s} {"|  Num Gifts  |":20s} {"Average Gift":20s}')
+    print(f'{"-"*76}')
+    [print(f'{name:20s} ${sum(donations[name]):20.2f} {len(donations[name]):13d}${sum(donations[name])/len(donations[name]):20.2f}') for name in donations.keys()]
+
```
robalford: See my note above about list comprehensions. From the Python [docs](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions):

"List comprehensions provide a concise way to create lists. Common applications are to make new lists where each element is the result of some operations applied to each member of another sequence or iterable, or to create a subsequence of those elements that satisfy a certain condition."

students/JeroenSweerts/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,105 @@
...
+    except:
+        while not response.isdigit():
+            print("You didn't enter a number.")
+            print()
+            response = input('Please choose between the following 3 actions:\
+            1. Send a Thank You; 2. Create a Report; 3. Send letters to everyone; 4. quit >  ')
+    arg_dict = {1:thankyou, 2:report, 3:thankyoueveryone}
+    while int(response) != 4:
+        try:
+            arg_dict.get(int(response))()
```
robalford: I think the only error that you need to catch here is `KeyError`, and that would let you eliminate a lot of the added validation and repeated user prompts because this will be raised by any input value that is not in the `arg_dict`

students/JeroenSweerts/Lesson6/test_mailroom4.py
```diff
@@ -0,0 +1,44 @@
...
+        for name in self.donations.keys():
+            assert os.path.isfile(name+'.txt')
+
+    def test_report(self):
+        text = f'{"Donor Name":20s} {"|  Total Given":20s} {"|  Num Gifts  |":20s} {"Average Gift":20s}\n'
+        text = text + f'{"-"*76}\n'
+        for name in self.donations.keys():
+            text = text + f'{name:20s} ${sum(self.donations[name]):20.2f} {len(self.donations[name]):13d}${sum(self.donations[name])/len(self.donations[name]):20.2f}\n'
+        print(text)
+        self.capturedOutput1 = io.StringIO()
```
robalford: You need to write some content to the `StringIO` object here, or it's value will just be an empty string. This test currently passes because you're comparing the two empty strings returned by the `StringIO` objects in the assert statement. I would suggest reworking your `mailroom4.report()` function so that it returns a string, rather than printing directly to the console, that way you could test the return value here using a similar approach to what you're already doing

PR #369, opened 2018-03-30
submit lesson07

tricaonguyen:


students/tri_nguyen/lesson07/html_render.py
```diff
@@ -0,0 +1,159 @@
+# ------------------------------------------- #
+# Student:   Tri Nguyen
+# Instructor: Natasha Aleksandrova
+# Date:  23-Mar-2018
+# ------------------------------------------- #
+
+# Create an Element class
+
+
+class Element():
```
robalford: You don't need the parentheses for class declaration

students/tri_nguyen/lesson07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+    tag_name = ''
+    indent = '    '  # 4 spaces
+
+    def __init__(self, content=None, **kwargs):
+        ''' initialize the Element object/instance '''
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
+        self.kwargs = kwargs
```
robalford: You might consider renaming this to `attributes` when you reassign here

students/tri_nguyen/lesson07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+        ''' store content to self.content list '''
+        self.content.append(content)
+
+    def render(self, file_out, cur_ind=''):
+        ''' render the tag and the strings in the content '''
+        file_out.write('{}<{}{}>\n'.format(cur_ind, self.tag_name, self.set_attr()))
+
+        '''for attr in self.kwargs:
+            file_out.write(' {}="{}"'.format(attr, self.kwargs[attr]))
+        file_out.write('>\n')'''
```
robalford: The convention in Python for multi-line block comments is to use the `#` character before each line and use triple double quoted strings for function/method docstrings. It's also a good idea to delete commented out code before committing, any old work will be stored in your version control history so you can always go back to it if you want

students/tri_nguyen/lesson07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+            else:
+                file_out.write('{}{}{}\n'.format(cur_ind, self.indent, item))
+
+        file_out.write('{}</{}>\n'.format(cur_ind, self.tag_name))
+
+    def set_attr(self):
+        if self.kwargs:
+            for k, v in self.kwargs.items():
+                attr, val = k, v
+            return ' {}="{}"'.format(attr, val)
```
robalford: On line 212 of `run_html_render.py`, both an `id` and `style` attribute get passed into the `Ul` constructor, but I only see the `style` attribute in your rendered html output. Can you see where you're losing the `id` attribute in the the `set_attr()` method above and correct?

students/tri_nguyen/lesson07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+    ''' render tags like <hr/> and <br/> '''
+
+    ''' this does not work, and not sure why
+    def __init__(self, content=None, **kwargs):
+        super().__init__(content=None, **kwargs)
+        if self.content is not None:
+            raise TypeError('Self closing tags cannot have any content.') '''
+
+    def __init__(self, **kwargs):
+        self.kwargs = kwargs
```
robalford: I'd look at overriding the `append()` method rather than `__init__` and raising an error if the user tries to add any content there

PR #374, opened 2018-04-01
added a readme file

BrandonHenson:
I accidentally did all of this in the wi2018 repo first


students/BrandonHenson/Lesson2/Part1.py
```diff
@@ -0,0 +1,21 @@
+def FunGrid():
```
natasha-aleksandrova: style tip: per pep8 python functions are lower case snake_case

PR #375, opened 2018-04-01
lesson 08 assignment completed

rmart300:


students/rmart300/lesson08/circle.py
```diff
@@ -0,0 +1,72 @@
...
+    @property
+    def radius(self):
+        return self._radius
+
+    @radius.setter
+    def radius(self, radius):
+        if radius < 0:
+            self._radius = 0
+        else:
+            self._radius = radius
```
robalford: The getter and setter don't seem necessary here for `radius`, since you could handle all this logic in the __init__ method

students/rmart300/lesson08/circle.py
```diff
@@ -0,0 +1,72 @@
...
+    @diameter.setter
+    def diameter(self, diameter):
+        if diameter < 0:
+            self._radius = 0
+        else:
+            self._radius = diameter / 2
+
+    @property
+    def area(self):
+        return math.pow(self._radius,2) * math.pi
```
robalford: you can use the `**` operator here for exponent

students/rmart300/lesson08/circle.py
```diff
@@ -0,0 +1,72 @@
...
+        return self
+
+    def __str__(self):
+        return f"Circle with radius: {self._radius}"
+
+    def __repr__(self):
+        return f"Circle({self._radius})" 
+
+    def __add__(self, other_circle):
+       self._radius += other_circle.radius if isinstance(other_circle,Circle) else other_circle
```
robalford: looks like this block is under indented

students/rmart300/lesson08/circle.py
```diff
@@ -0,0 +1,72 @@
...
+       return self
+
+    __iadd__ = __add__
```
robalford: I think it would be more explicit to define these methods separately rather than trying to share the same logic e.g.:
```
    def __add__(self, other):
        return Circle(self.radius + other.radius)

    def __iadd__(self, other):
        self.radius += other.radius
        return self
```

students/rmart300/lesson08/circle.py
```diff
@@ -0,0 +1,72 @@
...
+        self._radius *= factor
+        return self
+
+    __rmul__ = __mul__
+
+    def __lt__(self, other_circle):
+        return True if self._radius < other_circle.radius else False        
+
+    def __eq__(self, other_circle):
+        return True if self._radius == other_circle.radius else False
```
robalford: You can use the [`@functools.total_ordering`](https://docs.python.org/3.6/library/functools.html#functools.total_ordering) class decorator to make sure you're covering all your bases with the comparison methods

robalford:
@rmart300 remember to submit this assignment on the course website as well so we can grade.

PR #376, opened 2018-04-01
seeking feedback on suitability of current 7 tests, suggested number of additional tests to write

johnpharmd:


students/johnpharmd/lesson06/mailroomtest.py
```diff
@@ -0,0 +1,73 @@
...
+        self.make_test_list()
+        assert type(self.test_list) == list
+
+    def test_donor_list_strings(self):
+        self.make_test_list()
+        for s in self.test_list:
+            if '-' not in s:
+                assert s.isalpha()
+            else:
+                assert s.replace('-', '').isalpha()
```
natasha-aleksandrova: if you wanted to test `get_donor_list` function here is an example of a test:
```
def test_get_donor_list():
    actual = mailroom4.get_donor_list()
    expected = ['Gates', 'Brin', 'Cerf', 'Musk', 'Berners-Lee', 'Wojcicki', 'Avey']
    self.assertEquals(actual, expected)
```

in very basic terms you are typically testing actual with expected

`make_test_list` isn't all that helpful and if you wanted to make something available to all tests then use `setUp` method.

students/johnpharmd/lesson06/mailroomtest.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_prompt_title(self):
+        self.test_title = mailroom4.prompt_title()
+        assert self.test_title in ('Prof.', 'Dr.', 'Ms.', 'Mr.')
```
natasha-aleksandrova: unit tests are typically run in automated way, there is usually a system like Jenkins for example that will run your tests on each commit or at build time. Because of that your unit tests cannot typically rely on user interaction.

so how does something like this get tested? that's where we get into the mocking and patching, it is a bit of an advanced concept but simply put you would be "mocking" out (faking) the user input portion and would supply predefined mocked input values for your code.

students/johnpharmd/lesson06/mailroomtest.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_prompt_donation(self):
+        self.test_donation = mailroom4.prompt_donation()
+        assert str(self.test_donation).isdigit()
+
+    def test_add_donor(self):
+        self.new_donor = mailroom4.add_donor('Anderson', 'Mr.')
```
natasha-aleksandrova: no need to use `self` here

students/johnpharmd/lesson06/mailroomtest.py
```diff
@@ -0,0 +1,73 @@
...
+        pass
+
+    def test_send_letters(self):
+        mailroom4.send_letters()
+        cwd_list = os.listdir(os.getcwd())
+        today = datetime.date.today()
+        for donor in mailroom4.get_donors_amts():
+            donor_letter = donor + today.strftime('%Y%m%d') + '.txt'
+            donor_letter = donor_letter.strip('\'')
+            assert donor_letter in cwd_list
```
natasha-aleksandrova: I would say this test is the best depiction of a unit test

natasha-aleksandrova:
I'm going to close this PR since it's just for feedback but you can continue commenting here if you want to have a discussion or have questions.

johnpharmd:
Thanks, Natasha!

PR #381, opened 2018-04-03
Lesson2 (version 2 of file)

kmsnyde:
Completed lesson 2.  I had to make a name change to my funcs after I originally pushed the file and then pushed a "v2" file; I think I deleted original mislabeled file.

There are 3 functions: grid(), print_grid(num); print_grid2(row_col, num).

I think I understand how to get the results in the exercise but probably not in an efficiently coded way (or pythonic way).  Will need to work on it.


students/kmsnyde/lesson2/fizzbuzz.py
```diff
@@ -0,0 +1,13 @@
...
+
+@author: Karl M. Snyder
+"""
+
+num = range(1, 101)
+for i in num:
+    if (i % 3 == 0) and (i % 5 == 0): print('FizzBuzz')
+    elif i % 3 == 0: print('Fizz')
+    elif i % 5 == 0: print('Buzz')
+    else: print(i)
```
natasha-aleksandrova: it is preferred style wise to have the `print` on a separate line

students/kmsnyde/lesson2/series.py
```diff
@@ -0,0 +1,67 @@
...
+    assert len(lucas(5)) == 5 # AssertionError if not 5
+    assert len(sum_series(8)) == 8
+    
+    # Tests if the first number of the fibonacci (0) and lucas (2) sequences
+    # are correct.  AssertionError if False.
+    assert fibonacci(5)[0] == 0
+    assert lucas(5)[0] == 2
+    # Tests if the second element of the sequence (third argument) is correct
+    # Assertion error if False.
+    assert sum_series(5, 3, 2)[1] == 3 #AssertionErrof if not 2
```
natasha-aleksandrova: the functions should be returning fib or lucas number at N index for the sequence.

students/kmsnyde/lesson2/series.py
```diff
@@ -0,0 +1,67 @@
...
+    return list
+lucas(20)
+
+def sum_series(count, start1=0, start2=1):
+    """Creates a sequence of length "count", in which the first 2 elements
+    default to 0 and 1, or which may be changed to any 2 numbers when calling
+    the function using start1/start2 argument positions."""
+    
+    if start1 == 0 and start2 == 1:
+        list = []
```
natasha-aleksandrova: avoid using built in names because it will overwrite the built-in. notice github has a special syntax highlighting for `list`

natasha-aleksandrova:
@kmsnyde saw your message on the submission form.

You shouldn't have any problems adding more commits to your branch while the PR is open. I guess depends on what commands you were trying to run.

If you're still having problems let us know via slack or email!

kmsnyde:
Natasha,

I updated my code in series.py based on your feedback and did the pull request (391); this will still be graded?  I did not update fizzbuzz.py since you already graded it but understand your comment not to put a single statement after else: in the future.  Thanks for your feedback and all that will be in the future. I know this must be a time consuming activity for you.

 

 

Regards,
Karl

 

 

-----Original Message-----

natasha-aleksandrova:
@kmsnyde just to clarify, when we leave feedback in the PR it is usually for your benefit and learning, and it will not affect grading. You may review the comments and do nothing, or do a separate PR with updates if you would like additional feedback.

As far as the grade, you will get a full grade even if we leave suggestions for improvements.

If we determine that assignment submission is incomplete or wrong, we will note such in the PR comment and will recommend another PR with completed or fixed assignment. In such case initial grade would be incomplete.

Let me know if you have any other questions.

PR #383, opened 2018-04-03
Lesson 4 Kata assignment

itsabhister:


students/itsabhi/Lesson4/codekata.py
```diff
@@ -0,0 +1,70 @@
+#! /usr/bin/env python3
+
+
+def wordmap(fh):
```
robalford: It's a good idea to use meaningful names for variables, args and functions

students/itsabhi/Lesson4/codekata.py
```diff
@@ -0,0 +1,70 @@
...
+        line = fh.readline()
+        oneline += line.rstrip('\r')
+        if not line:
+            break
+
+    print(oneline)
+    word_list = oneline.split()
+    n = 0
+    word_dict = {}
+    while(n+2 < len(word_list)):
```
robalford: The parens are not needed here for the while statement.

students/itsabhi/Lesson4/codekata.py
```diff
@@ -0,0 +1,70 @@
...
+        f_word = word_list[n]
+        s_word = word_list[n+1]
+        t_word = word_list[n+2]
+
+        if (f_word + " " + s_word) not in word_dict:  # If key exists in dict
```
robalford: This seems to be the wrong comment here, I'm guessing you rearranged the order of the conditional statements at some point? It's a good idea in general to use comments sparingly, aiming instead to write code that explains itself, because it's really easy to leave misleading comments like this lying around

students/itsabhi/Lesson4/codekata.py
```diff
@@ -0,0 +1,70 @@
...
+            #Handling period, bang or question mark
+            if s_word.endswith("." or "!" or "?"):
+                punctuation = s_word[-1]
+            else:
+                punctuation = ""
```
robalford: Check the built in `str.replace()` or `str.translate()` methods for a way to simplify the punctuation removal and apply it to the entire text so you don't end up with repeated code here

students/itsabhi/Lesson4/codekata.py
```diff
@@ -0,0 +1,70 @@
...
+            word_dict[f_word + " " + s_word] = [t_word + punctuation]
```
robalford: I'd recommend using two word tuples rather than strings for the dict keys, this will simplify things when you generate the new text later

students/itsabhi/Lesson4/codekata.py
```diff
@@ -0,0 +1,70 @@
...
+        else:                                         #If key does not exist dict
```
robalford: Check out  the built in `dict.setdefault()` or `collections.defaultdict` from the Python Standard Library for a more concise way of dealing with this problem of checking for existing keys in the4 dict and either adding, or appending values

students/itsabhi/Lesson4/codekata.py
```diff
@@ -0,0 +1,70 @@
...
+            #Handling period, bang or question mark
+            if s_word.endswith("." or "!" or "?"):
+                punctuation = s_word[-1]
+            else:
+                punctuation = ""
+# Append value to existing value list
+            word_dict[f_word + " " + s_word] = word_dict[f_word + " " + s_word] + [t_word + punctuation]
```
robalford: you could use `list.append()` here

students/itsabhi/Lesson4/codekata.py
```diff
@@ -0,0 +1,70 @@
...
+
+        n += 1
+
+    print(word_dict)
+    return word_dict
+
+
+def generate(word_dict):
+    #This function generates the trigram output based on the dictionary input
+    pointer_key = "He was"
```
robalford: You can use the Python `random` module to select a random key per the assignment instructions

students/itsabhi/Lesson4/codekata.py
```diff
@@ -0,0 +1,70 @@
...
+    space_count = 1
+    print(pointer_key)
+
+    while space_count < 200:
+        try:
+            pointer_value = word_dict[pointer_key][0]
+            print(pointer_value)
+            pointer_key = pointer_key.split()[1] + " " + pointer_value
+        except KeyError:
+            print("Encountered a KeyError: " + pointer_key)
```
robalford: You should be able to simply skip any keys that aren't in the dict and continue the loop

students/itsabhi/Lesson4/codekata.py
```diff
@@ -0,0 +1,70 @@
...
+            pointer_value = word_dict[pointer_key[0:len(pointer_key)-1]][0]
+            print(pointer_value)
+            pointer_key = pointer_key[0:len(pointer_key)-1].split()[1] + " " + pointer_value
+        space_count += 1
+
```
robalford: This currently prints a single word per line to the console. Can you think of a way to build the new text into one long string to print? It might be helpful to build a list first and then use the `str.join()` method

robalford:
@itsabhister your trigram algorithm appears to produce the desired result, although it currently only prints a single word per line of the final text and there are a lot places where things could be a bit cleaner, more DRY or simplified. 

It's definitely a tricky assignment and you've done some good work on it here. So I'm going to approve the PR, but I would strongly suggest that you spend some time reworking your implementation per my comments. 

When you're ready, check out this really nice solution that Natasha came up with: 

https://github.com/natasha-aleksandrova/Self_Paced-Online/blob/master/solutions/lesson4/trigrams.py

students/itsabhi/Lesson4/codekata.py
itsabhister: Great suggestion, Rob. Yes, I was struggling with trying to avoid the single words per line. Didn't think of str.join(), but it totally makes sense now.

PR #384, opened 2018-04-03
Submit hw4 ( mailroom part2,  trigram )

mayc4t:


mayc4t:
Hello Natasha, 

Thanks a lot for all the comments you gave. Especially when you recommend what to do to make it better.

One question. I didn’t see the grade on trigram program, does that mean a different instructor will grade it separately?

Thanks,
Lily

PR #385, opened 2018-04-03
Part 3 of Lesson 02

UWTestEngr:
Lesson 02 should be complete.

### DIFF
--- a/students/Matt_Hudgins/Lesson02/series.py
+++ b/students/Matt_Hudgins/Lesson02/series.py
@@ -1,0 +1,73 @@
+'''
+    File Name: fizzbuzz.py
+    Author: Matt Hudgins
+    Date created: 4/2/2018
+    Date last modified: 4/2/2018
+    Python Version 3.6.4
+'''
+
+
+def fibonacci(n):
+    """This function calculates the nth number of the Fibonacci
+    Series"""
+    if n == 1:
+        return 1
+    elif n == 2:
+        return 1
+    elif n < 0:
+        print("Invalid Input")
+        return 0
+    else:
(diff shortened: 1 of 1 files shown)


students/Matt_Hudgins/Lesson02/series.py
```diff
@@ -0,0 +1,73 @@
...
+    Date created: 4/2/2018
+    Date last modified: 4/2/2018
+    Python Version 3.6.4
+'''
+
+
+def fibonacci(n):
+    """This function calculates the nth number of the Fibonacci
+    Series"""
+    if n == 1:
```
robalford: This should produce the series starting with 0 index (like you did for Lucas below) so that `fib(0) == 0, fib(1) == 1` etc.

students/Matt_Hudgins/Lesson02/series.py
```diff
@@ -0,0 +1,73 @@
...
+        return lucas(n - 2) + lucas(n - 1)
+
+def sum_series(n, x = 0, y = 0):
+    """Depending on the input parameters this function will either
+    calculate the nth number of the Fibonacci Series or the Lucas
+    numbers"""
+    if x == 0 or y == 0:
+        return fibonacci(n)
+    else:
+        return lucas(n)
```
robalford: The goal here is to create a series that defaults to fibonacci but can take any two numbers as the
first two values to produce a new series of numbers. So your function should work in such a way that the Lucas series would be produced by calling with `x=2` and `y=1` as the keyword arguments, so that `sum_series(7, 2, 1) == 29`. Other series could then be produced by substituting new values for the optional keyword arguments

PR #387, opened 2018-04-04
Committing final versions of break_me.py, Fizz_Buzz.py, and series.py

willhwhite:
Committing final versions of break_me.py (lesson01), Fizz_Buzz.py (lesson02), and series.py (lesson02)


students/WWhite/lesson02/Fizz_Buzz.py
```diff
@@ -0,0 +1,17 @@
+# -------------------------------------#
```
natasha-aleksandrova: note that convention is to use lower case for python filenames

PR #390, opened 2018-04-04
Adding last two lesson 2 assignments

TinaBrownfield:
Adding last two lesson assignments from lesson 2.

### DIFF
--- a/students/TinaB/lessonTwo_TB/FizzBuzz.py
+++ b/students/TinaB/lessonTwo_TB/FizzBuzz.py
@@ -1,0 +1,41 @@
+# Goal:
+# Write a program that prints the numbers from 1 to 100 inclusive.
+# But for multiples of three print “Fizz” instead of the number.
+# For the multiples of five print “Buzz” instead of the number.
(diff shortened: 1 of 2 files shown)


students/TinaB/lessonTwo_TB/FizzBuzz.py
```diff
@@ -0,0 +1,41 @@
...
+            counter += 1
+        elif counter % 3 == 0:
+            print("Fizz")
+            counter += 1
+        elif counter % 5 == 0:
+            print("Buzz")
+            counter += 1
+        else:
+            print(counter)
+            counter += 1
```
natasha-aleksandrova: you could have the counter incremented only once outside of if/else block

however, I would recommend `range` function here which avoids relying on the counter var

natasha-aleksandrova:
@TinaBrownfield make sure to submit your PR link on the course website so you can get the grade.

You will want to paste this link: https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/390

the number at the end is the unique ID for your Pull Request

PR #392, opened 2018-04-05
Finished mailroom.py changes

hortw1991:


students/thorn/lesson04/trigram.py
```diff
@@ -34,10 +34,13 @@ def read_text(intext):
 def create_trigrams_try_two(text_list):
     trigrams = {}
     # Add 3rd word to trigram pairs.  Create key entry if not already in dict.
+    # TODO: learn about default dicts :(
```
natasha-aleksandrova: `defaultdict` is awesome!

here is a quick demonstration:
```
from collections import defaultdict

dirs = defaultdict(list)
dirs["directory1"].append("file1")
dirs["directory1"].append("file2")
print(dirs)
print(dict(dirs))
```

the only difference is how you declare your dictionary and then you just use without worrying about getting a KeyError on append with a new key for example

students/thorn/lesson04/trigram.py
```diff
@@ -34,10 +34,13 @@ def read_text(intext):
...
     for i in range(len(text_list)-2):
         try:
             trigrams[(text_list[i], text_list[i+1])].append(text_list[i+2])
-        except:
+        except KeyError:
+            trigrams[(text_list[i], text_list[i+1])] = (text_list[i+2])
+        except AttributeError:
```
natasha-aleksandrova: you can use `except (KeyError, AttirubuteError):` if you have the same logic for both exceptions

students/thorn/lesson04/mailroom.py
```diff
@@ -97,12 +72,27 @@ def create_report():
...
+    # List is sorted by donor amounts desc.  Create preset list to pass in 
+    # vars once the list is sorted.
+    line_in = "{:<26}| ${:>14,.2f}|{:>15}| ${:>13,.2f}\n"
+    lines = []
+    # Append relevant information to the list to be sorted.
+    # List comprehension for fun
+    lines = [[donor for donor, donations in donors.items()],
+             [sum(donations) for donor, donations in donors.items()],
+             [len(donations) for donor, donations in donor.items()],
+             [(sum(donations)/len(donations)) for donor, donations in donor.items()]]
```
natasha-aleksandrova: a regular (single) loop would be better here because you are iterating over `donor` 4 times when it could be just once

PR #393, opened 2018-04-05
Lesson 06 Complete

stuartkershaw:


students/stuartkershaw/lesson06/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+def apply_selection(selection):
+    arg_dict = {
+        '1': send_thank_you,
+        '2': show_donor_table,
+        '3': generate_letters,
+        '4': quit
+    }
+    try:
+        if not arg_dict.get(selection):
+            raise KeyError
```
robalford: Raising `KeyError` is not necessary here, since it will be raised by Python if `selection` is not a key in the dict in the call to `arg_dict.get(selection)()`

students/stuartkershaw/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,61 @@
...
+from mailroom import list_donors
+from mailroom import show_donor_table
+from mailroom import generate_letters
+
+from os import listdir
+from os.path import isfile, join
+
+
+def test_donors():
+    assert isinstance(donors, dict) is True
```
robalford: This could just be `assert isinstance(donors, dict)` without the `is True`, but this test doesn't seem necessary to me since `donors` should always be a `dict`

students/stuartkershaw/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,61 @@
...
+    captured = capsys.readouterr()
+    assert captured.out == 'Donor Name          Total Given    Num Gifts'\
+        '      Average Gift   \n'\
+        '_________________________________________________________________\n'\
+        'Anne Hathaway        2              10250          5125.0         \n'\
+        'Geronimo Jackson     3              1800           600.0          \n'\
+        'Dingo Dogson         3              550            183.33         \n'\
+        'Leo Kottke           2              200            100.0          \n'\
+        'Flask McCreole       1              1000           1000.0         \n'\
+        'Piglet Norquist      3              65             21.67          \n'
```
robalford: You don't need to hard code out the entire output here. I would just test the header and one of the rows, with something like:
```
assert 'Piglet Norquist      3              65             21.67          \n' in captured.out
```
It can be tricky to know where to draw the line in some cases, but generally aim for testing as small a part of the generated/dynamic output as possible to still have a meaningful test.

robalford:
@stuartkershaw nice work here. Your mailroom code is nicely factored and very clean and readable overall, and you've off to a good start with testing your program. You might consider factoring out some of the donor lookup and creation code into their own small functions so that you could more easily test those aspects of your program, but that could be incorporated into the next version of mailroom. This PR is ready to merge.

students/stuartkershaw/lesson06/mailroom.py
natasha-aleksandrova: to clarify, `arg_dict[selection]` would correctly raise the `KeyError`

PR #394, opened 2018-04-05
Dustin_L - Lesson08 - Sparse array

rdlaner:
Lesson08 completed. Please provide any feedback on the sparse array implementation as well as the unit tests. Any areas of improvement or suggestions on how it could be more pythonic, efficient, etc. would be greatly appreciated. Thanks!


students/Dustin_L/lesson08/sparse_array.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python3
+"""Sparse Array Module"""
+
+
+class SparseArray():
+    """Sparse Array class"""
+    def __init__(self, seq):
+        self.num_elems = 0
+        self.elements = {}
```
natasha-aleksandrova: you don't need these two lines since you are redefining them below.

students/Dustin_L/lesson08/sparse_array.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python3
+"""Sparse Array Module"""
+
+
+class SparseArray():
```
natasha-aleksandrova: no need for parens

students/Dustin_L/lesson08/sparse_array.py
```diff
@@ -0,0 +1,120 @@
...
+    """Sparse Array class"""
+    def __init__(self, seq):
+        self.num_elems = 0
+        self.elements = {}
+
+        self.contains_zero = True if 0 in seq else False
+        self.elements = {i: arg for i, arg in enumerate(seq) if arg}
```
natasha-aleksandrova: since the main point of sparse array is not store zeros for efficiency, your dictionary should exclude zero values

rdlaner: Hi Natasha,
Perhaps I am missing something, but I am quite sure that my implementation does not store zeros in the self.elements. The conditional "if arg" in the dictionary comprehension should prevent any zeros from being entered as a value in the self.elements dictionary.
The "contains_zero" attribute is primarily for use in the "__contains__" magic method.

I'm pretty sure this works, please let me know if I am missing something.

natasha-aleksandrova: my apologies!! I totally missed the `if arg` part! I will take another look.

students/Dustin_L/lesson08/sparse_array.py
```diff
@@ -0,0 +1,120 @@
...
+                if self.elements[i] != other[i]:
+                    return False
+            elif other[i] != 0:
+                return False
+
+        return True
+
+    def __getitem__(self, index):
+        if isinstance(index, slice):
+            strt = index.start if index.start is not None else 0
```
natasha-aleksandrova: another shorter way `start = index.start or 0`

students/Dustin_L/lesson08/sparse_array.py
```diff
@@ -0,0 +1,120 @@
...
+            if i in self.elements.keys():
+                yield self.elements[i]
+            else:
+                yield 0
+
+    def __len__(self):
+        return self.num_elems
+
+    def __lt__(self, seq):
+        return bool(self.num_elems < len(seq))
```
natasha-aleksandrova: no need to cast to bool here

students/Dustin_L/lesson08/sparse_array.py
```diff
@@ -0,0 +1,120 @@
...
+        else:
+            self.elements[index] = value
+
+    def __str__(self):
+        elems = []
+        for i in range(self.num_elems):
+            if i in self.elements.keys():
+                elems.append(str(self.elements[i]))
+            else:
+                elems.append('0')
```
natasha-aleksandrova: you could do `elems = [str(self.elements.get(i, 0)) for i in range(self.num_elems])`

students/Dustin_L/lesson08/sparse_array.py
```diff
@@ -0,0 +1,120 @@
...
+            elif index in self.elements.keys():
+                return self.elements[index]
+            return 0
+
+    def __iter__(self):
+        for i in range(self.num_elems):
+            if i in self.elements.keys():
+                yield self.elements[i]
+            else:
+                yield 0
```
natasha-aleksandrova:
```
for i in range(self.num_elems):
    yield self.elements.get(i, 0)
```

students/Dustin_L/lesson08/sparse_array.py
```diff
@@ -0,0 +1,120 @@
...
+            strt = index.start if index.start is not None else 0
+            stop = index.stop if index.stop is not None else self.num_elems
+            step = index.step if index.step is not None else 1
+
+            ret_lst = []
+            for i in range(strt, stop, step):
+                if i in self.elements.keys():
+                    ret_lst.append(self.elements[i])
+                else:
+                    ret_lst.append(0)
```
natasha-aleksandrova:
```
ret_list = [self.elements.get(i, 0) for i in range(strt, stop, step)]
```

natasha-aleksandrova:
oh and 👍 on using git branches!

students/Dustin_L/lesson08/sparse_array.py
rdlaner: Nice one, thanks!

students/Dustin_L/lesson08/sparse_array.py
rdlaner: Nice, I forgot about default values with get(); this is a good one to keep handy.

PR #396, opened 2018-04-06
adding mailroom part 2

iamAndrewKim:


students/AndrewKim/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,49 @@
+
+donor_history = {'Andrew Kim': [5.0, 3.0], 'Jamie Park': [4.0], 'Tim Duncan': [3.0], 'Billy Yan': [2.0]}
+
+def thank_you():
```
robalford: Use 2 blank lines between module level functions for Pep8

students/AndrewKim/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,49 @@
...
+    #prompt for a Full Name
```
robalford: Add a space after the `#` in block comments for Pep8

students/AndrewKim/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,49 @@
...
+    name = input("Type the donor's full name: ")
+    #if user types 'list'
+    while name == "list":
+        for a in donor_history,keys():
```
robalford: You've got a comma here instead of a period in the call to `keys()` which causes the program to crash. But this can just be `for a in donor_history:` because the `in` operator checks for key membership in dict

students/AndrewKim/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,49 @@
...
+            print(a)
+        name = input("Type the donor's full name: ")
+
+    #if the name is not in the list
+    if name not in donor_history.keys():
```
robalford: Don't need the call to `keys()` here

students/AndrewKim/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,49 @@
...
+                    4 - Quit\n''')
+    choice = int(choice)
+    if choice == 1:
+        thank_you()
+    elif choice == 2:
+        report()
+    elif choice == 3:
+        send_all()
+    elif choice == 4:
+        break
```
robalford: Can you rework this section to use a dict to switch between the users selections as described in the assignment instructions? See [Using a Dictionary to switch](https://uwpce-pythoncert.github.io/PythonCertDevel/modules/DictionaryAsSwitch.html#dict-as-switch) for what this means.

PR #400, opened 2018-04-07
Mailroom Part 2 submission

mmsunday1:


students/M_Sunday/lesson04/dict and set lab.py
```diff
@@ -0,0 +1,80 @@
...
+# -------------------------
+
+# Dictionaries 2
+
+data2 = {'name': 'Chris', 'city': 'Seattle', 'cake': 'Chocolate'}
+
+data_cop = data2.copy()
+
+for i in data_cop.keys():
+    # new_value = 0
```
robalford: You can use `dict.items()` to loop through keys and values, e.g.
```
for k, v in data_cop.items():
```

students/M_Sunday/lesson04/Mailroom_Part_2.py
```diff
@@ -0,0 +1,109 @@
...
+
+def report():
+    summary_data = calc_data()
+    list_details(summary_data)
+
+
+def send_all_letters():
+    summary_data = calc_data()
+    for item in summary_data:
+        filename_format = "{}_{}.txt".format(item[0].split()[0], item[0].split()[1])
```
robalford: This call to `split()` will raise an IndexError if a user only enters their first name. You might look at using `str.replace()` as an alternative to dealing with spaces in the file names

students/M_Sunday/lesson04/Mailroom_Part_2.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+if __name__ == "__main__":
+
+    menu_map = {'1': thank_you, '2': report, '3': send_all_letters, '4': 'exit'}
+
+    while True:
+        menu_sel = menu_display()
+        if menu_map[menu_sel] == 'exit':
+            break
```
robalford: You might consider including an `exit` function that exits the system so you don't need the conditional here

students/M_Sunday/lesson04/dict and set lab.py
```diff
@@ -0,0 +1,80 @@
...
+# -------------------------
+
+# Sets 1
+
+s2 = {4, 12, 16, 18}
+
+s3 = {3, 6, 9}
+
+s4 = {4, 12, 16}
+
```
robalford: Can you think of a way to build these sets dynamically?

PR #405, opened 2018-04-07
Dustin_L - Lesson08 - Circle

rdlaner:
Circle lesson completed. Please let me know any feedback you may have. Thanks!


rdlaner:
Also, is there an established way for ordering the methods in a class? Alphabetical, grouped, magic methods first or last?

natasha-aleksandrova:
I don't believe there is a pep8 for ordering but in general magic methods first then public methods. you may also try to order by reading order

students/Dustin_L/lesson08/circle.py
```diff
@@ -0,0 +1,122 @@
...
+"""Circle Module"""
+
+import math
+
+
+class Circle:
+    """Circle Class"""
+    def __init__(self, radius):
+        if radius < 0:
+            raise ValueError
```
natasha-aleksandrova: good idea to supply error message for user of the class 
`raise ValueError("value has to be greater than 0")`

PR #407, opened 2018-04-08
Lesson 7 - This was a fun lesson :)

ghassan24601:


students/ghassan/lesson07/html_render.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __init__(self, content=None, **kwargs):
+        self.content = [content] if content else []
+        self.html_data = kwargs
+
+    def append(self, data):
+        self.content.append(data)
+
+    def render(self, file_out, cur_indent=''):
+        file_out.write('{}{}{}'.format(cur_indent, self.left, self.tag))  # noqa: E501
```
robalford: Is the `noqa` comment here for line length? Note that the Pep8 79 char per line limit is a fairly strict recommendation that is often not followed in industry. From the Pep8 docs:

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively > increasing the maximum length to 99 characters), provided that comments and docstrings are still wrapped at 72 characters.

> The Python standard library is conservative and requires limiting lines to 79 characters (and docstrings/comments to 72).

I have my linter set to somewhere around 99 chars per line. You can also add line breaks at opening function parens to keep under line limit and still adhere to Pep8, like this:
```
file_out.write('{}{}{}'.format(
            cur_indent, self.left, self.tag))
```

students/ghassan/lesson07/html_render.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class SefClosingTag(Element):
+    def render(self, file_out, cur_indent=''):
+            file_out.write('{}{}{}'.format(cur_indent, self.left, self.tag))  # noqa: E501
+            for k, v in self.html_data.items():
+                file_out.write(' {}="{}"'.format(k, v))
+            file_out.write('/{}\n'.format(self.right))
+            if self.content:
+                raise TypeError
```
robalford: Looks like you need to add the `Hr` and `Br` as sub-classes of `SelfClosingTag`

robalford:
@ghassan24601 - sparse array didn't look complete enough to review, but we can merge this for now and you can resubmit it with the next assignment if you'd like us to review it.

PR #411, opened 2018-04-08
Lesson 04 - mailroom.py

prgrover:


students/prgrover/lesson04/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+# Email template for Thank You message
+email = """Dear {:s},
+
+          Thank you for your very kind donation of ${:.2f}.
+          
+          It will be put to very good use.
+                         
+                Sincerely,
+                    -Hogwarts School of Witchcraft and Wizardry
+        """
```
robalford: Checkout `text wrap.dedent` from the Python standard library for a way to left align the output string while maintaining your current indent level in the code here (so you won't end up with the extra indentation in your generated letters)

students/prgrover/lesson04/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+        if donation_amount == "e":
+            return
+        elif donor_name in donor_dict:
+            donation_amount = float(donation_amount)
+            donor_dict[donor_name].append(donation_amount)
+            print(email.format(donor_name, donation_amount))
+        else:
+            donation_amount = float(donation_amount)
+            donor_dict[donor_name] = [float(donation_amount)]
+            print(email.format(donor_name, donation_amount))
```
robalford: You could eliminate some repetition here moving the shared functionality for new and existing donors outside of these conditional statements (float conversion and print statements)

students/prgrover/lesson04/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+        return
+
+    if donor_name == "list":
+        for donor in donor_dict.keys():
```
robalford: The `in` operator checks for key membership in `dict` so no need to call `keys()` here

students/prgrover/lesson04/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+
+"""
+The funtions below product the main selection menu
+"""
+
+def menu_selection(prompt, dispatch_dict):
+    while True:
+        response = input(prompt)
+        if dispatch_dict[response]() == "exit menu":
+            break
```
robalford: I think this is a bit confusing as written because the dispatched functions get called from within the `if` statement check for 'exit menu'. I'd recommend rewriting the `quit()` function so that it actually exits the system and more explicitly calling the user selected dispatch function here. 

I'd also suggest renaming your function to something other than `quit()` because that reassigns the Python built in `quit`  (see how GitHub highlights to bring this to your attention)

students/prgrover/lesson04/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+
+def print_report():
+    create_report()
```
robalford: What is the purpose of wrapping these functions like this?

PR #413, opened 2018-04-08
Lesson03

BrandonHenson:
I completed the lesson 3 assignments and checked them online for Pep8 format.

### DIFF
--- a/students/BrandonHenson/Lesson3/L3_Slicing.py
+++ b/students/BrandonHenson/Lesson3/L3_Slicing.py
@@ -1,0 +1,46 @@
+# Brandon Henson
+# Lesson 3
+# Slicing
+# 4/4/18
+
+
+# with the first and last items exchanged
+def exchange_first_last(seq):
+    return(seq[-1:] + seq[1: -1] + seq[:1])
+
+
+# with every other item removed
+def every_other_removed(seq):
+    return (seq[::2])
+
+
+# the first 4 and the last 4 items removed, and then every other item
+def first_last_removed_ev_other(seq):
+    return (seq[4:-4:2])
+
+
+# with the elements reversed (just with slicing)
+def elements_reversed(seq):
+    return (seq[-1::-1])
+
+
+# the middle third, then last third, then the first third in the new order
+def mid_last_first(seq):
+    sections = len(seq) // 3
+    firstsec = seq[:sections]
+    middlesec = seq[sections:sections]
+    lastsec = seq[sections:]
+    return(middlesec + lastsec + firstsec)
+
+# Tests
+a_string = "this is a string"
+a_tuple = (2, 54, 13, 12, 5, 32)
+test_list = (1, 2, 3, 4, 5, 6, 7, 8, 9)
+test_list2 = (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15)
+assert exchange_first_last(a_string) == "ghis is a strint"
+assert exchange_first_last(a_tuple) == (32, 54, 13, 12, 5, 2)
+# Write a test or two like that for each of the above functions.
+assert mid_last_first(test_list) == (4, 5, 6, 7, 8, 9, 1, 2, 3)
+assert elements_reversed(test_list) == (9, 8, 7, 6, 5, 4, 3, 2, 1)
+assert first_last_removed_ev_other(test_list2) == (5, 7, 9, 11)
+assert every_other_removed(test_list) == (1, 3, 5, 7, 9)
--- a/students/BrandonHenson/Lesson3/list_lab.py
+++ b/students/BrandonHenson/Lesson3/list_lab.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+# list_lab
+# Brandon Henson
+# 4/6/18
+
+
+def series_one():
+    print("__________________SERIES 1_______________")
+    # Create a list that contains “Apples”, “Pears”, “Oranges” and “Peaches”
+    list1 = ["Apples", "Pears", "Oranges", "Peaches"]
+    # Display the list
+    print(list1)
+    # Ask the user for another fruit
+    newfruit = input("Please enter another fruit \n")
+    # add it to the end of the list
+    list1.append(newfruit)
+    # Display the list
+    print(list1)
+    # Ask the user for a number
+    num1 = input("Please enter a number\n ")
+    num2 = int(num1) - 1
+    # display the number and fruit back to the user
+    print(int(num1), list1[num2])
+    # Add another fruit to the beginning of the list using “+” and display
+    addafruit = input("What fruit do you want to add to the beginning ? \n")
+    list2 = [addafruit] + list1
+    print(list2)
+    # Add another fruit to the beginning of the list using insert() and display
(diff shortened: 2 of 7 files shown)


students/BrandonHenson/Lesson3/L3_Slicing.py
```diff
@@ -0,0 +1,46 @@
+# Brandon Henson
+# Lesson 3
+# Slicing
+# 4/4/18
+
+
+# with the first and last items exchanged
+def exchange_first_last(seq):
+    return(seq[-1:] + seq[1: -1] + seq[:1])
```
robalford: The parentheses are not necessary here for the return statement. Same for all functions below.

students/BrandonHenson/Lesson3/L3_Slicing.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+# the first 4 and the last 4 items removed, and then every other item
+def first_last_removed_ev_other(seq):
+    return (seq[4:-4:2])
+
+
+# with the elements reversed (just with slicing)
+def elements_reversed(seq):
+    return (seq[-1::-1])
```
robalford: This can be `return seq[::-1]`

students/BrandonHenson/Lesson3/list_lab.py
```diff
@@ -0,0 +1,91 @@
...
+            list1.remove(i)
+        else:
+            continue
+    print(list1)
+
+
+def series_three():
+    print("__________________SERIES 3_______________")
+    list1 = ["Apples", "Pears", "Oranges", "Peaches"]
+    print (list1)
```
robalford: Remove whitespace between `print` and `(` here for Pep8

students/BrandonHenson/Lesson3/list_lab.py
```diff
@@ -0,0 +1,91 @@
...
+    list2 = []
+    for i in (list1):
```
robalford: The parens are not necessary here around `list1`. Also note that if you use slicing syntax to loop through a copy of your list like this:
```
for i in list1[:]:
```
then you can call `list1.remove()` in your loop to remove the items from the original list.

students/BrandonHenson/Lesson3/mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+# Lesson 03 mailroom.py
+
+donor_history = [
+            ['Brandon Henson', 1005.49, 3116.72, 5200],
+            ['Alicia Henson', 21.47, 1500],
+            ['Michael Green', 2400.54],
+            ['Brandon Henson Jr', 355.42, 579.31],
+            ['Kaiya Henson', 636.9, 850.13, 125.23]
+    ]
+name_list = []
```
robalford: Instead of building a second list here for donor lookups you might want to define a function that takes a name and returns the donor/donation list for that donor if it exists.

students/BrandonHenson/Lesson3/mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+for i in donor_history:
+    q = str(i[0])
```
robalford: No need for the call to `str` here. In general it's a good idea to use meaningful variable names, for everything except for counter/index variable and things like that. Since you're going to the trouble of reassigning this, I'd use `name = i[0]` or else you could just do `name_list.append(i[0])`

students/BrandonHenson/Lesson3/mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+        exit()
+    else:
+        print("Enter 1, 2, or 3")
+        main_menu()
+
+# If the user (you) selects ‘Send a Thank You’, prompt for a Full Name
+def menu_1():
+    fullname = input("Enter a full name\n")
+# If the user types ‘list’, show them a list of the donor names and re-prompt
+    if fullname.lower() == 'list':
```
robalford: You should probably provide the 'list' option to the user in your prompt

students/BrandonHenson/Lesson3/mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+        print(name_list)
+        menu_1()
+    elif fullname in name_list:
+        nameindex = (name_list.index(fullname))
```
robalford: No need for parens here

students/BrandonHenson/Lesson3/mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+        amount = float(input("Donation amount? \n"))
+        donor_history[nameindex].append(amount)
+        print("Thank you {} for your donation of ${}".format(fullname, amount))
+    elif fullname not in name_list:
+        new_profile = []
+        new_profile.append(fullname)
+        name_list.append(fullname)
+        amount = float(input("Donation amount? \n"))
+        new_profile.append(amount)
```
robalford: You could define the list as `new_profile = [full_name, amount]` here

students/BrandonHenson/Lesson3/mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+        donor_history.append(new_profile)
+        nameindex = (name_list.index(fullname))
```
robalford: You don't appear to be using this variable anywhere

students/BrandonHenson/Lesson3/mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+        print("Thank you {} for your donation of ${}".format(fullname, amount))
+    print()
```
robalford: You can add multiple `\n` chars to your string instead of the empty calls to print

students/BrandonHenson/Lesson3/strformat_lab.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def task_three(tuple1):
+    print("__________Task Three__________")
+    # Rewrite to take an arbitrary number of values.
+    # "the 3 numbers are: {:d}, {:d}, {:d}".format(1,2,3)
+    seq = []
+    for i in tuple1:
+        seq.append(i)
+    l = len(seq)
```
robalford: Same comment as above on variable names, instead of assigning to `l`, I'd just use the call to `len(seq)` in the code below.

students/BrandonHenson/Lesson3/strformat_lab.py
```diff
@@ -0,0 +1,66 @@
...
+    print("{3:02d} {4} {2} {0:02d} {1}".format(*five_numbers))
+
+
+def task_five():
+    print("__________Task Five__________")
+    list1 = ['oranges', 1.3, 'lemons', 1.1]
+    # Write an f-string that will display:
+    # The weight of an orange is 1.3 and the weight of a lemon is 1.1
+    print(f'The weight of an {list1[0] [:-1]} is {list1[1]} and the weight\
+    of a {list1[2] [:-1]} is {list1[3]}')
```
robalford: You can use this syntax to get rid of the extra white space in your string output:
```
print(f'The weight of an {list1[0] [:-1]} is {list1[1]} and the weight '
      f'of a {list1[2] [:-1]} is {list1[3]}')
```
Also note that it's ok to set your line length limit up to 99 chars for Pep8 despite the 79 char recommendation

PR #414, opened 2018-04-09
trigram exercise complete

mmsunday1:


students/M_Sunday/lesson04/kata_14.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/python
+import os
+import random
+
+
+def openfile(file):
+    text_ = []
+    f = open(file)
```
natasha-aleksandrova: `file` is a built-in, use alternate var names
also use `with` when working with files

students/M_Sunday/lesson04/kata_14.py
```diff
@@ -0,0 +1,54 @@
...
+    for i, word in enumerate(text_):
+        if "." in word:
+            text_[i] = word.replace(".", "")
+        if "," in word:
+            text_[i] = word.replace(",", "")
+        if "(" in word:
+            text_[i] = word.replace("(", "")
+        if ")" in word:
+            text_[i] = word.replace(")", "")
+    return text_
```
natasha-aleksandrova: why not use `text` variable?

students/M_Sunday/lesson04/kata_14.py
```diff
@@ -0,0 +1,54 @@
...
+        text_.append(word)
+    for i, word in enumerate(text_):
+        if "." in word:
+            text_[i] = word.replace(".", "")
+        if "," in word:
+            text_[i] = word.replace(",", "")
+        if "(" in word:
+            text_[i] = word.replace("(", "")
+        if ")" in word:
+            text_[i] = word.replace(")", "")
```
natasha-aleksandrova: there is a way to do this in bulk:
```
chars_to_replace = ("(", ")", ",")
for char in chars_to_replace:
    <thetext>.replace(char, "")
```

do the replacement on original text. no reason to loop over lines to create a list and then loop over list again just to replace.

students/M_Sunday/lesson04/kata_14.py
```diff
@@ -0,0 +1,54 @@
...
+    text = openfile(filename)
+    trigram_dict = create_dict(text)
+
+    first_word_in = random.randint(0, len(text)-1)
+    first_word = text[first_word_in]
+    second_word = text[first_word_in + 1]
+
+
+    def create_story(first_, second_, story_, dict_=trigram_dict):
+        if (first_, second_) in dict_:
```
natasha-aleksandrova: not sure why you're using the var names ending in `_`? also it is usually not a good idea to use mutable data types as default values.

students/M_Sunday/lesson04/kata_14.py
```diff
@@ -0,0 +1,54 @@
...
+            story_.append(dict_[first_, second_])
+            if len(story_) == 1:
+                create_story(second_, story_[0], story_)
+            else:
+                create_story(story_[-2], story_[-1], story_)
+        return story_
+
+    story__ = []
+    story = create_story(first_word, second_word, story__, trigram_dict)
```
natasha-aleksandrova: what is the reason for passing in `story_` to your function if it's just an empty list? 

this section is a bit hard to read with all the various versions of vars and inline function in the main block.

PR #419, opened 2018-04-10
except exercise complete with try/except blocks

mmsunday1:


students/M_Sunday/lesson05/Mailroom_Part_3.py
```diff
@@ -0,0 +1,118 @@
...
+def report():
+    summary_data = calc_data()
+    list_details(summary_data)
+
+
+def send_all_letters():
+    summary_data = calc_data()
+    for item in summary_data:
+        filename_format = "{}_{}.txt".format(item[0].split()[0], item[0].split()[1])
+        thank_you_text = open(filename_format, 'w+')
```
natasha-aleksandrova: prefer `with` statement when working with files

students/M_Sunday/lesson05/Mailroom_Part_3.py
```diff
@@ -0,0 +1,118 @@
...
+                donor_data.pop(donor_name)
+
+
+def calc_data():
+    summary_data = [[] for data in donor_data]
+    for i, item in enumerate(donor_data):
+        donations_quantity = len(donor_data[item])
+        donations_value = sum(donor_data[item])
+        donations_average = (float(donations_value / donations_quantity))
+        summary_data[i].extend([item, donations_quantity, donations_value, donations_average])
```
natasha-aleksandrova: not sure what the benefit is for pre-populating the list with empty list?

you can create `summary_data` inline, you also don't need enumerate

students/M_Sunday/lesson05/Mailroom_Part_3.py
```diff
@@ -0,0 +1,118 @@
...
+          "3: Send letters to everyone\n"
+          "4: Quit\n")
+    menu_sel_ = input("------> "
+                      "Select an option from the menu: ")
+    return menu_sel_
+
+
+def list_display():
+    print("\n---- List of all Donors ----")
+    [print("- {}".format(item)) for item in donor_data]
```
natasha-aleksandrova: it is not recommend to use comprehensions purely as a shortcut and to ignore returned data

students/M_Sunday/lesson05/Mailroom_Part_3.py
```diff
@@ -0,0 +1,118 @@
...
+    else:
+        if donor_name in donor_data:
+            add_donation(donor_name)
+        else:
+            donor_data[donor_name] = []
+            try:
+                add_donation(donor_name)
+            except ValueError:
+                print("Donation amount must be a number with no commas or currency symbols")
+                donor_data.pop(donor_name)
```
natasha-aleksandrova: seems like ValueError catching would be better with that specific code where it is converted to float
you code could also be simplified with `defaultdict`

PR #420, opened 2018-04-11
Added refactored mailroom, including more exceptions.

hortw1991:


students/thorn/lesson05/mailroom_with_exceptions.py
```diff
@@ -0,0 +1,153 @@
...
+    - Add that amount to donation history of selected user
+    - Print thank you note
+    - Returns to original prompt or quit to prompt
+    """
+    # Exception block test -> see if they entered an actual titled name
+    # Ignore the fact that it lowercases the name later
+    try:
+        name_request = input("Please enter a donor's full name, or 'list' for a full list of donors.  'Quit' or 'Return' valid commands. ")
+        name_request[0] = name_request.title[1]
+    except:
```
natasha-aleksandrova: it is better to catch specific exceptions

students/thorn/lesson05/mailroom_with_exceptions.py
```diff
@@ -0,0 +1,153 @@
...
+    # Append relevant information to the list to be sorted.
+    lines = [[] for _ in range(len(donors.keys()))]
+    i = 0
+    for name, donation in donors.items():
+        try:
+            lines[i].append(name)
+            lines[i].append(sum(donation))
+            lines[i].append(len(donation))
+            lines[i].append(sum(donation)/len(donation))
+            i += 1
```
natasha-aleksandrova: you shouldn't need to "prefill" the list and add the item to the list in one shot. 
```
lines.append([name, sum(donation), len(donation), sum(donation) / len(donation)])
```

students/thorn/lesson05/mailroom_with_exceptions.py
```diff
@@ -0,0 +1,153 @@
...
+            '1': send_thanks,
+            '2': create_report,
+            '3': send_letters,
+            '4': quitter
+        }
+        try:
+            user_choices[choice]()
+        except KeyError:
+            print("Invalid choice.")
+            continue
```
natasha-aleksandrova: `continue` isn't needed here

PR #422, opened 2018-04-11
String Formatting exercise for Lesson3

sbolsen:


students/sbolsen/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,47 @@
...
+
+#TASK_ONE
+a = (2, 123.4567, 10000, 12345.67)
+print("file_{:03d}: {:03.2f} {:0.2E} {:0.2E}".format(a[0], a[1], a[2], a[3]))
+
+#TASK_TWO
+
+a = (2, 123.4567, 10000, 12345.67)
+print('file_{first}: {second} {third} {fourth}'.format(first = "{:03d}".format(a[0]), second = "{:03.2f}".format(a[1]), third = "{:0.2E}".format(a[2]), fourth = "{:0.2E}".format(a[3])))
+
```
robalford: You can move the format specifications to the replacement fields in the main string here, so the whole thing would look like this:
```
print('file_{first:03d}: {second:03.2f} {third:0.2E} {fourth:0.2E}'.format(
    first=a[0], second=a[1], third=a[2], fourth=a[3]))
```
Also, note that I started a new line at the opening parenthesis for Python/Pep8 line length recommendation.

students/sbolsen/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,47 @@
...
+#TASK_THREE
+
+def formatter(tuple_in):
+    x = tuple_in
+    y = len(tuple_in)
```
robalford: It's a good idea in general to use meaningful variable names, and in this case, it might be more explicit to just pass `len(tuple_in)` directly into `format()` without reassigning

students/sbolsen/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,47 @@
...
+    element = "{:d}" * y
```
robalford: The `x` and `element` variable are assigned but not used in this function

PR #423, opened 2018-04-11
Lesson 7 Assignment: HTML Renderer

JeroenSweerts:


students/JeroenSweerts/Lesson7/html_render.py
```diff
@@ -0,0 +1,169 @@
+from io import StringIO
+
+class Element():
```
natasha-aleksandrova: parens are not necessary here

students/JeroenSweerts/Lesson7/html_render.py
```diff
@@ -0,0 +1,169 @@
...
+    indentation = 4
+    def __init__(self, content=None, **attributes):
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
+        self.attributes = attributes
+
+    def append(self, string=None):
+        '''Method that can add another string to the content.'''
```
natasha-aleksandrova: your docstrings should be using triple double quotes

students/JeroenSweerts/Lesson7/test_html_render.py
```diff
@@ -0,0 +1,167 @@
...
+import os.path
+import io
+import sys
+from io import StringIO
+
+
+class TestHTMLRender(unittest.TestCase):
+
+    def test_Element_render(self):
+        f = StringIO()
```
natasha-aleksandrova: you could use `setUp` method to create `StringIO` instance so that it can be available in each test

PR #427, opened 2018-04-12
Adding Lesson Three Mailroom and strformat_lab, updating previous labs

TinaBrownfield:
Updated first two labs based on comments, and adding mailroom and strformat_lab.

### DIFF
--- a/students/TinaB/lesson03/mailroom.py
+++ b/students/TinaB/lesson03/mailroom.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+"""
+Goal:
+The Program
+
+Script to pull list of donors and print a thank you letter.
+
+"""
+# list of donors
+donors_list = [
+    ['William Gates, III', 326892.24, 4],
+    ['Mark Zuckerberg', 65982.55, 3],
+    ['Jeff Bezos', 52636.27, 1],
+    ['Paul Allen', 877.33, 2],
+    ['Steven Hawking', 326892.24, 5],
+    ['Justin Timberlake', 999658.25, 3]]
+
+
+def send_thank_you():
+    """ function for sending thank you message"""
+    donors_list_copy = []
+    for donor in donors_list:
+        donors_list_copy.append(donor)
+
+    while True:
+        print()
+        print('Welcome to the Send A Thank You Menu:')
+        thankyou_input = input('\tEnter The Last Nanme of a donor \n'
+                               '\tType "list" for a list of donors, or \n'
+                               '\tType "quit" to quit\n'
+                               '---->  ')
+        # print(thankyou_input)
+        if thankyou_input.isalpha():
+            if thankyou_input.lower().startswith('q'):
+                break
+            elif thankyou_input.lower() == 'list':
+                print_donors_names()
+                thankyou_input = input('Please enter the name of the donor: ')
+                # print(thankyou_input)
+            # search through donors for matching name
+            for idx, donor in enumerate(donors_list_copy):
+                if thankyou_input.lower() in donor[0].lower():
+                    donor_check = input(
+                        "Is this the donor you are looking for:  {}?   Please enter yes or no:  ".format(donor[0]))
+                    if donor_check.lower().startswith('y'):
+                        donor_amount = float(
+                            input('Please enter a donation amount:  '))
+                        donors_list[idx][1] += donor_amount
+                        donors_list[idx][2] += 1
+                        print_thank_you(donor, donor_amount)
+                        break
+                #Sprint('{}  {}'.format(donor[0], thankyou_input.lower()))
+            else:
+                print("No donor matching that name is in the list. We will add {} to the donor list. ".format(
+                    thankyou_input))
+                donor_fname = input(
+                    "Please enter {}'s first name: ".format(thankyou_input))
+                donor_name = donor_fname + ' ' + thankyou_input
+                donor_amount = float(
+                    input('Please enter a donation amount:  '))
+                donors_list.append([donor_name, donor_amount, 1])
+                print()
+                print_thank_you(donor_name, donor_amount)
+
+
+def print_donors_names():
+    """ prints list of donors"""
+    print()
+    for donor in donors_list:
+        print(donor[0])
+    print()
+
+
+def print_thank_you(donor, amount):
+    """ prints thank you message"""
+    amounts = float(amount)
+    print('Dear {},\n '.format(donor))
+    print(
+        '\tThank you for your generous donation of ${:,.2f}\n'.format(amounts))
+    print('Sincerely, \n'
+          'ChickTech Donations Department')
+
+
+def print_main_menu():
+    """ prints main menu  """
+    print("Welcome to the Mailroom App")
+    print('\nOptions Menu:\n'
+          '\t1. Send a Thank You\n'
+          '\t2. Create a Report\n'
+          '\t3. Quit\n')
+
+
+def print_report():
+    """Print report to match example from assignment for donor list """
+    print()
+    title = ['Donor Name', '|  Total Given', '| Num Gifts',
+             '  | Average Gift']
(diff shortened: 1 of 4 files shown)


students/TinaB/lesson03/mailroom.py
```diff
@@ -0,0 +1,128 @@
...
+        # print(thankyou_input)
+        if thankyou_input.isalpha():
+            if thankyou_input.lower().startswith('q'):
+                break
+            elif thankyou_input.lower() == 'list':
+                print_donors_names()
+                thankyou_input = input('Please enter the name of the donor: ')
+                # print(thankyou_input)
+            # search through donors for matching name
+            for idx, donor in enumerate(donors_list_copy):
```
robalford: I can't see any reason to loop over a copy of the original list here so I don't think you need `donor_list_copy` and can just use the original `donor_list` throughout

students/TinaB/lesson03/mailroom.py
```diff
@@ -0,0 +1,128 @@
...
+                if thankyou_input.lower() in donor[0].lower():
```
robalford: Note that this look up would work for first or last name or any substring of the name. If you're going to ask the user for first and last separately, you should probably store those values separately in your data structure.

students/TinaB/lesson03/mailroom.py
```diff
@@ -0,0 +1,128 @@
...
+                    donor_check = input(
+                        "Is this the donor you are looking for:  {}?   Please enter yes or no:  ".format(donor[0]))
+                    if donor_check.lower().startswith('y'):
```
robalford: I'd check for 'yes' here since this will work for any word that begins with 'y'

students/TinaB/lesson03/mailroom.py
```diff
@@ -0,0 +1,128 @@
...
+                        donor_amount = float(
+                            input('Please enter a donation amount:  '))
+                        donors_list[idx][1] += donor_amount
+                        donors_list[idx][2] += 1
+                        print_thank_you(donor, donor_amount)
```
robalford: This should be `print_thank_you(donor[0], donor_amount)`

students/TinaB/lesson03/mailroom.py
```diff
@@ -0,0 +1,128 @@
...
+                        break
+                #Sprint('{}  {}'.format(donor[0], thankyou_input.lower()))
```
robalford: It's a good idea to remove commented code before committing to git. As long you're committing your work as you go along, you can always go back to a previous version if you want to.

students/TinaB/lesson03/mailroom.py
```diff
@@ -0,0 +1,128 @@
...
+    """ prints list of donors"""
+    print()
+    for donor in donors_list:
+        print(donor[0])
+    print()
+
+
+def print_thank_you(donor, amount):
+    """ prints thank you message"""
+    amounts = float(amount)
```
robalford: No need for the reassignment and float conversion here, you can just pass `amount` into `format()`

students/TinaB/lesson03/mailroom.py
```diff
@@ -0,0 +1,128 @@
...
+        print_main_menu()
+        selected_option = int(input('Enter Your Selection Here: '))
+        if selected_option == 1:
+            send_thank_you()
+        elif selected_option == 2:
+            print_report()
+        elif selected_option == 3:
+            print('The End - Clsoing Program')
+            break
+    else:
```
robalford: I think you can remove this entire `else` clause, it doesn't appear to be doing anything and it's causing the program to crash as written because the `continue` statements are outside of the original `while` loop

students/TinaB/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,99 @@
...
+So you need to find a string formatting operator that will “pad” the number with zeros for you.
+The second element is a floating point number. You should display it with 2 decimal places shown.
+The third value is an integer, but could be any number. You should display it in scientific notation, with 2 decimal places shown.
+The fourth value is a float with a lot of digits – display it in scientific notation with 3 significant figures.
+"""
+print("Task One:")
+#output to match
+print("file_002 :   123.46, 1.00e+04, 1.23e+04")
+#print("file_{:03d} :   {:06.2f}, {:.2E}, {:.2E}".format(2, 123.4567, 10000, 12345.67))
+print("file_{:0>03} :   {:06.2f}, {:.2E}, {:.2E}".format(
```
robalford: I think you want lower case `e`'s here

students/TinaB/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,99 @@
...
+"""
+Dynamically Building up Format Strings
+Rewrite: "the 3 numbers are: {:d}, {:d}, {:d}".format(1,2,3)
+to take an arbitrary number of values.
+ """
+print("Task Three:")
+
+def formatter(in_tuple):
+    """ function to print input tuple """
+    formatted_string = "the numbers are: "
```
robalford: This variable isn't being used in this function

students/TinaB/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,99 @@
...
+    form_string = "{:d}, " * (len(in_tuple)-1) + "{:d}"
```
robalford: This can be simplified:
```
form_string = "{:d}, " * len(in_tuple)
```

students/TinaB/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,99 @@
...
+Write an f-string that will display: The weight of an orange is 1.3 and the weight of a lemon is 1.1
+Now see if you can change the f-string so that it displays the names of the fruit in upper case, 
+and the weight 20% higher.
+"""
+print("Task Five:")
+element_list = ['oranges', 1.3, 'lemons', 1.1]
+# String to match:
+print("The weight of an orange is 1.3 and the weight of a lemon is 1.1")
+# part one match test string
+print(f"The weight of an {element_list[0][:-1]} is {element_list[1]} and the weight of a {element_list[2][:-1]} is {element_list[3]}")
```
robalford: Watch the line lengths on these strings, you can use the following syntax for Pep8:
```
print(f"The weight of an {element_list[0][:-1]} is {element_list[1]} and the weight of a {element_list[2][:-1]} is "
         f"{element_list[3]}")
```

students/TinaB/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,99 @@
...
+table_list = [('Carina', 27, 15, 156.25), ('Timothy', 2, 127.63),
+              ('Paul', 12, 1758.17), ('Jeffery', 65, 199), ('Ruben', 107, 165, 248.25),
+              ('Tristan', 85, 8, 156.92), ('Shannon', 44, 556.2), ('Elizabeth', 15, 166, 156.25),
+              ('Tom', 39, 825), ('Tricia', 93, 156), ('Allegra', 68, 962.30)]
+
+# print table header:
+print('{:<10}{:<6}{:>12}'.format('Name', 'Age', 'Cost'))
+# table
+for item in table_list:
+    print('{:<12s}{:<6}{:>10,.2f}'.format(*item))
```
robalford: The alignment on the second column looks a bit off here. Numbers are generally right-aligned in spreadsheets etc.

students/TinaB/lesson03/mailroom.py
TinaBrownfield: I ran into an issue that if the user put in two names (first and last) the loop wouldn't function correctly. So to fix that I set it up as search by last name with the addition of adding a first name when the donor wasn't found.

students/TinaB/lesson03/mailroom.py
TinaBrownfield: ok I will change that.  Thank you for the suggestion.

PR #429, opened 2018-04-13
Working except for <a>

hortw1991:
My  <a> seems to have some formatting issues that I was hoping to get some feedback on.  if I can't get it working by fixing something at the top level with the Element class, I will edit it with a render method specific to its instance.


students/thorn/lesson06/mailroom_6.py
```diff
@@ -0,0 +1,179 @@
...
+
+def get_donor_names():
+    """ Returns the donor keys - representing the donor names. """
+    return donors.keys()
+
+
+def get_donor_totals(donor):
+    """ Returns the total donation for a donor.  Not currently active. Except for testing. """
+    print(sum(donors[donor]))
+    return(sum(donors[donor]))
```
robalford: I would get rid of this unimplemented function and the test, since it's basically testing the built-n `sum()`

students/thorn/lesson06/mailroom_6.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def print_thanks(name_request, donation_amount):
+    message = f"""Dear {name_request},
+        Thank you for you generous donation of ${donation_amount:.2f}.
+        It will truly help the children.
+
+        Sincerely,
+        Donation Recievers"""
+    print(message)
```
robalford: You could return a value here to print, to make this testable

students/thorn/lesson06/mailroom_6.py
```diff
@@ -0,0 +1,179 @@
...
+    After name is used -> prompt for donation amount and turn into a number.
+    - Add that amount to donation history of selected user
+    - Print thank you note
+    - Returns to original prompt or quit to prompt
+    """
+    # Exception block test -> see if they entered an actual titled name
+    # Ignore the fact that it lowercases the name later
+    try:
+        name_request = input("Please enter a donor's full name, or 'list' for a full list of donors.  'Quit' or 'Return' valid commands. ")
+        name_request[0] = name_request.title[1]
```
robalford: It looks like there are a couple of potential bugs here. As it's written, this line will throw a TypeError every time because it's missing the parentheses in the call to `title()` so you're trying to set the index on the `title` method itself. However, the assignment to `name_request[0]` would still fail because strings are immutable in Python and don't support item assignment.

students/thorn/lesson06/mailroom_6.py
```diff
@@ -0,0 +1,179 @@
...
+    except:
```
robalford: It's always a good idea to hone in on a particular exception rather than use a bare `except` statement because bugs like the ones mentioned above can slip by silently. I don't think you really need the try/except block here though since the call `str.upper()` should work, even for values like '2' because the `input()` function will to try convert any user input to a string.

students/thorn/lesson06/mailroom_6.py
```diff
@@ -0,0 +1,179 @@
...
+    # List request - list of donors returned
+    # Exception added to make user format it properly.
+    if name_request.lower() == 'list':
+        # print('\n'.join(get_donor_names()))
+        donor_names = get_donor_names()
+        print('\n'.join(donor_names))
+        return
+
+    name_request = name_request.title()
+    donation_amount = float(input(("Please enter a donation amount for {}: ".format(name_request))))
```
robalford: This would be a good place for a try/except block to catch a ValueError for non-numerical input

students/thorn/lesson06/mailroom_6.py
```diff
@@ -0,0 +1,179 @@
...
+
+    for name, donation in donors.items():
+        try:
+            lines[i].append(name)
+            lines[i].append(sum(donation))
+            lines[i].append(len(donation))
+            lines[i].append(sum(donation)/len(donation))
+            i += 1
+        except IndexError:
+            continue
```
robalford: You can do this without the counter variable like this:
```
    lines = []
    for name, donation in donors.items():
        lines.append([name, sum(donation), len(donation), (sum(donation) / len(donation))])
```

students/thorn/lesson06/mailroom_6_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+class mailroom_6_unit_test(unittest.TestCase):
+    ''' Sets up test information. '''
+    donors = {
+    'Tom Horn': [599.23, 1000.00],
+    'Theo Hartwell': [0.01, 0.01, 0.1],
+    'Bailey Kimmitt': [8723.22, 27167.22, 91817.66],
+    'Paul Hubbell': [90012.32, 2312.24],
+    'David Beckham': [1817266.11, 123123.66, 111335.112]
+    }
```
robalford: You can use `unittest.TestCase.setUp` method for defining shared data and set up logic between test methods. See [this section](https://docs.python.org/3/library/unittest.html#organizing-test-code) of the unittest docs for how to organize test code.

students/thorn/lesson06/mailroom_6_test.py
```diff
@@ -0,0 +1,55 @@
...
+    def test_create_report_lines(self):
+        ''' Returns sorted list of the original donors dictionary that matches above.  Not a great test -> doesn't account for rounding in the final product. '''
+        self.assertEqual(mailroom_6.create_report_lines(), self.sorted_donors)
+
+    def test_create_report(self):
+        ''' Compares a line in the report to make sure they're formatted correctly. '''
+        becks_line = 'David Beckham             | $  2,051,724.88|              3| $   683,908.29'
+        report = mailroom_6.create_report().split('\n')
+        self.assertEqual(report[2], becks_line)
+
```
robalford: I'd recommend adding a test for the `send_letters()` function that checks that the files get created with the proper content

students/thorn/lesson07/html_render.py
```diff
@@ -0,0 +1,169 @@
...
+    ''' <head> tag '''
+    tag = 'head'
+
+
+class OneLineTag(Element):
+    ''' Overrides the render method to put everything on one line for certain tags.
+
+    Should be similar to how element acts but without new lines.
+    '''
+    def render(self, fileout, cur_ind=''):
```
robalford: You want to stick with the same parameter names in subclass methods, so this should be `file_out` with underscore

students/thorn/lesson07/html_render.py
```diff
@@ -0,0 +1,169 @@
...
+
+class Meta(SelfClosingTag):
+    ''' Meta Tag - self closing '''
+    tag = "meta"
+
+
+
+''' Not working quite right. '''
+
+class A(Element):
```
robalford: This should subclass from `OneLineTag`. You will then need to customize the `OneLineTag` render method for the proper formatting, and since you'll need to include the code for rendering attributes there as well, that might be a good place for some refactoring to avoid code duplication by moving the attribute rendering code into a separate method for use in the various `Element` subclasses.

students/thorn/lesson07/html_render_unit_tests.py
```diff
@@ -0,0 +1,91 @@
...
+#         test = hr.Html()
+#         e.render(hello)
+#         f = StringIO()
+#         test.render(f)
+#         lines = f.getvalue()
+#         lines.strip()
+#         print(lines)
+
+
+class Step3Tests(unittest.TestCase):
```
robalford: I would use a single class for all of your test methods in this file. And you can use `unittest.TestCase.setUp` method for shared data and set up logic between methods.

students/thorn/lesson07/html_render_unit_tests.py
```diff
@@ -0,0 +1,91 @@
...
+        test.render(f)
+        # Check for start and end with title
+        self.assertEqual(f.getvalue()[:7], "<title>")        
+        self.assertEqual(f.getvalue()[-9:], "</title>\n")
+
+
+class Step4Tests(unittest.TestCase):
+    def test_element_accepts_kwargs(self):
+        ''' Test to just make sure element accepts different # kwargs. '''
+        test = hr.Element("some text content", id="TheList", style="line-height:200%")
```
robalford: Looks like this needs a bit more work

PR #430, opened 2018-04-13
Adding mailroom.py for Lesson3

sbolsen:


students/sbolsen/lesson03/mailroom.py
```diff
@@ -0,0 +1,83 @@
...
+            if users_choice == '1':
+                thanked_user = thank_you()
+                user(thanked_user)
+            elif users_choice == '2':
+                create_report()
+                prompt_user()
+            elif users_choice == '3':
+                quit()
+            else:
+                prompt_user()
```
natasha-aleksandrova: this will matter in next part of the mailroom assignment, but basically it will be better to have redirection to a single function.

students/sbolsen/lesson03/mailroom.py
```diff
@@ -0,0 +1,83 @@
...
+
+def create_report():
+    title = ('Donor Name', 'Total Given', 'Num Gifts', 'Average Gift')
+    print('{:20} | {:>17} | {:>17} | {:>17}'.format(*title))
+    print('{:-<80}'.format(''))
+    for donor in donors:
+        a = sum(donor[1])
+        b = len(donor[1])
+        total = a // b
+        print('{:20} $ {:>17} $ {:>17} $ {:>17}'.format(donor[0], a, b, total))
```
natasha-aleksandrova: this report should be sorted

students/sbolsen/lesson03/mailroom.py
```diff
@@ -0,0 +1,83 @@
...
+    if user_exists:
+        pass
+    else:
+        donors.append([select_user, []])
+    update_donation = int(input('Please enter donation: '))
+    for donor in donors:
+        donations = False
+        for i in donor:
+            if not donations:
+                donations = True
```
natasha-aleksandrova: what is the significance of `donations` var?

students/sbolsen/lesson03/mailroom.py
```diff
@@ -0,0 +1,83 @@
...
+        select_user = input('Enter a full name: ')
+    return select_user
+
+
+def user(select_user):
+    user_exists = False
+    for item in donors:
+        for donor in item:
+            if select_user == donor:
+                user_exists = True
```
natasha-aleksandrova: you can just have `return` here to exit function, no need to have additional var

students/sbolsen/lesson03/mailroom.py
```diff
@@ -0,0 +1,83 @@
...
+            elif users_choice == '3':
+                quit()
+            else:
+                prompt_user()
+        except Exception as e:
+            print("Exception [{}]".format(e))
+
+
+if __name__ == "__main__":
+    main()
```
natasha-aleksandrova: there are some flow issues: run script then select 2 then 1 (nothing happens), 1 again.

PR #432, opened 2018-04-14
Circle Assignment

alexLaws:


students/alexLaws/lesson08/Circle.py
```diff
@@ -0,0 +1,53 @@
...
+        self.radius = diameter / 2
+
+    @property
+    def area(self):
+        return (self.radius ** 2) * pi
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        radius = diameter / 2
+        return Circle(radius)
```
natasha-aleksandrova: this should be `cls` since first argument is the class: `return cls(radius)`

students/alexLaws/lesson08/test_circle.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def test_alt_construct_diameter():
+    a = Circle.from_diameter(10)
+    assert a.diameter == 10
+
+
+def test_str_function():
+    a = Circle(5)
+    assert a.__str__() == "Cirle with Radius: {}".format(a.radius)
```
natasha-aleksandrova: use `assert str(a) == ".."` instead of magic methods

students/alexLaws/lesson08/test_circle.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def test_repr_function():
+    a = Circle(5)
+    assert a.__repr__() == "Circle(5)"
```
natasha-aleksandrova: use `assert repr(a) == " ... "` instead of magic methods

PR #433, opened 2018-04-14
Dustin_L - Lesson09 - Mailroom with classes

rdlaner:
Completed Lesson09. Any and all feedback welcome!

### DIFF
--- a/students/Dustin_L/lesson09/mailroom4.py
+++ b/students/Dustin_L/lesson09/mailroom4.py
@@ -60,16 +60,22 @@ class Donor:
 
 class DonorDatabase(defaultdict):
     """A database of Donors"""
-    THANK_YOU_FMT = ('\nDear {:s},\n'
-                     'Thank you for your generous donation of ${:.2f}.\n'
-                     '\t\tSincerely,\n'
-                     '\t\t  -Your conscience')
-
     def __init__(self, *donors):
         if not donors:
             donors = []
 
         super().__init__(Donor, {d.name: d for d in donors})
+        self.min_col_width = 12
+        self.def_pad = 5
+        self.col_sep = ' | '
+        self.col_1_name = 'Donor Name'
+        self.col_2_name = 'Total Given'
+        self.col_3_name = 'Num Gifts'
+        self.col_4_name = 'Average Gift'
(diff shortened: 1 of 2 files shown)


students/Dustin_L/lesson09/mailroom4.py
```diff
@@ -0,0 +1,302 @@
...
+    donor_db = DonorDatabase(*donors)
+
+    opt_dict = dict(zip(PROMPT_OPTS, (send_thank_you,
+                                      create_report,
+                                      send_letters,
+                                      quit_mailroom)))
+    choice = ''
+    while choice != PROMPT_OPTS[-1]:
+        choice = get_usr_input()
+        opt_dict.get(choice)(donor_db)
```
natasha-aleksandrova: should catch a `KeyError` here

students/Dustin_L/lesson09/mailroom4.py
```diff
@@ -0,0 +1,302 @@
...
+            donors = []
+
+        super().__init__(Donor, {d.name: d for d in donors})
+        self.min_col_width = 12
+        self.def_pad = 5
+        self.col_sep = ' | '
+        self.col_1_name = 'Donor Name'
+        self.col_2_name = 'Total Given'
+        self.col_3_name = 'Num Gifts'
+        self.col_4_name = 'Average Gift'
```
natasha-aleksandrova: this would be better as:
`self.cols = ['Donor Name', 'Total Given'.. ]`
then pass into format: `(*self.cols)`

students/Dustin_L/lesson09/test_mailroom4.py
```diff
@@ -0,0 +1,299 @@
...
+        self.assertTrue(self.donor.donations == self.test_dons)
+
+    def test_set_donations(self):
+        """Test setting donations attribute"""
+        with self.assertRaises(AttributeError):
+            self.donor.donations = [100, 20]
+
+    def test_get_num_donations(self):
+        """Test num_donations attribute"""
+        self.assertTrue(self.donor.num_donations == len(self.test_dons))
```
natasha-aleksandrova: `self.assertEquals`  I believe would give you a better error message on failure and also makes more sense here

PR #434, opened 2018-04-14
lesson 07 done

ZhuojingXie:


students/ZhuojingXie/lesson06/test_Mailroom4.py
```diff
@@ -0,0 +1,33 @@
...
+class TestMailRoom4(unittest.TestCase):
+
+
+    def test_list_name(self):
+
+        data = [{'name':'Andy','donation':[960,256,123.5,40]},{'name':'Bryce','donation':[30,45,27]}]
+
+
+
+        self.assertEqual(data[0], {'name':'Andy','donation':[960,256,123.5,40]})
```
natasha-aleksandrova: this does not look like it is testing the mailroom?

students/ZhuojingXie/lesson06/test_Mailroom4.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+
+
+    def send_letters_to_everyone(self):
+        data = [{'name':'Andy','donation':[960,256,123.5,40]},{'name':'Bryce','donation':[30,45,27]}]
+        mailroom4.letters_to_everyone()
+        assert os.path.isfile('Andy.txt')
```
natasha-aleksandrova: when using `unittest` module it is better to use their built in assert methods as they will give you more descriptive message (`self.assertTrue`)

natasha-aleksandrova: btw, this is an example of a good test for the mailroom

students/ZhuojingXie/lesson07/test_html_render.py
```diff
@@ -0,0 +1,44 @@
+import unittest
+import html_render as hr
+from io import StringIO
+
+class TestHtmlRender(unittest.TestCase):
+
+    def test_Element(self):
```
natasha-aleksandrova: this should still be lower case snake_case -> `def test_element`

natasha-aleksandrova:
@ZhuojingXie make sure that you have submissions for all of the assignments in this PR on the course website

PR #435, opened 2018-04-15
lesson08 - Circle class

ghassan24601:


students/ghassan/lesson08/test_circle.py
```diff
@@ -0,0 +1,60 @@
+from circle import Circle
+from math import pi
+
+
+def test_Circle():
```
natasha-aleksandrova: should be lower case

students/ghassan/lesson08/test_circle.py
```diff
@@ -0,0 +1,60 @@
...
+    c = Circle(1)
+    assert a < b
+    assert b > a
+    assert c == a
+
+
+def test_sort():
+    list_unsorted = [Circle(6), Circle(7), Circle(8), Circle(4), Circle(
+        0), Circle(2), Circle(3), Circle(5), Circle(9), Circle(1)]
+    assert list_unsorted.sort() == list_unsorted.sort()
```
natasha-aleksandrova: hm this tests that the same thing equals to itself?

ghassan24601: good point! :)

PR #436, opened 2018-04-15
Add test_mailroom.py

22e88:
Working state, but one issue still unfixed, see comments in line 61 to 65 in test_mailroom.py. I leave it like this for the moment, as for time reasons i want to move on to the next topics now. You can find the full history of my experiments in my branch LESSON06...


students/elmar_m/lesson06/mailroom.py
```diff
@@ -0,0 +1,213 @@
...
+        ARGS:
+    p:  prompt which is shown to the user
+    d:  the dispatcher dictionary which holds the menu options
+    '''
+    try:
+        while True:
+            response = input(p)
+            if d[response]() == 'exiting':
+                break
+    except KeyError:
```
natasha-aleksandrova: your `KeyError` should be around line 204 and no need to call `menu(p, d)` as you are in a while True loop

students/elmar_m/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+        mocked / patched at this point and returns a faked return value instead of 
+        being executed / tested here once again in real. 
+        This leads to misleading test results at this point!
+        Possible workaround: put test_add_amount() before mytestclass.test_add().
+        '''
+        assert m.add_amount(test_donor) is True
+
+
+def test_report():
+    assert m.report() is True
```
natasha-aleksandrova: this doesn't really test that report was created correctly, this only tests that the function ran and there was no exception (but really your test would fail anyway if there was an exception).

your report function could return a report string, and your test could assert that it was generated correctly.

students/elmar_m/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+
+def test_mail():
+    assert m.mail() is True
+
+
+def test_list_donors():
+    assert m.list_donors() is True
+
+
+class mytestclass(TestCase):
```
natasha-aleksandrova: this should be upper case camel case (`MyTestClass`)

students/elmar_m/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+            with patch('sys.stdout', new=StringIO()) as fake_output:
+                m.add_amount = MagicMock(return_value=True)
+                m.add()                
+                self.assertEqual(fake_output.getvalue(), '>> berta already in list\n')
+
+        with patch('builtins.input', side_effect=['nobody']) as fake_input:
+            with patch('sys.stdout', new=StringIO()) as fake_output:
+                m.add_amount = MagicMock(return_value=True)
+                m.add()
+                self.assertEqual(fake_output.getvalue(), '>> nobody not in list, adding it\n')
```
natasha-aleksandrova: I don't think you need to mock print statements or `add_amount`

you should be able to verify that a donor was added to `donors` or a donation amount was added to existing donor in `donors`

PR #440, opened 2018-04-16
lesson05 ready for review

idcrickmore:
I added a number of exception handling statements to mailroom_pt3 but found only limited application for comprehension statements (Primarily in the report() function). If you see any other obvious locations that comprehension statements would make sense please point them out.

also, please note that the instructions for the Exceptions Exercise state that the final printout should be 

"Whoops! there is no joke for: spam

Customer: Not much of a cheese shop really, is it?
Shopkeeper: Finest in the district, sir.
Customer: And what leads you to that conclusion?
Shopkeeper: Well, it's so clean.
Customer:  It's certainly uncontaminated by cheese."

I could be wrong (I was pretty confused by the instruction in this exercise) but I think the final print out should start with, "Spam, Spam, Spam, Spam, Beautiful Spam", not "Whoops! there is no joke for: spam".

Thank you,

Ian David


students/idcrickmore/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,140 @@
...
+1. Send Thank You
+2. Export Letters to Everyone
+3. Create a Report
+4. Quit\n""")
+
+    choice = input("-> ").lower()[0]
+
+    # use .get() on the 'options' dict to call functions
+    # keep calling the 'menu' function if user input not in dict
+    options.get(choice, menu)()
```
natasha-aleksandrova: should have an exception handling here if user enters invalid choice.

i would recommend accessing by name instead of via get() to catch `KeyError`

students/idcrickmore/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,140 @@
...
+        else:
+            name_check = name_check.title()
+            # check to see if 'name_check' is in the donor dict...
+            # if it is, append it
+            try:
+                donors[name_check].append(donation)
+            except KeyError:
+                # if 'name_check' isn't in the donor dict...
+                # set  the default format value as a list ([]) and append it
+                donors.setdefault((name_check), []).append(donation)
```
natasha-aleksandrova: seems unnecessary to catch KeyError when you could use `setdefault` on line 45

students/idcrickmore/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,140 @@
...
+
+    print("{:<15}{:>17}{:>15}{:>10}".format(*column))
+    print("---------------------------------------------------------------")
+
+    # loops through 'donors_report' and
+    # dumps all values for 'name' into .format
+    for name in donors_report:
+        print('{:<20} ${:>13.2f}{:>12}  ${:>10.2f}'.format(*name))
+
+    menu()
```
natasha-aleksandrova: from extendability point of view, it makes more sense to have while True loop with redirection to "work" functions.

Imagine you need to add another functionality option, whoever adds that function needs to remember to call `main` on it which is error prone.

If you had while True loop, whoever would extend the program would only need to consider new functionality and not worry about the main program flow.

PR #441, opened 2018-04-16
Lesson07

nDruP:
A working html_render.py!


students/nDruP/lesson07/html_render.py
```diff
@@ -0,0 +1,118 @@
...
+    pass
+
+
+class SelfClosingTag(Element):
+    def __init__(self, **kwargs):
+        self.attributes = kwargs
+
+    def render(self, file_out, cur_ind=""):
+        file_out.write(self.open_tag_str(cur_ind)[:-1]+" />"+'\n')
+    pass
```
natasha-aleksandrova: you don't need `pass` keywords here and other places

students/nDruP/lesson07/html_render.py
```diff
@@ -0,0 +1,118 @@
+class Element:
+    tag = "html"
+    indent = 1
+
+    def __init__(self, content=None, **kwargs):
+        self.substance = () if content is None else (content,)
```
natasha-aleksandrova: does it need to be a `tuple`?

PR #442, opened 2018-04-16
Lesson 07

MikeS-1:
Here's what I have so far. Lots of work, this one!


students/MikeShand/lesson07/test_html_render.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+
+
+import html_render as hr
+
+
+def test_element():
+    test1 = hr.Element('intro', color = 'blue')
+    test1.append('test text')
+    assert 'test text' and 'intro' in test1.content
```
natasha-aleksandrova: are you attempting to assert that both 'test text' & 'intro' are in the test1.content? if that is so then the above code does not do that

let me know, I may be misunderstanding

MikeS-1:
Unfortunately, that's exactly what I was trying to do! Any insight you can provide would be greatly appreciated.

Also, I felt like I should test more of the functionality of the element object  like the render function and the ability to incorporate other subclasses objects. I'd be grateful for any insights you can share on that subject as well.

natasha-aleksandrova:
i see!
so assert does a boolean test for True or False, order of operation here:
`'test text' and 'intro' in test1.content`
will check that `"test text"` is truthy and it is! so will always pass

here is what you want to do instead (or I guess options):
* `['test text', 'intro'] == test1.content` 
* if i'm lazy and don't feel like worrying about order: `sorted(['test text', 'intro']) == sorted(test1.content)`
* if you do not know how many elements in your left side list, then you can check that your expected is a subset of actual list: `set(['test text', 'intro']).issubset(test1.content)`

then just add assert in front of it like: `assert ['test text', 'intro'] == test1.content`

natasha-aleksandrova:
on your testing comment, yes agreed. Testing `render` functions would be the most useful test, you can pass in `StringIO` instance for `file_out` and verify correct string was written

PR #443, opened 2018-04-16
completed lesson09 mailroom assignment

rmart300:


students/rmart300/lesson09/donor.py
```diff
@@ -0,0 +1,18 @@
+
+
+class Donor(object):
```
robalford: In Python 3 it's not necessary to explicitly inherit from `object` (though all classes do by default). So this can be `class Donor:` without the parens.

students/rmart300/lesson09/donor.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def __init__(self, name):
+
+        try:
+            name.split()[1]
+        except IndexError as e:
```
robalford: You don't need the `as e` part of the `except` clause here, since you're not referencing the `e` variable in the `except` block. This syntax is generally used for logging or printing specific errors in the except block when you catch the exception.

students/rmart300/lesson09/donor_collection.py
```diff
@@ -0,0 +1,92 @@
+import os
+from donor import Donor
+
+class DonorCollection(object):
+
+    letter_directory = 'temp/'
+
+    def __init__(self):
+        # dictionary to hold donors and list of donation amounts
+        donor_list = ['Fred Smith', 'Terrie Ann', 'Murray Martin', 'Josh Jones', 'Jane Doe']
```
robalford: I'd recommend letting the user create this list when the object is instantiated, rather than beginning with the same hard-coded list for each DonorCollection object.

students/rmart300/lesson09/donor_collection.py
```diff
@@ -0,0 +1,92 @@
...
+        self.donation_dict = {}
+        self.donor_set = set()
+
+        # build initial dictionary of donors and donation amounts
+        amount_list = [ 500, 100, 1000]
```
robalford: Donations should probably be stored as an attribute on the `Donor` object, since they belong to a particular donor

students/rmart300/lesson09/donor_collection.py
```diff
@@ -0,0 +1,92 @@
...
+        """ Print a list of donors, sorted by total historical donation amount"""
+        title = "{0:20} | {1:15} | {2:10} | {3:15}".format('Donor Name','Total Given','Num Gifts','Average Gift')
+        print(title)
+
+        #sort the dictionary by descending order of the sum of values
+        sorted_dict = sorted(self.donation_dict.items(), key=lambda x: sum(int(v) for v in x[1]),reverse=True)
+        for donor, amount_list in sorted_dict:
+            donation_count = len(amount_list)
+            donation_total = sum(int(v) for v in amount_list)
+            donation_average = round(donation_total/donation_count,2)
```
robalford: These donation calculations would be a good use case for `@properties` on the `Donor` model, if you move the donation data there

students/rmart300/lesson09/mailroom.py
```diff
@@ -0,0 +1,36 @@
+from datetime import date
+from donor_collection import DonorCollection
+
+def validate_user_selection(action, action_dict):
+    """ validate the user selection versus available actions """
+
+    key_list = list(action_dict.keys())
+    try:
+        if int(action) not in key_list:
```
robalford: This can be `if int(action) not in action_dict:` without the assignment to `key_list` because the `in` operator checks for key membership in dict

students/rmart300/lesson09/test_donor_collection.py
```diff
@@ -0,0 +1,51 @@
...
+import unittest
+import sys, os 
+
+class TestDonorCollection(unittest.TestCase):
+
+    bad_name = 'Wolf'
+    bad_amount = 'not a number'
+    good_name = 'Wolf Man'
+    good_amount = '900'
+
```
robalford: You can use the `unittest.TestCase.setUp` method for shared test data like this

PR #446, opened 2018-04-17
this is a readme file

vvinodh:


students/vvinodh/Lesson1/break_me.py
```diff
@@ -0,0 +1,19 @@
...
+    a=a +1
+    print(a)
+type_error_func()
+
+def Syntax_error():
+    print("Generating syntax error)
+
+import math
+x='Hello'
+math.calculate(x)
```
natasha-aleksandrova: I know this is a silly exercise but couple things to keep in mind as you learning python (style wise):
* you want your imports at the top of the file
* function names are `snake_case` so are variables
* you want spacing between operators like `=` unless it's a keyword argument

PR #447, opened 2018-04-17
lesson_09 submission

slickw1lly:


students/khtruong/lesson_09/mailroom_oop.py
```diff
@@ -0,0 +1,180 @@
...
+        if donors:
+            self.donors = donors
+        else:
+            self.donors = []
+
+    def donor_and_amount(self):
+        """Return prompt for donor name and donation amount and print to
+        screen.
+        """
+        fullname = fullname_input()
```
natasha-aleksandrova: "There should be no input function in the class that holds the data!"

as mentioned in the instructions you should separate the user inputs from your data handling code.

the user interaction code should be calling this class with already collected input like: 
`dbobj.add_donation(amount)` where amount was collected outside of class

doing this keeps the class clean and on topic plus easier to unit test the class itself

natasha-aleksandrova: to give you a more practical example, imagine you are now asked to write a GUI for this mailroom program instead of cli.

you should be able to take your `Database` class and use it in the front end code without refactor

students/khtruong/lesson_09/mailroom_oop.py
```diff
@@ -0,0 +1,180 @@
...
+                donor.numdonation, donor.avgdonation))
+
+    def send_letters(self):
+        """Create thank you letter for each donor and save as text files"""
+        for donor in self.donors:
+            fullname = donor.fullname
+            amount = donor.donations[-1]
+            filename = ''.join([fullname, ' ',
+                                str(date.today()),
+                                '.txt'])
```
natasha-aleksandrova: something like an f-string would read better here:
`f"{fullname}_{date.today()}.txt"` (I would also avoid empty spaces in file names)

students/khtruong/lesson_09/mailroom_oop.py
```diff
@@ -0,0 +1,180 @@
...
+def menu_selection(prompt, selection_dict):
+    """Return options for user to select from."""
+    while True:
+        userinput = input(prompt)
+        if selection_dict.get(userinput) is None:
+            print('\n{} is not a valid selection. '
+                  'Please try again!'.format(userinput))
+        else:
+            if selection_dict.get(userinput)() == 'Exit Menu':
+                break
```
natasha-aleksandrova: catching `KeyError` and using `sys.exit` would be very nice here

PR #448, opened 2018-04-17
submit lesson08

tricaonguyen:


students/tri_nguyen/lesson08/Circle_class.py
```diff
@@ -0,0 +1,94 @@
...
+    def __str__(self):
+        return '{} with radius: {}'.format(self.__class__.__name__, self.radius)
+
+    def __repr__(self):
+        return '{}({})'.format(self.__class__.__name__, self.radius)
+
+    def __add__(self, other):
+        if isinstance(other, Circle):
+            other = other.radius
+            return '{}({})'.format(self.__class__.__name__, self.radius + other)
```
natasha-aleksandrova: this should return a circle object not a string representation of it

students/tri_nguyen/lesson08/Circle_class.py
```diff
@@ -0,0 +1,94 @@
...
+            raise TypeError('You can only add 2 Circle instances.')
+
+    def __iadd__(self, other):
+        return self.__add__(other)
+
+
+    def __mul__(self, other):
+        if isinstance(other, (int, float)):
+            self.radius *= other
+            return '{}({})'.format(self.__class__.__name__, self.radius)
```
natasha-aleksandrova: same here, should return object

students/tri_nguyen/lesson08/SparseArray_class.py
```diff
@@ -0,0 +1,45 @@
+class SparseArray:
+    def __init__(self, array):
+        self.array = array
+
+    def store_nonzero(self):
+        return {idx:item for idx, item in enumerate(self.array) if item !=0}
```
natasha-aleksandrova: this would be better to go into constructor and stored non-zero array into `self.array`

students/tri_nguyen/lesson08/test_Circle_class.py
```diff
@@ -0,0 +1,108 @@
+# ------------------------------------------- #
+# Student:   Tri Nguyen
+# Instructor: Natasha Aleksandrova
+# Date:  01-Apr-2018
+# ------------------------------------------- #
+
+import unittest
+from Circle_class import Circle
```
natasha-aleksandrova: per pep8, modules should be lower case:
`from circle_class import Circle` and _class shouldn't really be part of it in real scenarios -> `from circle import Circle`

PR #449, opened 2018-04-17
lesson04 trigram submission

prgrover:


students/prgrover/lesson04/trigram.py
```diff
@@ -0,0 +1,115 @@
...
+    print()
+    print(sample)
+
+
+# Function calls to execute the process
+pure_text = open_text()
+cleansed_text = clean_text(pure_text)
+word_list = text_to_list(cleansed_text)
+trigrams = list_to_dict(word_list)
+create_new_sample(trigrams)
```
natasha-aleksandrova: nice work on breaking the down the problem into reasonable functions that help readability

students/prgrover/lesson04/trigram.py
```diff
@@ -0,0 +1,115 @@
...
+
+def create_new_sample(trigram_dict):
+    """
+    Create a new text sample using the trigram dictionary.
+
+    Args:
+    trigram_dict:  Dictionary of trigrams.
+    """
+    num_words = get_sample_size()
+    dict_keys = list(random.choice(list(trigram_dict.keys())))
```
natasha-aleksandrova: not necessary to call `keys()`

PR #451, opened 2018-04-17
Adding mailroom2.py for Lesson04

sbolsen:


students/sbolsen/lesson04/mailroom2.py
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+
+def donor_list():
+    for donor in donors:
+        print(donor)
+
+
+def prompt_user():
+    choice = input('Please select from the following: \n1) Send a Thank You \n2) Create a Report \n3) Send letters to everyone \n4) Quit\n')
```
robalford: Watch the line lengths on these strings. You can do something like this for Pep8:
```
    choice = input('Please select from the following: \n1) Send a Thank You \n2) Create a Report \n3) Send letters to '
                   'everyone \n4) Quit\n')
```
Or use triple quoted strings to preserve line breaks in your strings

students/sbolsen/lesson04/mailroom2.py
```diff
@@ -0,0 +1,78 @@
...
+        donor_list()
+        select_user = input('Enter a full name: ')
+    return select_user
+
+
+def user():
+    select_user = thank_you()
+    update_donation = int(input('Enter a donation amount: '))
+    try:
+        if select_user in donors.keys():
```
robalford: The `in` operator checks for key membership in dict, so no need to call `keys()` here

students/sbolsen/lesson04/mailroom2.py
```diff
@@ -0,0 +1,78 @@
...
+        a = int(prompt_user())
+        decision = {
+                1: user,
+                2: create_report,
+                3: send_letters,
+                4: quit,
+                }
+        try:
+            decision[a]()
+        except Exception as e:
```
robalford: It's always a good idea to narrow in on the particular exception that you want to catch, seems like `KeyError` is what you're after in this case

PR #452, opened 2018-04-17
Pull Request (plus cleanup/sync)

kmsnyde:


students/kmsnyde/lesson04/trigrams.py
```diff
@@ -0,0 +1,96 @@
...
+        new_list.extend(k.split(' '))
+        #print(new_list)
+        
+def build_list():
+    for i, (k, v) in enumerate(d.items()):
+        if i < len(d.keys()):
+            init_key = new_list[-2] + ' ' + new_list[-1]
+            #print(init_key)
+            #print(random.choice(d[init_key]))
+            new_list.append(random.choice(d[init_key]))
```
natasha-aleksandrova: I think you emailed about this issue (IndexError) which I see as well when running your code.

You want to stop your for loop when you can no longer find a key (word pair). 
`if i < len(d.keys())` is not a good way to stop your loop as you don't know when you run out of word pairs to build your sentence. You should exit your loop if `if not d.get(init_key): return`

students/kmsnyde/lesson04/trigrams.py
```diff
@@ -0,0 +1,96 @@
...
+d = defaultdict(list, [])
+for i, w in enumerate(text):
+    if i < len(text)-2:
+        d[(text[i] + ' ' + text[i+1])].append(text[i+2])
+#print(d)
+new_list = []
+for i, (k, v) in enumerate(d.items()):
+    if (i == 40): #40 chosen at random
+        new_list.extend(k.split(' '))
+        #print(new_list)
```
natasha-aleksandrova: 1. you should have a `break` statement to avoid wasting loop cycles
2. would be good to use random lib for this random selection

PR #454, opened 2018-04-18
mailroom, part 3, with exception handling added. 

natrod22:
Added exception handling for the 'thank_all()' function's 'FileNotFound' Error, provided an error message, and allowed the user another attempt at providing a valid file location.

### DIFF
--- a/students/NatalieRodriguez/Lesson05/mailroom3.py
+++ b/students/NatalieRodriguez/Lesson05/mailroom3.py
@@ -1,0 +1,169 @@
+#!/usr/bin/env python3
+
+# Lesson 05: Mailroom: Part 3, With Exception Handling
+# Natalie Rodriguez
+# April 17, 2018
+
+# Requirements:
+# 1)use dictionaries, when possible
+# 2)use a dictionary to switch between user selections
+# 3)Try to use a dict and the .format() method to do the letter as one big template
+# rather than building strings in parts.
+# 4) Add file writing functionality, i.e. a function (and a menu item to invoke it)
+# to go through all the donors in your donor data structure, generate a thank you
+# and write it to disk as a text file.
+
+import os
+import pathlib
+
+# Initial donor list and the amounts they have donated
+donors = {'Luke Rodriguez':[12.75,50.31,42.59],'Emma Burgess':[390.87,30.00],'Greg Cayetano':[875.60,1120.00],
+'Hannah Watson':[20.58,1120.14],'Emily Connor':[10.00],'Catherine Davis':[400.00],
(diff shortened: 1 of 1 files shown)


students/NatalieRodriguez/Lesson05/mailroom3.py
```diff
@@ -0,0 +1,169 @@
...
+
+    response = ''
+    while response != '4':
+        print()
+        for i in options:
+            print(i, options[i][0])
+
+        # Get user's selection
+        response = ''
+        while not response in options:
```
robalford: This should be `while response not in options:`

students/NatalieRodriguez/Lesson05/mailroom3.py
```diff
@@ -0,0 +1,169 @@
...
+            response = input("Type your selection: ").strip()
+
+        options[response][1]()  # Call helper function
```
robalford: This would be a good place for a try/except block to validate that the user input is a key in the dictionary

students/NatalieRodriguez/Lesson05/mailroom3.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+def exit():
```
robalford: I would rename this function to something other than `exit` since this is overwriting a built-in Python function

students/NatalieRodriguez/Lesson05/mailroom3.py
```diff
@@ -0,0 +1,169 @@
...
+            if donation == 'exit':
+                exit()
+                return
+
+            elif donation.strip('0123456789.') == '' and len(
+                    donation) > 0 and donation.count('.') <= 1 and float(
+                donation) > 0.0:
+                break
+
+        donation = float(donation)
```
robalford: This would be another good opportunity for a try/except block on the float conversion, to validate that you're getting numeric input

students/NatalieRodriguez/Lesson05/mailroom3.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+def thank_all():
+
+    cur_dir = os.getcwd()
+    print('\nThe current directory is %s' % cur_dir)
+    new_dir = input('\nEnter the directory you want to save to '
+                    '(Press enter to save in the current location): ').strip()
+#exception handling changes here: indented all for try/except block
+    try:
```
robalford: It's a good idea to try and have as little code inside your `try` block as possible, so that you only include code that might raise the expected exception. Otherwise, you can end up with unanticipated and unhandled errors slipping by.

PR #455, opened 2018-04-18
add kata.py

iamAndrewKim:


students/AndrewKim/lesson04/kata.py
```diff
@@ -0,0 +1,53 @@
...
+    book = book.translate(puncs_table)
+
+    words = book.split()
+    for i in range(len(words)-2):
+        trigram_dict[words[i] + ' ' + words[i+1]].append(words[i+2])
+    return trigram_dict
+
+def create_trigram_output():
+    trigram_text = list()
+    trigram_dict = defaultdict(list, {})
```
natasha-aleksandrova: no need to define this here since you're redefining a line below

students/AndrewKim/lesson04/kata.py
```diff
@@ -0,0 +1,53 @@
...
+    trigram_dict = create_trigram_words()
+
+    trigram_key = random.choice(list(trigram_dict))
+    trigram_val = trigram_dict.get(trigram_key)[0]
+
+    tri_key_arr = trigram_key.split()
+    for each_text in tri_key_arr:
+        trigram_text.append(each_text)
+    trigram_text.append(trigram_val)
```
natasha-aleksandrova: isn't this just word pair? no need for a `for` loop. -> `trigram_text.extend(tri_key_arr)`

students/AndrewKim/lesson04/kata.py
```diff
@@ -0,0 +1,53 @@
...
+
+    rep = 0
+    while trigram_key and rep < 100:
```
natasha-aleksandrova: I think you can remove `trigram_key` condition since you are breaking and I don't see when it would be falsy

PR #462, opened 2018-04-20
Lesson08|circle.py

nDruP:
I forgot to do frequent commits. Overall a pretty straightforward lesson. I would've liked to see how magic methods and properties carry over to subclasses, but we'll probably get to that in a future lesson.

### DIFF
--- a/students/nDruP/lesson08/circle.py
+++ b/students/nDruP/lesson08/circle.py
@@ -1,0 +1,98 @@
+"""
+A Circle can be defined by either specifying the radius or the diameter, and the user can query the circle for either its radius or diameter.
+
+Other abilities of a Circle instance:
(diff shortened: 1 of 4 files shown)


students/nDruP/lesson08/circle.py
```diff
@@ -0,0 +1,98 @@
...
+        Print the circle and get something nice
+        Be able to add two circles together
+        Be able to compare two circles to see which is bigger
+        Be able to compare to see if there are equal
+        (follows from above) be able to put them in a list and sort them
+
+You will use: properties, a classmethod, a define a bunch of “special methods”
+"""
+from math import *
+
```
natasha-aleksandrova: you generally want to avoid star imports

PR #463, opened 2018-04-20
submitting lesson8 for review

wesmoy17:


students/WesleyM/Lesson8/circle.py
```diff
@@ -0,0 +1,44 @@
...
+    
+    def __mul__(self, other):
+        return Circle(self.radius * other)
+    
+    def __lt__(self,other):
+        return self.radius < other.radius
+        
+    def __eq__(self, other):
+        return self.radius == other.radius
+
```
robalford: If you use the `@total_ordering` decorator from the `functools` module of the Python standard library to decorate your class, you only need to define `__eq__` and `__lt__`, otherwise, you should probably define the rest of the comparisons like `__ne__` for `!=`, `__le__` for `<=` etc.

robalford:
@wesmoy17 I don't see the assignment submission on the course website. Remember to submit it there as well so we can grade.

PR #464, opened 2018-04-20
Adding trigram excercise for Lesson04

sbolsen:


students/sbolsen/lesson04/trigram.py
```diff
@@ -0,0 +1,31 @@
...
+    new_dict = {}
+    for item in range(0, len(split_text)-2):
+        new_dict.setdefault('{} {}'.format(split_text[item], split_text[item+1]), [])
+        new_dict['{} {}'.format(split_text[item], split_text[item+1])].append('{}'.format(split_text[item+2]))
+        chapter = random.choice(list(new_dict)).split()
+    try:
+        while True:
+            for word in range(0, len(chapter)-1):
+                chapter.append(random.choice(new_dict[' '.join(chapter[-2:])]))
+    except KeyError:
```
natasha-aleksandrova: suggestion: put try/except around line 20 which is where the KeyError can happen, and in except block call `break`
then write new file in separate code block below.

PR #465, opened 2018-04-21
requesting feedback on current progress

johnpharmd:


students/johnpharmd/lesson07/html_render.py
```diff
@@ -0,0 +1,50 @@
...
+        print('open_tag ==', self.open_tag)
+        print('self.content ==', self.content)
+        for l in self.content:
+            for s in l:
+                file_out.write(cur_ind + s + '\n')
+        file_out.write(self.close_tag)
+
+
+class Html(Element):
+    """renders html"""
```
natasha-aleksandrova: you should define class variable `tag = 'html'`

you don't the `__init__` in all of the subclasses, you just need to define `tag` class variable for each child class

students/johnpharmd/lesson07/html_render.py
```diff
@@ -0,0 +1,50 @@
+class Element(object):
+    """renders an html element"""
+
+    tag = ''
+    indent = '    '
+
+    def __init__(self, content=''):
+        self.open_tag = '<' + self.tag + '>'
+        self.close_tag = '</' + self.tag + '>'
```
natasha-aleksandrova: since you're not going to use `self.open_tag` and `self.close_tag` in various methods there is not reason to put it in `__init__`, instead have it as local vars in `render` method

students/johnpharmd/lesson07/html_render.py
```diff
@@ -0,0 +1,50 @@
...
+    def append(self, new_string):
+        # self.content = '<br>'.join([self.content, new_string])
+        return self.content.append(new_string)
+
+    def render(self, file_out, cur_ind=''):
+        file_out.write(self.open_tag)
+        print('open_tag ==', self.open_tag)
+        print('self.content ==', self.content)
+        for l in self.content:
+            for s in l:
```
natasha-aleksandrova: I don't think you need the second `for` loop

PR #467, opened 2018-04-22
 Lesson3 Edits + Dictionary/Set Lab and Mailroom, Part 2

DeFeNdog:


students/carlos_novoa/lesson04/mailroom.py
```diff
@@ -0,0 +1,254 @@
...
+            'first_name': 'Andy',
+            'last_name': 'Bernard'
+        },
+        {
+            'donations': [
+                500.00
+            ]
+        }
+    ]
+]
```
natasha-aleksandrova: you should use `dict` as a data structure for donors instead of a list, which is easier and more efficient in this use case

students/carlos_novoa/lesson04/mailroom.py
```diff
@@ -0,0 +1,254 @@
...
+def exit_menu():
+    menu_selection(main_prompt, main_dispatch)
+    return "break"
+
+
+def menu_selection(prompt, dispatch_dict):
+    while True:
+        r = input(prompt)
+        if dispatch_dict[r]() == "break":
+            break
```
natasha-aleksandrova: what is the reason for `exit_menu` called in functions?

you are in while True loop which should run forever until `q` option is selected
your while True will call a function via dispatch_dict, your function will do its work then exit and then loop continues waiting for user input, there isn't really a need for manual control of the flow

students/carlos_novoa/lesson04/mailroom.py
```diff
@@ -0,0 +1,254 @@
...
+    if donor_found:
+        # append existing donor
+        append_donation(donor_name, donation)
+    else:
+        # add new donor
+        add_donor(donor_name, donation)
+
+    donor_email(donor_name, donation)
+    exit_menu()
+    return
```
natasha-aleksandrova: you don't need `return` here

students/carlos_novoa/lesson04/mailroom.py
```diff
@@ -0,0 +1,254 @@
...
+    cwd = os.getcwd()
+    date = now.strftime('%Y-%m-%d')
+    path = cwd + '/letters/'
+    ext = '.txt'
+    for d in _donors:
+        fn = d[0]['first_name']
+        ln = d[0]['last_name']
+        file_path = "{}{}_{}_{}{}".format(path, date, fn, ln, ext)
+        donations = format_donation(sum(d[1]['donations']))
+        with open(file_path, 'w') as file:
```
natasha-aleksandrova: don't use `file` as var name, it's a built-in

students/carlos_novoa/lesson04/mailroom.py
```diff
@@ -0,0 +1,254 @@
...
+        {
+            'first_name': 'Jim',
+            'last_name': 'Halpert'
+        },
+        {
+            'donations': [
+                1.00
+            ]
+        }
+    ],
```
natasha-aleksandrova: if you are to use a list then you would do a list of dictionaries, there is no reason to have two dicts, so it should look like this:
```
donors = [{"first_name": "Jim", "last_name": "Halpert", "donations": [1.00]}, {"first_name": "Pam", "last_name": "beesly", "donations": [3]}]
```

students/carlos_novoa/lesson04/mailroom.py
```diff
@@ -0,0 +1,254 @@
...
+                    '               Sincerely,\n'
+                    '                  -The Team')
+            body = text.format(fn, ln, donations)
+            file.write(body)
+    return
+
+
+def donor_name_split(donor_name):
+    donor_name = donor_name.split(' ')
+    return [donor_name[0], donor_name[1]]
```
natasha-aleksandrova: this is recreating what `donor_name.split(' ')` returns which means you don't need a function for this and can do it inline

this is what i mean when i say they are equal
```
donor_name.split() == [donor_name[0], donor_name[1]]
```

PR #470, opened 2018-04-22
Lesson 9: assignment object oriented mailroom

JeroenSweerts:


students/JeroenSweerts/Lesson9/mailroom5.py
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env python3
+
+class Donor():
```
robalford: You don't need the parens here for class definition

students/JeroenSweerts/Lesson9/mailroom5.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self, name=None):
```
robalford: Don't see why this needs to default to `None`

students/JeroenSweerts/Lesson9/mailroom5.py
```diff
@@ -0,0 +1,158 @@
...
+        self._name = name
+        self._donations = []
```
robalford: You don't appear to be using this `_donations` attribute anywhere, or ever storing any values in it

students/JeroenSweerts/Lesson9/mailroom5.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @property
+    def name(self):
+        """Property for the donor's name."""
+        return self._name
+
+    @name.setter
+    def name(self, val):
+        """Setter for the donor's name."""
+        self._name = val
```
robalford: These getters and setters are not necessary, since they're not doing any extra work beyond directly accessing and setting the values of the instance attributes. In Python `@propery` is generally used for calculated values, and `setter` is used where there's some extra work that needs to be done when an attribute is set.

students/JeroenSweerts/Lesson9/mailroom5.py
```diff
@@ -0,0 +1,158 @@
...
+        """Property for the donor's list of donations."""
+        return self._donations
+
+    @donations.setter
+    def donations(self, val):
+        """Setter for the donor's list of donations."""
+        self._donations = val
+
+
+class Donor_Collection(dict):
```
robalford: It doesn't seem like you need to subclass from `dict` here, since you aren't using this class as a dict, but rather passing a dict into its constructor as an instance attribute.

students/JeroenSweerts/Lesson9/mailroom5.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self, donors):
+        self._donors = donors
```
robalford: This overwrites the built-in `dict` class's constructor without calling `super().__init__()`

students/JeroenSweerts/Lesson9/mailroom5.py
```diff
@@ -0,0 +1,158 @@
...
+        self._donors = val
+
+    def add_donor(self, donor):
+        """Add a new donor to the collection."""
+        self.donors[donor.name] = donor.donations
+
+    def donate(self, name, donation):
+        """Add a donation to the list of donations for a specific donor.
+        If the donor doesn't exist yet, then add it to the donor collection first."""
+        if name not in self.donors.keys():
```
robalford: `in` checks for key membership in dict

students/JeroenSweerts/Lesson9/mailroom5.py
```diff
@@ -0,0 +1,158 @@
...
+            new_donor = Donor(name)
```
robalford: This seems to be the only place that you're using the `Donor` class, and all it's doing is passing the donor's name into the `donor` dict, so as this is written, you really don't need the `Donor` class at all. You should think about reworking you code to encapsulate all the donor-related logic inside of the `Donor` class. This might include storing and accessing donations, as well as calculating the donation values that you reference in the report.

students/JeroenSweerts/Lesson9/mailroom5.py
```diff
@@ -0,0 +1,158 @@
...
+            self.add_donor(new_donor)
+        if not donation.isdigit():
+            raise Exception("ERROR: You didn't enter a number or a negative number.")
```
robalford: It's almost always better to catch a specific exception, so that you can write the correct code to handle it.

students/JeroenSweerts/Lesson9/mailroom5.py
```diff
@@ -0,0 +1,158 @@
...
+        outfile.write("\t\t   -The Team \n")
+        outfile.close()
+
+    def create_report(self):
+        """For all donors in the Donor_Collection class, write a report containing
+        the donor's name, how much he gave in total, how many gifts the donor made and
+        how much the donor gave on average."""
+        print(f'{"Donor Name":20s} {"|  Total Given":20s} {"|  Num Gifts  |":20s} {"Average Gift":20s}')
+        print(f'{"-"*76}')
+        [print(f'{name:20s} ${sum(self.donors[name]):20.2f} {len(self.donors[name]):13d}${sum(self.donors[name])/len(self.donors[name]):20.2f}') for name in self.donors.keys()]
```
robalford: List comprehensions are designed to return a value and are generally used for building a new list based on some filter or mapping operation performed while iterating through another list

students/JeroenSweerts/Lesson9/test_mailroom5.py
```diff
@@ -0,0 +1,59 @@
...
+        self.DonorDict = mailroom5.Donor_Collection(self.donors)
+        mailroom5.thankyoueveryone(self.DonorDict)
+
+    def test_create_report(self):
+        self.DonorDict = mailroom5.Donor_Collection(self.donors)
+        self.DonorDict.create_report()
+
+    def test_report(self):
+        self.DonorDict = mailroom5.Donor_Collection(self.donors)
+        mailroom5.report(self.DonorDict)
```
robalford: Looks like you need to finish up writing these tests

PR #474, opened 2018-04-23
Adding except_exercise.py for Lesson05

sbolsen:


students/sbolsen/lesson05/except_exercise.py
```diff
@@ -0,0 +1,54 @@
...
+
+from except_test import fun, more_fun, last_fun
+
+
+# Figure out what the exception is, catch it and while still
+# in that catch block, try again with the second item in the list
+
+try:
+    first_try = ['spam', 'cheese', 'mr death']
+    joke = fun(first_try[0])
```
robalford: This [18] is the only line that should be inside the `try` block here, since that's the function call that raises `NameError`. The list assignment should happen before the try/except, since there's nothing there related to the caught exception, and the call to `fun(first_try[1])` should go in the `except` block after the initial `NameError` is caught.

students/sbolsen/lesson05/except_exercise.py
```diff
@@ -0,0 +1,54 @@
...
+
+# Finally, while still in the try/except block and regardless of whether
+# there were any exceptions, call the function last_fun with no
+# parameters. (pun intended)
+
+langs = ['java', 'c', 'python']
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    more_joke = more_fun(langs[1])
```
robalford: There should be an `else` clause here calling `more_fun` with the last language in the list

PR #476, opened 2018-04-24
Submitting test_mailroom.py

githubtater:


students/Roy_Tate/lesson06/mailroom4.py
```diff
@@ -0,0 +1,205 @@
...
+    return printed_list
+
+
+# Helper function to add donations to appropriate donor's account
+def add_donation(name, new_donation):
+    if name in donor_dict.keys():
+        donor_dict[name].append(new_donation)
+    else:
+        donor_dict[name] = [new_donation]
+    return donor_dict
```
natasha-aleksandrova: since you're not using the return in your code, probably don't need to return donor_dict?

students/Roy_Tate/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,66 @@
...
+    donor_dict = {}
+    donor_dict['Bill'] = [1100, 2100, 3100]
+    donor_dict['Mark'] = [1000, 300000.23, 2000]
+    donor_dict['Jeff'] = [50, 450.355]
+    donor_dict['Roy'] = [4500, 7500, 11221, 30232, 323]
+
+    # Sum the donations and verify they exist in the donor report
+    def test_correct_sums_printed_in_report(self):
+        for donor, donations in self.donor_dict.items():
+            assert str(donor) in mailroom4.create_report()
```
natasha-aleksandrova: you're not using `donations`

students/Roy_Tate/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # We want to ensure our signature appears as intended
+    # This also ensures that nothing appears after our signature
+    def test_verify_output_file_has_correct_signature(self):
+        output_file = 'Roy.txt'
+        signature = 'The good guys at the best organization'
+        with open(output_file, 'r') as f:
+            lines = f.readlines()
+            self.assertEqual(signature, lines[-1].strip())
```
natasha-aleksandrova: I don't see a mailroom function being called here? If I run this test I will not have Roy.txt file

also your test should first check that file exists, like `assert os.path.isfile`, ideally your asserts should fail not your test code

students/Roy_Tate/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def test_add_donation_new_donor(self):
+        name = 'BillyBobHexagon'
+        donation = 7000
+        # add the new donor to the list
+        donation_list = mailroom4.add_donation(name, donation)
+        # The list returned should contain our new donor
+        assert 'BillyBobHexagon' in donation_list
+        # Ensure the new donation was appended
+        assert donation_list[name] == [donation]
```
natasha-aleksandrova: you don't need to return `donation_list` just for the sake of the test
```
import mailroom

    
def test_add():
    mailroom.add()  # this function just adds a newkey to donors dict simulating new donor
    assert "newkey" in mailroom.donor_dict
```
you can access your "DB" dict via mailroom module

students/Roy_Tate/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,66 @@
...
+        donation = 112233445566
+        # add the new donor to the list
+        donation_list = mailroom4.add_donation(name, donation)
+        # Ensure the new donation was appended
+        assert donation in donation_list[name]
+
+    def test_single_email_output(self):
+        simple_email = 'To: {0}\n' \
+                       'Donation amount: ${1:.2f}'
+        print(mailroom4.print_email(simple_email, 'George', 7000))
```
natasha-aleksandrova: not really test unless you're asserting

students/Roy_Tate/lesson06/test_mailroom4.py
natasha-aleksandrova: if you're using unittest module then it is preferred to use their assert methods like `self.assertEquals` or `self.assertTrue` as they will provide nicer error messages.

if you're using pytest then you want to use `assert` statements.

PR #477, opened 2018-04-24
added Python Pushups exercise found in the break_me.py file inside session01

letoui:


students/ian_letourneau/session01/break_me.py
```diff
@@ -0,0 +1,22 @@
+
+## Ian Letourneau
+## 04/24/2018
+## A script to test various errors
+
+def nameError():
+	print (a)
```
natasha-aleksandrova: make sure your editor is set to exactly 4 spaces and you don't need the spaces around parens

PR #484, opened 2018-04-27
Lesson07

22e88:
Adding lesson07 files html_render.py and test_html_render.py (and run_html_render.py, just for the records...:)
Please review...
Thanks

### DIFF
--- a/students/elmar_m/lesson07/html_render.py
+++ b/students/elmar_m/lesson07/html_render.py
@@ -1,0 +1,1 @@
+#!/usr/bin/env python3
--- a/students/elmar_m/lesson07/html_render.py
+++ b/students/elmar_m/lesson07/html_render.py
@@ -24,6 +24,9 @@ class Element:
 
 class Html(Element):
     tag = 'html'
+    def render(self, io, ind):
+        io.write('<!DOCTYPE html>\n')
+        Element.render(self, io, ind)
 
 class Body(Element):
     tag = 'body'
--- a/students/elmar_m/lesson07/html_render.py
+++ b/students/elmar_m/lesson07/html_render.py
@@ -1,6 +1,10 @@
 #!/usr/bin/env python3
 
-# file: html_render.py
+'''
+file: html_render.py
+elmar_m / 22e88@mailbox.org
+Lesson07: HTML renderer
+'''
 
 class Element:
     tag = 'element'
@@ -64,9 +68,13 @@ class SelfClosingTag(Element):
 class Hr(SelfClosingTag):
     tag = 'hr'
 
+class A(Element):
+    tag = 'a'
+    def __init__(self, target, linktext):
+        self.target = target
+        self.link = linktext
+        self.targetstring = 'href={}'.format(self.target)
 
-
-
-
-
+    def render(self, io, ind):
+        io.write('{}<{} {}>{}</{}>\n'.format(ind, self.tag, self.targetstring, self.link, self.tag))
 
--- a/students/elmar_m/lesson07/html_render.py
+++ b/students/elmar_m/lesson07/html_render.py
@@ -5,10 +5,10 @@
 # define class 'Element':
 class Element:
 
+    # define attributes:
     tag = 'element'
     i_level = 0
 
-    # define attributes:
     def __init__(self, content=None):
         self.content = [content] if content else []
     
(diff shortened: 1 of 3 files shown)


students/elmar_m/lesson07/html_render.py
```diff
@@ -0,0 +1,114 @@
...
+'''
+file: html_render.py
+elmar_m / 22e88@mailbox.org
+Lesson07: HTML renderer
+'''
+
+class Element:
+    tag = 'element'
+    props = []
+    propstring = ''
```
robalford: I don't think that `props` and `propstring` need to be initialized as class attributes with empty values since you're only using them to build the `tagopen` attribute

students/elmar_m/lesson07/html_render.py
```diff
@@ -0,0 +1,114 @@
...
+    def __init__(self, content=None, **kwargs):       
+        self.content = [content] if content else []  
+        self.attrs = kwargs
+        self.tagopen = self.tag
+
+        if self.attrs:
+            for key, value in self.attrs.items():
+                self.props.append('{}="{}"'.format(key,value))
+            self.propstring = ' '.join(self.props)
+            self.tagopen = '{} {}'.format(self.tag, self.propstring)
```
robalford: I would move this logic out of `__init__` and into a separate method that returns the `tagopen` value. Then `prop` and `propstrings` can be variables defined within the scope of that method, rather than class attributes

students/elmar_m/lesson07/html_render.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class Body(Element):
+    tag = 'body'
+
+
+class P(Element):
+    tag = 'p'
+    props = []
+    propstring = ''
```
robalford: See my note above to avoid having to redefine these here

students/elmar_m/lesson07/html_render.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class Head(Element):
+    tag = 'head'
+
+
+class Title(Element):
```
robalford: You should define a `OneLineTag` class for `Title`, `A` and `H` to inherit from

students/elmar_m/lesson07/html_render.py
```diff
@@ -0,0 +1,114 @@
...
+    tag = 'title'
+    def __init__(self, text):
+        self.text = text
```
robalford: You should be able to reference `self.content` in the `render` method and then you don't need to override `__init__` here

students/elmar_m/lesson07/html_render.py
```diff
@@ -0,0 +1,114 @@
...
+class Hr(SelfClosingTag):
+    tag = 'hr'
+
+
+class A(Element):
+    tag = 'a'
+    def __init__(self, target, linktext):
+        self.target = target
+        self.link = linktext
+        self.targetstring = 'href={}'.format(self.target)
```
robalford: If you inherit from a `OneLineTag` super class with a `render` method that works for this subclass, then all you need to do here is pass in the link as a kwarg, so your `__init__` function would look something like this (note the call to `super().__init__`, so we still have access to all the same instance attributes in our subclass methods):
```
    def __init__(self, link, *args, **kwargs):
        kwargs['href'] = link
        super().__init__(*args, **kwargs)
```

students/elmar_m/lesson07/test_html_render.py
```diff
@@ -0,0 +1,165 @@
...
+        self.assertTrue(type(obj.props) == list)
+        self.assertTrue(len(obj.props) == 0)
+        self.assertTrue(hasattr(obj, 'propstring'))
+        self.assertTrue(type(obj.propstring) == str)
+        self.assertTrue(len(obj.propstring) == 0)
+
+
+if __name__ == '__main__':
+    ut.main()
+
```
robalford: These tests need some work. Most of the assert statements are basically testing that Python's class and type systems work as expected. Although it might be a useful exercise to write these kinds of tests while you're learning, in practice you don't need them because Python itself is already very well tested so there's no need to test it yourself.

More useful tests would confirm the logic in your instance methods produces the expected results for various forms of input. So for each class that you define, you might write a test that passes in some content and asserts that the expected result is rendered as html.

PR #487, opened 2018-04-28
Lesson 06 mailroom4.py

kmsnyde:
PR for mailroom4.py and test_mailroom4.py.  Other updates and labs included.


students/kmsnyde/lesson06/mailroom4.py
```diff
@@ -0,0 +1,88 @@
...
+def menu():
+        menu_dict = {'1:': 'Send a Thank You',
+                     '2:': 'Create a Report',
+                     '3:': 'Send letters to everyone',
+                     '4:': 'Quit'}
+        for k,v in menu_dict.items():
+            print(k, v)
+            
+def thank_you_greeting():
+    print('{}'.format('\nType a user\'s name or "list" to show names.'))
```
robalford: No need to call format here since there's no variable content in the string

students/kmsnyde/lesson06/mailroom4.py
```diff
@@ -0,0 +1,88 @@
...
+       
+def create_report():
+    sum_data = [(k, sum(v), len(v), sum(v)/len(v)) for k,v in
+                donor_dict.items()]
+    sum_data = sorted(sum_data, key=lambda x: x[1], reverse=True)
+    print('{:<20} {:>20} {:>20} {:>20}'.format('Donor Name', '| Total Given',
+          '| Num Gifts', '| Average Gift'))
+    print('{}'.format('-' * 83))
+    for item in sum_data:
+        print('{:<20} {:>20.02f} {:>20} {:>20.02f}'.format(*item))
```
robalford: This function would be more testable if it returned the report as a string

students/kmsnyde/lesson06/mailroom4.py
```diff
@@ -0,0 +1,88 @@
...
+        
+letters = 'Dear {},\n\n\tThank you for your total contributions in the amount of ${}.\n\n\tYou are making a difference in the lives of others.\n\n\t\tSincerely,\n\t\t"Working for America"'
+
+def send_letters_all():
+    for name, value in donor_dict.items():
+       with open('Thank_You - {}.txt'.format(name.lower().replace(' ', '_')),
+                 'w') as f:
+           f.write(letters.format(name, sum(value)))
```
robalford: I would move the letter writing logic into its own function so that you can test that

students/kmsnyde/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,40 @@
...
+"""
+Define tests that will work againsts mailroom4.py.
+"""
+
+import mailroom4 as mail
+import os.path
+
+def test_1():
+    donors = mail.donor_dict
+    assert "Karl Stick" in donors
```
robalford: This test is not necessary since it tests that defaultdict works as expected and there's no need to test Python

students/kmsnyde/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,40 @@
...
+# to fail
+def test_2():
+    assert mail.donations() == 50.00
+
+
+#In IPython promp after Donor Amount: from above test, type "May West" and then any number.
+#If you type May West, no error; else AssertionError
+def test_3():
+    mail.send_thanks()
+    assert "May West" in mail.donor_dict
```
robalford: Your tests should be able to be fully automated, which means they should be able to run without any external inputs. That makes testing some parts of mailroom a bit tricky. The best approach is to separate the parts of your program that collect user input from the business logic and to write small, focused functions with testable return values based on various forms of input

PR #488, opened 2018-04-29
lesson08 submission

mmsunday1:


students/M_Sunday/lesson08/circle.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/python
+from math import pi
+
+
+class Circle(object):
+
+    @staticmethod
+    def entry_check(the__radius):
+        if isinstance(the__radius, (str, list, tuple, dict)):
```
robalford: I'd go with a float conversion or some other method to directly raise ValueError on bad input rather than trying to list out all the possible bad types

students/M_Sunday/lesson08/circle.py
```diff
@@ -0,0 +1,78 @@
...
+            raise TypeError("Radius/Diameter entry must be a single, positive, and "
+                            "non-string value")
+        else:
+            if the__radius < 0:
+                raise ValueError("Radius/Diameter must be greater than 0")
+            else:
+                return the__radius
+
+    def __init__(self, the_radius):
+        self._radius = self.entry_check(the_radius)
```
robalford: I don't think you need to call entry_check here since you're calling it in the setter

students/M_Sunday/lesson08/circle.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __repr__(self):
+        return "Circle({})".format(self._radius)
+
+    def __str__(self):
+        return "Circle with radius: {0:.6f}".format(float(self._radius))
+
+    def __add__(self, other):
+        new_circle = self._radius + other._radius
```
robalford: In Python a leading underscore conventionally indicates a 'protected' attribute that should only be accessed within a class or subclass

students/M_Sunday/lesson08/circle.py
```diff
@@ -0,0 +1,78 @@
...
+        return Circle(new_circle)
+
+    def __lt__(self, other):
+        return self._radius < other._radius
+
+    def __gt__(self, other):
+        return self._radius > other._radius
+
+    def __eq__(self, other):
+        return self._radius == other._radius
```
robalford: You can use the functools.total_ordering decorator from the Python standard library and only define __eq__ and __lt__, otherwise should define the other comparion operators as well

students/M_Sunday/lesson08/circle.py
```diff
@@ -0,0 +1,78 @@
...
+    @classmethod
+    def from_diameter(cls, dia):
+        if isinstance(dia, (str, list, tuple, dict)):
+            raise TypeError("Diameter entry must be a single, positive, and "
+                            "non-string value")
+        else:
+            if dia < 0:
+                raise ValueError("Diameter must be greater than 0")
+            else:
+                return cls(dia / 2.0)
```
robalford: Why not call your entry_check method here?

students/M_Sunday/lesson08/test_circle.py
```diff
@@ -0,0 +1,58 @@
...
+
+        with self.assertRaises(TypeError):
+            c = cir.Circle([6, 9])
+
+        with self.assertRaises(TypeError):
+            c = cir.Circle((5, 10))
+
+        with self.assertRaises(TypeError):
+            c = cir.Circle({'radius': 5})
+
```
robalford: I would add more tests for the custom logic in your magic methods, diameter and alternative constructor

PR #489, opened 2018-04-29
completed assignment lesson 10

rmart300:


students/rmart300/lesson10/donor_collection.py
```diff
@@ -0,0 +1,128 @@
...
+
+            factor_list = [my_factor]*len(amount_list)
+            challenge_list = list(map(DonorCollection.challenge, amount_list, factor_list))
+            amount_list = challenge_list[:]
+            #print(amount_list)
+            challenge_sum = reduce(lambda x,y: x+y, amount_list)
+            print(f"If you limited the donations between {self.min_value} and {self.max_value}")
+            print(f"and increased by a factor of {my_factor}, then the total donation for {donor} would be {challenge_sum}!")
+
+
```
natasha-aleksandrova: note that in lesson9 where mailroom OO was introduced, the directions hinted to ensure no user interaction code is part of the data handling class.
the idea is that `DonorCollection` only deals with handling donor collection, and if for example we wanted to create mailroom UI instead of mailroom CLI, you could use this class without much change.

PR #491, opened 2018-04-30
Modified mailroom.py and added functionality to send letters to everyone

shibinmathew:


students/shibin_mathew/lesson4/trigram.py
```diff
@@ -0,0 +1,34 @@
+
+def read_file(filename):
+    # with open(path+'/'+filename, 'r+') as f:
+    #     data = f.read()
+    list_story = []
+    f = open(filename, 'r')
+    for line in f:
+            list_story.append(line.split())
```
natasha-aleksandrova: this line is over-indented (should be 4 spaces)

students/shibin_mathew/lesson4/trigram.py
```diff
@@ -0,0 +1,34 @@
+
+def read_file(filename):
+    # with open(path+'/'+filename, 'r+') as f:
+    #     data = f.read()
+    list_story = []
+    f = open(filename, 'r')
```
natasha-aleksandrova: prefer to use `with` statement when working with files

students/shibin_mathew/lesson4/trigram.py
```diff
@@ -0,0 +1,34 @@
...
+            list_story.append(line.split())
+    return list_story
+
+
+def word_library(list_story, dict_word):
+    #  Adds new words to current list from the story
+    list_punctutation = [',', '.', ';', "'", '"']
+    print(list_story)
+    for i in range(0, len(list_story)):
+        if i + 1 <= len(list_story) and i + 2 <= len(list_story):
```
natasha-aleksandrova: since you're getting length 3 times, might be good to extract into a var

natasha-aleksandrova: logically you only need to check `i + 2 <= len(list_story)` since if +2 will be less than size then + 1 will also be less than size

students/shibin_mathew/lesson4/trigram.py
```diff
@@ -0,0 +1,34 @@
...
+            for word in dict_word:
+                # print(list_story[i] + ' ' + list_story[i+1])
+                if list_story[i] + ' ' + list_story[i + 1] == word:
+                    # if list_story[i + 2] not in list_punctutation:
+                        dict_word[word].append(list_story[i + 2])
```
natasha-aleksandrova: this line is over-indented

students/shibin_mathew/lesson4/trigram.py
```diff
@@ -0,0 +1,34 @@
...
+
+if __name__ == '__main__':
+    filename = 'sherlock_small.txt'
+    dict_wordlist = {'I wish': ['I', 'I'], 'wish I': ['may', 'might'], 'may I': ['wish'], 'I may': ['I'],
+                     'it was': ['at'], 'with the': ['cat']}
+    list_story = read_file(filename)
+    print('Original List')
+    print(dict_wordlist)
+    print('New List')
+    print(word_library(list_story, dict_wordlist))
```
natasha-aleksandrova: I'm a little confused here 🙂 

why is `dict_wordlist` created manually here? you should read in a file and create an equivalent of `dict_wordlist` but using data from the story file. Then re-create story by choosing random starting point of word pair and then random selection of words from a list that matches your word pair key.

shibinmathew:
Hi,

I started working on the tiagram but have not completed it. This request
was for the first part of assignment 4 - sending letters to everyone.

Thanks,

Shibin

natasha-aleksandrova:
I see that makes more sense. Please let us know in the next PR that part of trigram was merged earlier.
Thanks.

students/shibin_mathew/lesson4/mailroom.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
+import sys
+
+
+def option_one():
```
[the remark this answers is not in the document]
shibinmathew: Thanks for the tip. This helped clean up my code!

PR #494, opened 2018-05-01
lesson06 complete

idcrickmore:
Please review mailroom_pt4.py and test_mailroom_pt4.py.

Thank you,

Ian David


students/idcrickmore/lesson06/test_mailroom_pt4.py
```diff
@@ -0,0 +1,39 @@
...
+        
+    assert expected == actual
+    
+    
+def test_letters():
+    assert os.path.isfile('Giovanni Cassini_2018-04-30.txt')
+    assert os.path.isfile('Edmond Halley_2018-04-30.txt')
+    assert os.path.isfile('Edwin Hubble_2018-04-30.txt')
+    assert os.path.isfile('Galileo Galilei_2018-04-30.txt')
+    assert os.path.isfile('Christiaan Huygens_2018-04-30.txt')
```
natasha-aleksandrova: if I run this test on my machine it will fail because those files do not exist.

you are missing a call to your function to generate those files -> `mailroom_pt4.letters()` and then you can do asserts

students/idcrickmore/lesson06/mailroom_pt4.py
```diff
@@ -0,0 +1,157 @@
...
+def donor_list():
+    # returns a list of all the donors
+    print("\n------------------ DONOR LIST ------------------\n")
+
+    name_list = list()
+
+    for name in donors:
+        name_list.append(name)
+
+    return(name_list)
```
natasha-aleksandrova: you can greatly simplify this function: -> `return list(donors)`

students/idcrickmore/lesson06/test_mailroom_pt4.py
natasha-aleksandrova: you can also simplify your test:
```
expected_files = ['file1', 'file2']
for ef in expected_files:
    assert os.path.isfile(ef)
```

PR #496, opened 2018-05-01
Lesson 3 - Mailroom Part 1

traub01-devtrain:
Lesson 3 - Mailroom Part 1 completed and ready for review.


students/JamesTraub/lesson03/mailroom.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
+# James Traub 04-17-18
+# Lesson 3 - Mailroom Lab Exercise, Part 1
+
+
+from datetime import date
+today = str(date.today())
+
+if __name__ == "__main__":
+    print('')
```
natasha-aleksandrova: this block should go at the bottom of your script and call your "main" from here (aka your entry point)

students/JamesTraub/lesson03/mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+        if menu_input == 1:
+            thanking()
+        elif menu_input == 2:
+            donor_report()
+        elif menu_input == 3:
+            print('\n'"Good Bye."'\n')
+            break
+        else:
+            print("Entry not recognized. Please try again.")
+            break
```
natasha-aleksandrova: I would say take out a break on line 75 and allow the user to be asked again,
also change line 63 to `while True` since you're breaking on line 72 no need to have a condition in your while statement.

students/JamesTraub/lesson03/mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+        print(donors_history)
+        print('\n'"Thank you for the donation in {}'s name!".format(name_or_list.title()))
+
+
+def donor_report():
+    print('\n'"This is the Donor History report listing Donor and Donation Amount.")
+    print("-------------------------------------------------------------------")
+    for list_entry in donors_history:
+        total_donation = sum(list_entry[1])
+        print("|{:>32}|{:>32}|".format(list_entry[0].title(), total_donation))
```
natasha-aleksandrova: this report should be sorted per spec

students/JamesTraub/lesson03/mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+    print('\n'"This is the complete list of charitable donors as of " + today + ".\n")
+    for i in donors_history:
+        print(i[0].title())
+        print('')
+    thanking()
+
+
+def thanking():
+    name_or_list = input("Enter donor first and last name or the word donors to see a list of names. ")
+    name_or_list = name_or_list.lower()
```
natasha-aleksandrova: you can also do `.lower()` call inline (line 36)

students/JamesTraub/lesson03/mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+    Enter a 3 to - \'Quit\'\n
+"""
+
+
+def donor_name_list():
+    print('\n'"This is the complete list of charitable donors as of " + today + ".\n")
+    for i in donors_history:
+        print(i[0].title())
+        print('')
+    thanking()
```
natasha-aleksandrova: would recommend calling itself (`thanking()`) after line 40, that way `donor_name_list` function only does what it says it does based on its name.

students/JamesTraub/lesson03/mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+
+def thanking():
+    name_or_list = input("Enter donor first and last name or the word donors to see a list of names. ")
+    name_or_list = name_or_list.lower()
+    donors = ''
+    if name_or_list == 'donors':
+        donor_name_list()
+    else:
+        donation_amount = float(input('\n'"Please enter a donation amount. No dollar sign needed, thank you. "'\n'))
+        for a, b in enumerate(donors_history):
```
natasha-aleksandrova: better to use explicit names like `idx, donor`

students/JamesTraub/lesson03/mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+            if name_or_list == b[0]:
+                donors_history[a][1].append(donation_amount)
+                break
+            else:
+                donors_history.append([name_or_list, [donation_amount]])
+                break
```
natasha-aleksandrova: I don't think this logic block works?
take another look at this, the `else` should be when you could not find a donor that matches that name which can only happen when you exhausted choices in your for loop.

I think the `else` should go with the `for` loop (same indent)

PR #499, opened 2018-05-02
Modified mailroom.py and added exception handling

shibinmathew:
Cleaned up mailroom.py by using defaultdict and added exception handling for lesson 5 part two.


students/shibin_mathew/lesson6/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
+import unittest
+from collections import defaultdict
+from mailroom import *
```
robalford: It's a good idea to avoid using `import *`. See this Stack Overflow [post](https://stackoverflow.com/a/2360808) for more info on why and what alternatives there are to use

students/shibin_mathew/lesson6/mailroom.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
+import sys
+from collections import defaultdict
+
+
+def print_donor_names(donors2):
+    list_test = []
+    for names in donors2:
+        print(names)
+        list_test.append(names)  # Debug - for unit testing
```
robalford: You generally shouldn't have return values that are only for testing or debugging. But you could make the function more testable (and easier to debug) by returning a list to be printed where it is called, rather than printing the list as a side effect of the function

students/shibin_mathew/lesson6/mailroom.py
```diff
@@ -0,0 +1,111 @@
...
+            return
+        else:
+            try:
+                donation = float(donation)
+                donors2[name].append(donation)
+                break  # break out of the while loop
+            except ValueError:
+                print("please input a valid amount")
+                continue
+    return None
```
robalford: Python functions return `None` by default, so there's no need to include this statement here

students/shibin_mathew/lesson6/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+        list_test = []
+        list_test2 = []
+        #  Test 1
+        for names in donors2:
+            list_test.append(names+'.txt')
+        self.assertEqual(send_letters(donors2), list_test)
+        #  Test 2
+        for names in donors:
+            list_test2.append(names+'.txt')
+        self.assertEqual(send_letters(donors), list_test2)
```
robalford: This test would be more useful if it confirmed that the files were actually written to disk with the expected content

students/shibin_mathew/lesson6/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+        #  Test 1
+        for names in donors2:
+            list_test.append([names, round(sum(donors2[names]), 2), len(donors2[names]),
+                             round(sum(donors2[names]) / len(donors2[names]), 2)])
+        self.assertEqual(create_report(donors2), list_test)
+        #  Test 2
+        for names in donors:
+            list_test2.append([names, round(sum(donors[names]), 2), len(donors[names]),
+                             round(sum(donors[names]) / len(donors[names]), 2)])
+        self.assertEqual(create_report(donors), list_test2)
```
robalford: You're on the right track here, but see my note in mailroom about writing testable functions and avoiding return values that are only there for your tests

students/shibin_mathew/lesson6/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+            list_test2.append(names)
+        self.assertEqual(print_donor_names(donors), list_test2)
+
+if __name__ == '__main__':
+    #  Example 1
+    list_ = [('Shibin', 25.25), ('Shibin', 456.25), ('Kimberly', 125.50), ('Kimberly', 5.55), ('Jordy', 55),
+             ('Jordy', 25), ('Andreck', 35)]
+    donors2 = defaultdict(list)
+    for name, amount in list_:
+        donors2[name].append(amount)
```
robalford: You should use the `unittest.TestCase.SetUp()` method for setting up shared test data like this

robalford:
@shibinmathew you're off to a good start with unit testing. Take a look at my notes and see if you can incorporate the suggestions into future assignments. This is ready to merge.

PR #502, opened 2018-05-03
Lesson 06: mailroom with unittest

mayc4t:
Files:
      mailroom4.py
      mailroom4_test.py

### DIFF
--- a/students/mayc4t/lesson06/mailroom4.py
+++ b/students/mayc4t/lesson06/mailroom4.py
@@ -1,0 +1,220 @@
+#!/usr/bin/env python3
+# #############################################################################
+# Written by: Mayc4t
+# Date: 4/27/2018
+# FileID: mailroom4.py
+#
+# Purpose: use unittest
+#Add a full suite of unit tests.
+
+#“Full suite” means all the code is tested. In practice, it’s very hard to test the user interaction, but you can test everything else. Make sure that there is as little untested code in the user interaction portion of the program as possible – hardly any logic.
+
+#This is a big step – you may find that your code is hard to test. If that’s the case, it’s a good sign that you should refactor your code.
+#
+# #############################################################################
+
+
+import collections
+from collections import *
+
+names = ["Lily Maycat", "Lulu Lemon", "Marc Jacobs", "Bobbi Brown", "Kate Spade"]
+gifts   = [i*[5000] for i in range(1,6)]
+
+def init_db():
+    global donors_db
+    donors_db = dict(zip(names, gifts))
+
+def quit():
+    """Quit the top menu."""
+    print ("Quitting this main menu now")
+    return "quit"
+
+# def list_all_donor():
+#     for each_dn in donors_db:
+#         print ("\t\t", each_dn)
+
+
+def is_donor_in_list ( dn_name):
+    """ Check if the dn name is in the dn or not """
+    lc_name = dn_name.strip().lower()
+    #print (dn_name, "  ", lc_name)
+    
+    isIn = False
+    ret_name = None
+
+    for each_dn_name in donors_db.keys():
+        lc_each_dn_name = each_dn_name.strip().lower()
+        if lc_name == lc_each_dn_name:
+            isIn = True
+            #print ("TRUE")
+            ret_name = each_dn_name
+            break
+        else:
+            ret_name = dn_name
+    return isIn, ret_name
+
+
+def enter_dn_name():
+    dn_name = None
+
+    while not dn_name :
+        ans = input_func( "\n\tEnter Donor Name to add"\
+                 "\n\t - list (all donors) "\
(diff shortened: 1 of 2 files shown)


students/mayc4t/lesson06/mailroom4.py
```diff
@@ -0,0 +1,220 @@
...
+
+import collections
+from collections import *
+
+names = ["Lily Maycat", "Lulu Lemon", "Marc Jacobs", "Bobbi Brown", "Kate Spade"]
+gifts   = [i*[5000] for i in range(1,6)]
+
+def init_db():
+    global donors_db
+    donors_db = dict(zip(names, gifts))
```
robalford: I would have this return the db rather than declaring a global inside the function

students/mayc4t/lesson06/mailroom4.py
```diff
@@ -0,0 +1,220 @@
...
+
+def quit():
+    """Quit the top menu."""
+    print ("Quitting this main menu now")
+    return "quit"
+
+# def list_all_donor():
+#     for each_dn in donors_db:
+#         print ("\t\t", each_dn)
+
```
robalford: It's a good idea to remove all commented code before committing

students/mayc4t/lesson06/mailroom4.py
```diff
@@ -0,0 +1,220 @@
...
+
+def is_donor_in_list ( dn_name):
+    """ Check if the dn name is in the dn or not """
+    lc_name = dn_name.strip().lower()
+    #print (dn_name, "  ", lc_name)
+    
+    isIn = False
```
robalford: Remember to use snake_case for Pep8

students/mayc4t/lesson06/mailroom4_test.py
```diff
@@ -0,0 +1,214 @@
...
+        mailroom4.init_db()
+
+    # Clean up after everytime testing
+    def tearDown(self):
+        """
+            Tear Down
+            Clear up all the extra .txt thank you note get sent
+        """
+        for file in os.listdir( os.getcwd()):
+             if '.txt' in file:  os.remove(file)
```
robalford: Nice use of the `setUp` and `tearDown` methods!

students/mayc4t/lesson06/mailroom4_test.py
```diff
@@ -0,0 +1,214 @@
...
+        result = mailroom4.is_donor_in_list("New Person")
+        self.assertTrue( expected, result)
+
+
+    def test_enter_dn_name(self):
+        """
+        Test Enter Donor Name 
+        """
+        def input_mock(self):
+            return 'Michael Kors'
```
robalford: You should be able to use the `mock.patch` `side_effect` parameter to mock input like this

students/mayc4t/lesson06/mailroom4_test.py
```diff
@@ -0,0 +1,214 @@
...
+        mailroom4.input_func = input_mock
+        got_dn_name = mailroom4.enter_dn_name()
+        assert got_dn_name == 'Michael Kors'
```
robalford: You can use `self.assertEqual` here

students/mayc4t/lesson06/mailroom4_test.py
```diff
@@ -0,0 +1,214 @@
...
+        #print('test_check_db: %s', mailroom4.donors_db)
+        golden_donors_db = {
+            'Lily Maycat': [5000],
+            'Lulu Lemon': [5000, 5000],
+            'Marc Jacobs': [5000, 5000, 5000],
+            'Bobbi Brown': [5000, 5000, 5000, 5000],
+            'Kate Spade': [5000, 5000, 5000, 5000, 5000],
+        }
+        result = mailroom4.init_db()
+        self.assertTrue(golden_donors_db , result)
```
robalford: I think you want `assertEqual` here, but not sure that will give you what you expect because `init_db()` currently doesn't have a return value

students/mayc4t/lesson06/mailroom4_test.py
```diff
@@ -0,0 +1,214 @@
...
+            'Lily Maycat': [5000],
+            'Lulu Lemon': [5000, 5000],
+            'Marc Jacobs': [5000, 5000, 5000],
+            'Bobbi Brown': [5000, 5000, 5000, 5000],
+            'Kate Spade': [5000, 5000, 5000, 5000, 5000],
+        }
+        expected = ['Bobbi Brown', 
+                     sum(dn_db['Bobbi Brown']), 
+                     len(dn_db['Bobbi Brown']),
+                     sum(dn_db['Bobbi Brown'])/len(dn_db['Bobbi Brown'])]
```
robalford: I'd probably just hard-code the values here since we don't need to test the built ins

PR #503, opened 2018-05-03
Herela: Lesson 9

harvey-herela:


students/HarveyHerela/lesson09/mailroom_actions.py
```diff
@@ -0,0 +1,68 @@
...
+    def get_total(donor_data):
+        return donor_data[1]
+
+
+    def create_row(donor):
+        fullname = donor.get_name()
+        total = sum(donor.get_donations())
+        num = len(donor.get_donations())
+        avg = total / num
+        return [fullname, total, num, avg]
```
robalford: You could store these calculated donation values as `@properties` on the `Donor` object

students/HarveyHerela/lesson09/mailroom_actions.py
```diff
@@ -0,0 +1,68 @@
...
+            letters.append(
+                (donor.get_name(),
+                donor_template.format(name=donor.get_name(), amount=last, total=total)))
+        except KeyError as kerr:
+            print("Key Error:", kerr, "is not a valid key.",
+                "The only valid keys are: ",
+                "{name}, {amount}, and {total}")
+            return "Failed!"
+
+    return letters
```
robalford: I think the functions in this module could appropriately be defined as methods on the `DonorCollection` class

students/HarveyHerela/lesson09/mailroom_db.py
```diff
@@ -0,0 +1,41 @@
+class Donor:
+    def __init__(self, firstname, lastname):
+        self.name = (firstname, lastname)
```
robalford: This would be a good use for a `@property`

students/HarveyHerela/lesson09/mailroom_main.py
```diff
@@ -0,0 +1,110 @@
+import mailroom_db
```
robalford: You can do `from mailroom_db import Donor, DonorCollection` and then you don't have to type out `mailroom_db` every time you use the classes

students/HarveyHerela/lesson09/mailroom_test.py
```diff
@@ -0,0 +1,117 @@
...
+        d2.add_donation(30)
+        d2.add_donation(40)
+        d3 = mailroom_db.Donor("Audrey", "Hepburn")
+        d3.add_donation(80)
+        d3.add_donation(90)
+        d3.add_donation(100)
+        db.add_donor(d1)
+        db.add_donor(d2)
+        db.add_donor(d3)
+        return db
```
robalford: You can use the `unittest.TestCase.setUp` method for shared test data like this

harvey-herela: I thought about that, but the impression I got was that setup runs every time, and I didn't want to run it everytime (namely, when testing adding a donor). I guess I could've created another subclass for testing that functionality?

PR #508, opened 2018-05-04
[DennisLee][L7][HTML Render] Ready for submission.

DennisLee-DennisLee:


students/DennisLee/lesson07/test_html_render.py
```diff
@@ -0,0 +1,691 @@
...
+        test elements. Also make sure the test input and output string
+        lists have the same number of items.
+        """
+        self.strs_out, self.file, self.fobj = None, None, None
+        self.e1, self.e2, self.e3 = None, None, None
+        self.assertEqual(len(consts.strs_before), len(consts.strs_after))        
+    def tearDown(self):
+        """Delete setup objects and test file (if necessary)."""
+        del self.strs_out, self.file, self.fobj
+        del self.e1, self.e2, self.e3
```
natasha-aleksandrova: what is the purpose of above?

students/DennisLee/lesson07/test_html_render.py
```diff
@@ -0,0 +1,691 @@
...
+                                  # be needed, but I keep getting errors
+                                  # without it here
+        meta = hr.Meta(charset="\t    UTF-8   \n")
+        title = hr.Title("\n\nRandom  \t   Page   \n\t   ")
+        head = hr.Head([meta, title])
+        li_1 = hr.Li("  First  item:  ")
+        para = hr.P("\n\nExplanatory\t\ttext.\n\n", **consts.attrs_pass)
+        link = hr.A("https://github.com/", "A useful site!")
+        para.append(link)
+        hr = hr.Hr()
```
natasha-aleksandrova: you were getting an import error because you are overwriting your html_render module by assigning instance to `hr`

I would also avoid aliases unless there is a really good reason for it, as it reduces readability and as you can see here can also be more likely to collide with var names.

students/DennisLee/lesson07/html_render.py
```diff
@@ -0,0 +1,327 @@
...
+		# Make sure arguments are passed correctly
+		self.verify_render_args(file_out, cur_ind)
+
+		# If `indent` member not currently defined, set it to the passed
+		# indentation argument, and change the current indent level to 0
+		if not self.indent:
+			self.indent, cur_ind = cur_ind, ""
+		file_out.write(cur_ind + '<{0}{1} />\n'.format(
+				self.tag, self.attribute_string))
+		return True
```
natasha-aleksandrova: your `render` method shouldn't return a value. I assume this was done for tests purposes, I would generally avoid adding returns just for unit tests. 
You can test your render function by using StringIO class for `file_out` instance and then getting the string written for assert.

students/DennisLee/lesson07/test_html_render.py
DennisLee-DennisLee: I thought we needed to clean up objects manually from memory after we're done with them?

I was experiencing lots of problems while using iPython to test code. I even discovered that if I ran tests in iPython, closed and then reopened iPython without changing any of the implementation or test code, and reran the tests again, I would get differing test results. At that point I decided to manually delete objects to avoid any data persistence issues and to run the tests using python.exe from the normal command line (instead of iPython). This was one of a few reasons why this assignment took me so long to complete.

students/DennisLee/lesson07/test_html_render.py
natasha-aleksandrova: I see, yes I'm not a fan of aliases generally 🙂

students/DennisLee/lesson07/test_html_render.py
natasha-aleksandrova: No you don't need to do that. There had to be something else going on causing side effects in the tests.

PR #509, opened 2018-05-04
Lesson 6 Unit tests for mailroom

itsabhister:


students/itsabhi/Lesson6/mailroom4v1.py
```diff
@@ -0,0 +1,99 @@
+#! /usr/bin/env python3
+
+
+def list_of_donors(donor_hist):
+    donor_list = []
+    for k in donor_hist:
+        donor_list.append(k)
+    return (donor_list)
```
robalford: You don't need the parens here

students/itsabhi/Lesson6/mailroom4v1.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def sending_thanksv2(donor_hist):
+    # If already exists, add donation amt to history, else update records as a new donor & then print thank you letter.
+    donor = input("Type the Full Name of the donor here or type List to display current donors : ")
+    if donor == "list":
+                for k in donor_hist:
+                    print(k)
+                donor = input("Type the Full Name of the donor here: ")
+    if donor in donor_hist.keys():
```
robalford: `in` checks for key membership in `dict`

students/itsabhi/Lesson6/mailroom4v1.py
```diff
@@ -0,0 +1,99 @@
...
+                break
+    else:
+        while True:
+            amount = input("Enter Donation value: ")
+            try:
+                donor_hist.update({donor: [int(amount)]})  #Convert donation value from string to integer
+            except ValueError:
+                print("Invalid donation entered..")
+            else:
+                break
```
robalford: Can you think of a way to make to avoid the double `while True` loops and make this section more DRY?

students/itsabhi/Lesson6/test_mailroom.py
```diff
@@ -0,0 +1,45 @@
...
+class DonationsTestCases(unittest.TestCase):
+
+    def setUp(self):
+    #Invoked before each test method
+        self.test_donors_dict = {
+        'A G': [10, 20, 30],
+        'N D': [40, 50]
+    }
+    def test_list_donors(self):
+    #Test donor list when the option "list" is selected under the Sending Thanks menu
```
robalford: There are a few Pep8 violations in this file, I would recommend using a code linter

students/itsabhi/Lesson6/test_mailroom.py
```diff
@@ -0,0 +1,45 @@
...
+        result = mailroom4v1.list_of_donors(self.test_donors_dict)
+        self.assertEqual(list(self.test_donors_dict.keys()), result)  # Convert to list and assert against result
+
+    @patch('mailroom4v1.input', side_effect = ['PJ Mask', 200])
+    def test_sending_thanks_add_donor(self, test_donors_dict):
```
robalford: The extra parameter here is for the patched input function and should probably be named something like `mock_input`. See this for more details on how you might want to use this parameter: https://stackoverflow.com/a/16053625

PR #510, opened 2018-05-05
Assignments_L01-L03

rgpag:
submitting assignments up through L03, not including L03 mailroom.


students/rgpag/lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,9 @@
+for i in range(1,101):
+	if i%3 ==0 and i%5==0:
```
robalford: Please check your code for Pep8 (Python style guide) violations. Following the style conventions is a best practice that makes your code more readable for other developers. You can paste your code into an online [Pep8 checker](http://pep8online.com) or configure your editor to flag Pep8 violations as you type.

students/rgpag/lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,9 @@
+for i in range(1,101):
+	if i%3 ==0 and i%5==0: 
+		print('FizzBuzz')
+	elif i%3==0:
+		print('Fizz')
+	elif i%5==0:
+		print('Buzz')
+	else:
+		print(i)
```
robalford: It's very important in Python that you configure your editor to use 4 space characters when you hit the tab key. A Pep8 linter will catch this as well, and there are instructions in the course materials for configuring this setting in Sublime Text.

students/rgpag/lesson02/GridPrinter.py
```diff
@@ -0,0 +1,62 @@
...
+	print(space_pipe+4*space_space+space_pipe+4*space_space+space_pipe)
+	print(space_pipe+4*space_space+space_pipe+4*space_space+space_pipe)
+	print(space_pipe+4*space_space+space_pipe+4*space_space+space_pipe)
+	print(space_pipe+4*space_space+space_pipe+4*space_space+space_pipe)
+
+	#print bottom of grid
+	print(space_plus+4*space_dash+space_plus+4*space_dash+space_plus)
+
+#Part2 n-sized 2x2 grid
+import math
```
robalford: The convention is for `import` statements to go at the top of the file

students/rgpag/lesson02/series.py
```diff
@@ -0,0 +1,72 @@
...
+	elif n==1:
+		return 1
+
+	else:
+		for i in range(n-1):
+			fibC=fibL1+fibL2
+			fibL2=fibL1
+			fibL1=fibC
+		return fibC
+		'''
```
robalford: It's a good idea to remove commented code before you commit. If you commit each small change to your code, you can always go back and look at the older version if you want to

students/rgpag/lesson02/series.py
```diff
@@ -0,0 +1,72 @@
...
+			lucL2=lucL1
+			lucL1=lucC
+		return lucC
+	'''
+	return sum_series(n,2,1)
+
+def sum_series(n,v1=0,v2=1):
+	"""sum series(n,v1=0,v2=1) - return "n"th value of sum series having v1 and v2 as first and second values"""
+	L1=v2
+	L2=v1
```
robalford: It's helpful to use meaningful variable names for something like this to make your code more readable

students/rgpag/lesson03/list_lab.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env
+
+#series1
+series1 = ["Apples","Pears","Oranges","Peaches"]
+def series_1():
+	print(series1)
+	response = [input("Please enter another fruit > ")]
+	series1=series1+response
```
robalford: Redefining this name in the local namespace of the function is causing an `UnboundLocalError` in the print call on line 6. I would suggest naming the global variable something like `fruits` to avoid this

students/rgpag/lesson03/list_lab.py
```diff
@@ -0,0 +1,77 @@
...
+	for i in range(len(series1)):
+		x=series1[i]
+		if x[0]=="P":
+			print(x)
+
+#series2
+def series_2():
+	series2=series1
+	print(series2)
+	del series2[len(series2)-1]
```
robalford: You can use negative index here to access the last item in the list

students/rgpag/lesson03/list_lab.py
```diff
@@ -0,0 +1,77 @@
...
+	while response2 in series22:
+		series22.remove(response2)
+	print(series22)
+
+#series3
+def series_3():
+	series3=series1
+	print(series3)
+	responsei=list(range(len(series3)))
+	for i in range(len(series3)):
```
robalford: This function could be simplified a lot by looping over a copy of the list and removing the items from the original list as you loop

students/rgpag/lesson03/list_lab.py
```diff
@@ -0,0 +1,77 @@
...
+		response3=input("Do you like " + series3[i].lower() + "? ")
+		while response3 not in ("yes","no"):
+			response3=input("Please respond 'yes' or 'no', do you like "+ series3[i].lower()+"? ")
+		responsei[i]=response3
+	print(responsei)
+	for j in range(len(series3)):
+		if responsei[j]=="no":
+			del series3[j]
```
robalford: This throws an `IndexError` if you answer 'no' to each item as it's currently written. See my previous comment for a strategy to avoid this

students/rgpag/lesson03/list_lab.py
```diff
@@ -0,0 +1,77 @@
...
+	print(series3)
+
+#series4
+def series_4():
+	series4 = series1[:]
+	for i in range(len(series4)):
+		temp = series4[i]
+		temp2=temp[::-1]
+		series4[i]=temp2
+		print(temp2)
```
robalford: This can be simplified to something like this:
```
reversed_fruits = []
for fruit in series4:
    reversed_fruits.append(fruit[::-1])
```
Also note that you can use the `enumerate` function if you need access to both items and their indexes in the loop.

students/rgpag/lesson03/slicingLab.py
```diff
@@ -0,0 +1,52 @@
+def exchange_first_last(seq):
+	""" return seq with first and last elements exchanged"""
+	L=len(seq)
+	return seq[-1]+seq[1:L-1]+seq[0]
```
robalford: The call the `len` is necessary since you can use negative index in your slice. This code will work to concatentate a string but for other sequences it will throw a `TypeError` if you try to add, for example, an integer and a list

students/rgpag/lesson03/slicingLab.py
```diff
@@ -0,0 +1,52 @@
...
+
+	a_string = "this is a string"
+	a_tuple = (1,2,3,4,5,6)
```
robalford: This code is unreachable because it's inside the function block after the `return` statement, you want to place your assert statements outside of the function at the global module level. You'll see a failing assert for the error I mentioned above

students/rgpag/lesson03/slicingLab.py
```diff
@@ -0,0 +1,52 @@
...
+
+	assert rem_odd(a_string) == 'ti sasrn'
+	assert rem_odd(a_tuple) == (1,3,5)
+
+def rem_ends_odds(seq):
+	"""return seq with first and last 4 elements dropped, and dropping odd elements of middle"""
+	temp_seq=seq[4:-4]
```
robalford: This could be done as a one liner without the need for the other function `seq[4:-4:2]`

robalford:
@rgpag I don't see the Lesson 3 assignment submissions on the course website. Remember to submit them there as well so we can grade.

PR #512, opened 2018-05-05
Finished Lesson2 assignments

seaneightysix:
My Lesson2 assignments are ready for review. Thank you.


students/Sean_Tasaki/Lesson2/FizzBuzz.py
```diff
@@ -0,0 +1,15 @@
+for i in range(1,101):
+	if (i % 3 == 0) and not (i % 5 == 0):
```
robalford: The parentheses are not necessary for the conditional statements here

students/Sean_Tasaki/Lesson2/FizzBuzz.py
```diff
@@ -0,0 +1,15 @@
...
+		print("Fizz")
+	
+	elif (i % 5 == 0) and not (i % 3 == 0):
+		print("Buzz")
+	
+	elif (i % 5 == 0) and (i % 3 == 0):
```
robalford: If you place this statement in the first `if` block, you can remove the `and not` clauses from the the other statements, since you will have already checked for numbers divisible by both 3 and 5

students/Sean_Tasaki/Lesson2/GridPart1.py
```diff
@@ -0,0 +1,24 @@
+def gridPrinter():
```
robalford: In Python the convention is to use `snake_case` for variable and function names

students/Sean_Tasaki/Lesson2/GridPart3.py
```diff
@@ -0,0 +1,15 @@
+def gridPatternPart3(rc, s):
```
robalford: It's a good idea for readability to use meaningful names for function parameters

students/Sean_Tasaki/Lesson2/series.py
```diff
@@ -0,0 +1,83 @@
...
+        return 1
+    if n == 0:
+        return 0
+    if n < 0:
+        print("Invalid data")
+    else:
+    	return fibonacci(n - 2) + fibonacci(n - 1)
+
+def lucas(n):
+	"""Return the nth value in the Lucas numbers series (starting with zero index.
```
robalford: Looks like the indentation level switches from 4 spaces to tabs here. Be sure to always use 4 spaces for indentation, you can configure your editor to enter four spaces automatically when you hit the tab key

students/Sean_Tasaki/Lesson2/series.py
```diff
@@ -0,0 +1,83 @@
...
+    # Calculates lucas series
+	elif (index0 == 2 and index1 == 1):
+		if n == 1:
+			return 1
+		if n == 0:
+			return 2
+		if n < 0:
+			print("Invalid data")     
+		else:
+			return lucas(n - 2) + lucas(n - 1)
```
robalford: The goal for the `sum_series` function is to default to fibonacci, but to be able to generate a new sequence based on any two input parameters (not just the lucas series). You can do this by using your variable parameter values (`index0` and `index 1`) in place of the hard coded values that you used for the first two items of the other sequences.

robalford:
@seaneightysix be sure to submit your assignments on the course website as well so we can grade. See the 'Assignment Submission Area' at the bottom of the assignment page on the website for instructions.

PR #513, opened 2018-05-05
Step 7-8 PR

johnpharmd:


students/johnpharmd/lesson08/circle_class.py
```diff
@@ -0,0 +1,54 @@
+import math
+
+
+class Circle:
+    """represents a simple circle, using properties, a class method, and
+    special methods"""
+
+    @classmethod
```
robalford: It would be helpful to add a docstring explaining what the purpose of this method is.

students/johnpharmd/lesson08/circle_class.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __str__(self):
+        return 'Circle with radius: ' + str(self.radius)
+
+    def __repr__(self):
+        return 'Circle(' + str(self.radius) + ')'
+
+    def __add__(self, other):
+        self.radius += other.radius
+        return self.__repr__()
```
robalford: These methods should return new circle objects, rather than changing the radius value of the circle that calls them

students/johnpharmd/lesson08/circle_class.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __mul__(self, value):
+        self.radius *= value
+        return self.__repr__()
+
+    def __rmul__(self, value):
+        return self.__mul__(value)
+
+    def __gt__(self, other):
+        return self.radius > other.radius
```
robalford: If you use the `functools.total_ordering` decorator from the python standard library, you only need to define `__eq__` and `__lt__`, otherwise you should define all of the comparison operators

students/johnpharmd/lesson08/circle_class.py
```diff
@@ -0,0 +1,54 @@
...
+    @diameter.setter
+    def diameter(self, value):
+        self.radius = value / 2
+
+    @property
+    def area(self):
+        return math.pi * self.radius ** 2
+
+    def from_diameter(self):
+        return self.diameter
```
robalford: The idea here is to let you instantiate a new circle object with only a diameter, so this should be a class method that derives the radius from the diameter and instantiates a new circle

PR #520, opened 2018-05-06
Lesson 7

BrandonHenson:
Finished Lesson 7


students/BrandonHenson/Lesson7/html_render.py
```diff
@@ -0,0 +1,107 @@
...
+class Element:
+    tag = ''
+    indent = ' '
+
+    def __init__(self, content=None, **attributes):
+        self.attributes = attributes
+        if content:
+            self.content = [content]
+        else:
+            self.content = []
```
robalford: This could be a one-liner:
```
```

students/BrandonHenson/Lesson7/test_html_render.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(test_element.content, ["some stuff"])
+
+    def test_append(self):
+        test = hr.Element('some stuff in here')
+        test.append('some more stuff in here')
+        self.assertEqual(test.content, ['some stuff in here', 'some more stuff in here'])
+
+    def test_render(self):
+        element = hr.Element("Test")
+        element.tag = 'html'
```
robalford: The tag attribute is already set to 'html' for all instances of the `HTML` class in your class definition, so it would make more sense to instantiate an `HTML` object rather than override the tag attribute on the `Element` instance

students/BrandonHenson/Lesson7/test_html_render.py
```diff
@@ -0,0 +1,69 @@
...
+        f = StringIO()
+        element.render(f)
+        self.assertEqual(f.getvalue(), "<html>\n" + element.indent + "Test\n</html>\n")
+        element.tag = 'body'
```
robalford: Same thing here: this is overriding the tag attribute on the `Element` instance instead of instantiating a new `Body` instance which would already have the tag attribute set to 'body'

students/BrandonHenson/Lesson7/test_html_render.py
```diff
@@ -0,0 +1,69 @@
...
+        element.render(f)
+        self.assertEqual(f.getvalue(), "<head>\n" + element.indent + "Test\n</head>\n")
+        assert os.path.isfile(cwd + '\\' + 'test_html_output1.html')
+        assert os.path.isfile(cwd + '\\' + 'test_html_output2.html')
+        assert os.path.isfile(cwd + '\\' + 'test_html_output3.html')
+        assert os.path.isfile(cwd + '\\' + 'test_html_output4.html')
+        assert os.path.isfile(cwd + '\\' + 'test_html_output5.html')
+        assert os.path.isfile(cwd + '\\' + 'test_html_output6.html')
+        assert os.path.isfile(cwd + '\\' + 'test_html_output7.html')
+        assert os.path.isfile(cwd + '\\' + 'test_html_output8.html')
```
robalford: It's usually better to use one of the unittest assert methods rather than Python assert statement if you can find one that works.

students/BrandonHenson/Lesson7/test_html_render.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_init(self):
+        element = hr.Element('test', keyword='this is a test')
+        self.assertEqual(element.content, ['test'])
+        self.assertEqual(element.attributes, {'keyword': 'this is a test'})
+
+    def test_selfclosingtag(self):
+        test = hr.SelfClosingTag()
+        test.tag = 'meta'
```
robalford: Here you should instantiate `Meta`, and then you don't need set the tag attribute, because it's already set for all instances of the `Meta` class, same thing as mentioned above

PR #523, opened 2018-05-07
Lesson 08 Complete

stuartkershaw:


students/stuartkershaw/lesson08/circle.py
```diff
@@ -0,0 +1,59 @@
...
+        self._radius = val / 2
+
+    @property
+    def area(self):
+        return self.radius * self.radius * pi
+
+    @classmethod
+    def from_diameter(cls, val):
+        self = cls(val / 2)
+        return self
```
natasha-aleksandrova: should be inline `return cls(val / 2)`

students/stuartkershaw/lesson08/test_circle.py
```diff
@@ -0,0 +1,100 @@
...
+
+    assert d.radius == 4
+    assert d.diameter == 8
+
+
+def test_circle_add():
+    c1 = Circle(2)
+    c2 = Circle(4)
+    c1 + c2
+    assert Circle(6)
```
natasha-aleksandrova: you should be asserting equality to something here

students/stuartkershaw/lesson08/test_circle.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def test_circle_mul():
+    c1 = Circle(4)
+    c1 * 3
+    assert Circle(12)
```
natasha-aleksandrova: you should be asserting equality to something here

PR #527, opened 2018-05-08
Assignments for Lesson 4

TrkCat:
Ready for review. Thanks!


students/TrkCat/Lesson04/mailroom.py
```diff
@@ -0,0 +1,141 @@
...
+    don_dict['num_don'] = len(don_dict['donations'])
+    don_dict['avg_don'] = don_dict['tot_don']/don_dict['num_don']
+
+
+def show_menu(prompt, disp_dict):
+    """Generate menu with dispatch dictionary"""
+    while True:
+        sel = input(prompt)
+        if disp_dict[sel]() is False:
+            return False
```
natasha-aleksandrova: since you're not consuming return value from this function, you can just do `return` or `break`

students/TrkCat/Lesson04/trigram_text.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def trigram_text(infile, out_len):
+    """Generate trigram text of length out_len from text file infile."""
+    punc = ['.', '!', '?', '"', ',', '$', '-']
+    tri_dict = gen_trigram_dict(infile, punc)
+    out_text = []
+    word_count = 0
+    while word_count < out_len:
+            sent_len = random.randint(6, 25)
```
natasha-aleksandrova: this block is over-indented (should be 4 spaces)

PR #528, opened 2018-05-08
Lesson 06 Submit

prgrover:


students/prgrover/lesson06/mailroom.py
```diff
@@ -0,0 +1,184 @@
...
+
+    sorted_donor_dict = OrderedDict(sorted(donor_dict.items(), key=lambda i: sum(i[1]), reverse=True))
+    
+    for (donor, donations) in sorted_donor_dict.items():
+        donor_name = donor
+        total_given = sum(donations)
+        num_donations = len(donations)
+        average_donation = (total_given / num_donations)
+        print ("{:30s}  ${:11.2f}   {:12d}    ${:12.2f}".format(donor_name, total_given, num_donations, average_donation))
+
```
robalford: You could have this function return the report as a string to be printed, rather than printing as a side effect of the function, to make it testable

students/prgrover/lesson06/mailroom_test.py
```diff
@@ -0,0 +1,26 @@
...
+    def test_list_donors(self):
+        self.assertEqual(list_donors(),'Harry Potter\nRonald Weasley\nHermione Granger\nDraco Malfoy\nNeville Longbottom\nAlbus Dumbledore')
+
+    def test_add_donation(self):
+        add_donation('Harry Potter', 12345.67)
+        self.assertEqual(12345.67,donor_dict["Harry Potter"][-1])
+
+    def test_create_email_file(self):
+        create_email_file()
+        self.assertTrue('Draco Malfoy.txt', os.path.exists('Draco Malfoy.txt'))
```
robalford: It would be nice to test that the email contains the correct content as well

PR #530, opened 2018-05-08
L03 Cleanup + mailroom Pt 1

rgpag:
cleaned up strformat_lab, list_lab, slicingLab per instructor feedback. Completed mailroom pt 1 for instructor review.

### DIFF
--- a/students/rgpag/lesson03/mailroom.py
+++ b/students/rgpag/lesson03/mailroom.py
@@ -1,0 +1,76 @@
+# init lists
+donors = ['Mike', 'Joe', 'Kyle', 'Nick', 'Sid']
+i_amount = [3000, 20, 40607, 65, 400]
+i_count = [4, 2, 3, 6, 7]
+msg = """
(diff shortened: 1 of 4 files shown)


students/rgpag/lesson03/mailroom.py
```diff
@@ -0,0 +1,76 @@
...
+    i_count[i] += 1
+    return donors[i], i_count[i], i_amount[i]
+
+
+def getKey(item):
+    return item[1]
+
+
+if __name__ == "__main__":
+    def mailRoom():
```
natasha-aleksandrova: where is this function called? this should be lower case -> `mailroom`

students/rgpag/lesson03/mailroom.py
```diff
@@ -0,0 +1,76 @@
...
+        while 1 == 1:
```
natasha-aleksandrova: should be `while True`

PR #532, opened 2018-05-09
Lesson05

JeffBennett1920:


students/JeffBennett/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,157 @@
...
+        print(key)
+
+
+def input_gift():
+    """input donor gift."""
+    while True:
+        try:
+            gift = float(input("\nEnter amount of gift: " +
+                               "or '0' to return to the main menu: "))
+            return gift
```
robalford: You generally only want to include the code in the try block that might raise the error, so I'd remove the return statement and place it after the except block at the end of the function

PR #533, opened 2018-05-09
Added lesson01 files.

RandyKline:


natasha-aleksandrova:
Hi there!

students/RandyKline/lesson01/break_me.py
```diff
@@ -0,0 +1,27 @@
...
+#print("This call raises type error")
+#typeErrFunc(myInt = "Randy", myStr = float([0,1,2,3]))
+print("The following function raises Syntax error")
+def attFunc(empname, emprole):
+    print ("Emp Name: ", empname)
+    print ("Emp Role: ", emprole)
+    print ("Attribute: ", emprole.sister)
+    return;
+print("The following raises the AttributeError")
+attFunc(empname = "Randy", emprole = "Master")
```
natasha-aleksandrova: recommend to install pep8 linter to catch style violations, for example: function names and var names should be all lower case

RandyKline:
One other question. I don't see my directory in the students directory on
the repository. Is that something that should update when I push from my
local repository?

Thanks for approving my work.

RandyKline:
I guess I need to develop patience. I see it now.

PR #535, opened 2018-05-10
Lesson 04, Trigram edits; Lesson 05, Mailroom Part 3

DeFeNdog:


students/carlos_novoa/lesson05/mailroom.py
```diff
@@ -0,0 +1,252 @@
...
+        if dispatch[r]() == "break":
+            break
+
+
+def main_menu():
+    menu_selection(main_prompt, main_dispatch)
+
+
+def donors_sub_menu():
+    menu_selection(donors_prompt, donors_dispatch)
```
natasha-aleksandrova: it would be good to move vars like `donors_prompt` and `donors_dispatch` into this function since they are only used here and more readable.

students/carlos_novoa/lesson05/mailroom.py
```diff
@@ -0,0 +1,252 @@
...
+            if not donor_name:
+                raise ValueError('String required.')
+        except ValueError as e:
+            print(e)
+            continue
+        else:
+            if donor_name in _donors:
+                print('Donor already exists, use append.')
+                return
+            _donors[donor_name] = []
```
natasha-aleksandrova: keep in mind that there is a `defaultdict` available where you can define values to default to a list

PR #536, opened 2018-05-10
Lesson 9 Mailroom

alexLaws:


students/alexLaws/lesson09/mailroom.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def add_donor(self, *args):
+        for arg in args:
+            self.donor_dictionary.append(arg)
+
+    def check_donor(self, name):
+        donor_exists = False
+        for donor in self.donor_dictionary:
+            if name == donor.name:
+                donor_exists = True
```
natasha-aleksandrova: you can do `return True` here and line 74 `return False`

students/alexLaws/lesson09/run_mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+    new_donor = True
+    for donor in donor_list_obj.donor_dictionary:
+        if name == donor.name:
+            donor.new_donation(amount)
+            temp = donor
+            new_donor = False
+    if new_donor:
+        temp = Donor(name, [amount])
+        donor_list_obj.add_donor(temp)
+    return temp
```
natasha-aleksandrova: should this use some of your functions?
```
donor = donor_list_obj.get_donor(name)
if donor:
     donor.new_donation(amount)
else:
    donor = Donor(name, [amount])
    donor_list_obj.add_donor(donor)
return donor
```

PR #537, opened 2018-05-10
First submission of Lesson 9 Mailroom OOO

wesmoy17:


students/WesleyM/Lesson9/mailroom.py
```diff
@@ -0,0 +1,120 @@
+class donor:
```
robalford: Remember to use CamelCase for class names

students/WesleyM/Lesson9/mailroom.py
```diff
@@ -0,0 +1,120 @@
...
+    def __init__(self, name, donation = None):
```
robalford: No space around the equal sign for keyword args per Pep8

students/WesleyM/Lesson9/mailroom.py
```diff
@@ -0,0 +1,120 @@
...
+        if not donation:
```
robalford: It would be safer to go with `if donation is None`

students/WesleyM/Lesson9/mailroom.py
```diff
@@ -0,0 +1,120 @@
...
+            donation = []
+        self.name = name
+        self.donation = donation
```
robalford: This assignment should go in an else clause above, or could be done as a one-liner:
```
self.donation = donation if donation is not None else []
```

students/WesleyM/Lesson9/mailroom.py
```diff
@@ -0,0 +1,120 @@
...
+        self.total = sum(self.donation)
+        if donation:
+            self.average = self.total/len(self.donation)
+        else:
+            self.average = 0
```
robalford: It's recommended to use `@property` for calculated values

students/WesleyM/Lesson9/mailroom.py
```diff
@@ -0,0 +1,120 @@
...
+    def add_donation(self, amount):
+        self.donation.append(amount)
+        self.total += amount
+        self.average = self.total / len(self.donation)
+
+class DonorDatabase:
+
+    def __init__(self, name):
```
robalford: Remember to set `name=None` if you want the default value to be an empty list

students/WesleyM/Lesson9/mailroom.py
```diff
@@ -0,0 +1,120 @@
...
+    def get_all_donor_names(self):
+        return [donor.name for donor in self.donors]
+        
+    def sort_donors(self):
+        return sorted(self.get_all_donor_names())
+    
+    def donor_input(self):
+        return input("Enter a donor name or input 'List' for a list of donors\n>")
+    
+    def send_thankyou(self):
```
robalford: Ideally, you would separate all the data handling code from the code that deals with user input, so rather than make it a `DonorDatabase` method, I'd move this function to the module level

students/WesleyM/Lesson9/mailroom.py
```diff
@@ -0,0 +1,120 @@
...
+                d1 = donor(don_input, [donation])
+                self.add_donor(d1)
+            except ValueError:
+                print("Please enter a name")
+        else:
+            for d in self.donors:
+                if d.name == don_input:
+                    d.add_donation(donation)
+
+        print("Thank you {} for your donation of ${:.2f}".format(d1.name, d1.total))
```
robalford: This breaks for existing donors, since you only assign the `d1` variable if it's a new donor

students/WesleyM/Lesson9/test_mailroom.py
```diff
@@ -0,0 +1,31 @@
...
+        donor3 = mr.donor("Rosa", [200])
+        database = mr.DonorDatabase([donor1, donor2, donor3])
+        self.assertEqual(database.donors[0].name, donor1.name)
+        self.assertEqual(database.donors[2].name, donor3.name)
+        
+        donor4 = mr.donor("Charles", [20])
+        database.add_donor(donor4)
+        self.assertEqual(database.donors[3].name, donor4.name)
+        self.assertEqual(database.donors[3].donation, donor4.donation)
+
```
robalford: Ideally you would have a test for every method defined on your classes.

PR #544, opened 2018-05-11
Finished Lesson 8

BrandonHenson:
I have finished lesson 8 (circle.py and test_circle.py)

### DIFF
--- a/students/BrandonHenson/Lesson8/circle.py
+++ b/students/BrandonHenson/Lesson8/circle.py
@@ -1,0 +1,57 @@
+# Brandon Henson
+# 5\11\18
+# Lesson 8
+# Circles
+
+import math
+
+circles = []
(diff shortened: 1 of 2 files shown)


students/BrandonHenson/Lesson8/circle.py
```diff
@@ -0,0 +1,57 @@
...
+        return "Circle({})".format(self.radius)
+
+    def __str__(self):
+        return "Circle with Radius: {}".format(self.radius)
+
+    def __add__(self, circle2):
+        return "Circle({})".format(self.radius + circle2.radius)
+
+    def __mul__(self, other):
+        return "Circle({})".format(self.radius * other)
```
robalford: These operations should return new circle objects rather than string representations

PR #545, opened 2018-05-12
Final Series

julienjass:


students/julienjass/Lesson02/print_grid.py
```diff
@@ -0,0 +1,50 @@
+#Part 1
+
+def makingGrid():
```
natasha-aleksandrova: per pep8 guidelines, should be snake_case -> `making_grid`

julienjass: Yes, I did submit it on the course website as well.

natasha-aleksandrova:
ok, didn't see it earlier but I see it now.

PR #550, opened 2018-05-12
Lesson3 assignments ready for review.

seaneightysix:
Lesson3 assignments ready for review.

### DIFF
--- a/students/Sean_Tasaki/Lesson04/mailroom_part2.py
+++ b/students/Sean_Tasaki/Lesson04/mailroom_part2.py
@@ -1,7 +1,7 @@
 """
 Sean Tasaki
-5/11/2018
-Lesson03.mailroom_part_1
+5/12/2018
+Lesson04.mailroom_part_2
 """
 
 donor_full_list = [['Bob Dylan', 120000.00, 3, 40000.00], ['Leonard Cohen', 600.00, 2, 300.00], ['Soren Kierkegaard', 1000.00, 2, 500.00], ['Italo Calvino', 20000.00, 4, 5000.00], ['Florence Feist', 300.00, 3, 100.00]]
--- a/students/Sean_Tasaki/Lesson04/mailroom_part2.py
+++ b/students/Sean_Tasaki/Lesson04/mailroom_part2.py
@@ -11,16 +11,20 @@ for x in donor_full_list:
 
 
 def main_menu():
-	response = input("Enter 1-3 from the following options: (1) Send a Thank You, (2) Create a Report, (3) Quit\n >")
-	if response == '1':
-		thank_you()
-	elif response == '2':
-		create_report()
-	elif response =='3':
-		exit()
-	else:
-		print("Invalid data. Enter 1-3 from the following options: (1) Send a Thank You, (2) Create a Report, (3) Quit\n> ")
-		main_menu()
+	main_menu_dict = {'1': thank_you, '2': create_report, '3': thank_you_letters, '4': quit}
+	main_prompt = "Enter 1-4 from the following options: (1) Send a Thank You to a Donor, (2) Create a Report, (3) Write a Thank You Letter to All Donors (4) Quit\n >> "
+	main_menu_response(main_prompt, main_menu_dict)
+
+def main_menu_response(prompt, main_menu_dict):
+	response = input(prompt)
+	if response not in main_menu_dict:
+			print("Invalid data")
+			main_menu()
+	while True:
+		main_menu_dict.get(response)()
+		if main_menu_dict[response]() == "exit menu":
+			exit()
+		
 
 def thank_you():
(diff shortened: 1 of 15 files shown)


students/Sean_Tasaki/Lesson3/strformat_lab.py
```diff
@@ -0,0 +1,55 @@
...
+
+def task_two():
+    results = ('file_0002', 123.46, 1.00e+04, 1.23e+04)
+    print("Task Two results:")
+    print(f'{results[0]:s} :  {results[1]:.2f}, {results[2]:.2e}\, {results[3]:.2e}')
+
+
+def task_three(tuple):
+    list = []
+    for i in tuple:
```
natasha-aleksandrova: don't use `list` or `tuple` for var names, you are overwriting a python built-in

students/Sean_Tasaki/Lesson3/slicing_lab.py
```diff
@@ -0,0 +1,49 @@
...
+# exchange first and last items in sequence
+def exchanged_first_last(seq):
+	first = seq[:1]
+	last = seq[-1:]
+	middle = seq[1: -1]
+	return (last + middle + first)
+
+# remove every other item in sequence
+def removed_every_other(seq):
+	return (seq[::2])
```
natasha-aleksandrova: don't need parens, can just be `return seq[::2]`

students/Sean_Tasaki/Lesson3/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+	response = input("Enter 1-3 from the following options: (1) Send a Thank You, (2) Create a Report, (3) Quit\n >")
+	if response == '1':
+		thank_you()
+	elif response == '2':
+		create_report()
+	elif response =='3':
+		exit()
+	else:
+		print("Invalid data. Enter 1-3 from the following options: (1) Send a Thank You, (2) Create a Report, (3) Quit\n> ")
+		main_menu()
```
natasha-aleksandrova: this would be better with a `while True` loop so that you are not calling `main_menu` in various places

students/Sean_Tasaki/Lesson3/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+
+def thank_you():
+	name = input("Enter the first and last name of the donor or enter 'list' to see a list of previous donor names or enter Q to exit to main menu\n> ")
+	if name.lower() == 'list':
+		print(donor_name_list)
+		thank_you()
+	elif name.upper() == 'Q':
+		main_menu()
+	elif name.title() in donor_name_list:
+		for x in donor_full_list:
```
natasha-aleksandrova: it is better to name your loop var something meaningful like `for donor in donor_full_list`

`x` isn't very clear or readable when reading the code

common exception is using one letter vars in list comprehensions

students/Sean_Tasaki/Lesson3/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+		amount = input(f"{name.title()} is a new donor. Enter the donation amount.\n> ")
+		if amount.lower() == 'q':
+			thank_you()
+		else:
+			donor_name_list.append(name.title())
+			new_list = ['null', 0.00, 0, 0.00]
+			new_list[0] = (name.title())
+			new_list[1] = (float(amount))
+			new_list[2] += 1
+			new_list[3] = (float(amount))
```
natasha-aleksandrova: there is not reason to "pre-fill", can be simplified to:
`donor_full_list.append([name.title(), float(amount), 1, float(amount)])`

students/Sean_Tasaki/Lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,93 @@
+"""
+Sean Tasaki
+5/12/2018
+Lesson04.mailroom_part_2
+"""
+
+donor_full_list = [['Bob Dylan', 120000.00, 3, 40000.00], ['Leonard Cohen', 600.00, 2, 300.00], ['Soren Kierkegaard', 1000.00, 2, 500.00], ['Italo Calvino', 20000.00, 4, 5000.00], ['Florence Feist', 300.00, 3, 100.00]]
+donor_name_list = []
+for x in donor_full_list:
+	donor_name_list.append(x[0])
```
natasha-aleksandrova: the main donor data structure should be converted to a dictionary as well

natasha-aleksandrova:
@seaneightysix I saw your email about indentation.

The reason it shows overindented is because it using tabs, not *spaces*.

in addition to setting tab size, you also need to set it to convert tabs to spaces. try this config:
```
"translate_tabs_to_spaces": true,
"tab_size": 4,
"indent_to_bracket": true,
"detect_indentation": false
```

PR #551, opened 2018-05-12
Assignment 05 submission

jorauk:


students/JonathanMauk/lesson05/mailroom03.py
```diff
@@ -0,0 +1,148 @@
+import os
+import sys
+
+donor_names = ["John Smith", "Jane Doe", "Alan Smithee", "Tom D.A. Harry", "Joe Shmoe"]
+donation_amounts = [[18774.48, 8264.47, 7558.71], [281918.99, 8242.13], [181.97, 955.16], [67.10, 500.98], [200.01]]
+
+donor_db = {name: donation for name, donation in zip(donor_names, donation_amounts)}
+# donor_db = {"John Smith": [18774.48, 8264.47, 7558.71], "Jane Doe": [281918.99, 8242.13],
```
robalford: This works for getting in practice with comprehensions, but in reality would probably be more appropriate to define as dict literal.

students/JonathanMauk/lesson05/mailroom03.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def mailroom():
+    """Generate main menu options and activate other functions."""
+    while True:
+        selection = input('MAILROOM v0.3\n------------------------\nChoose an option:\n1) Send a thank you' +
+                          '\n2) Create a report\n3) Send letters to everyone\n4) Quit\n> ')
+        menu_dict = {'1': thank_you, '2': report, '3': thank_all, '4': quit_program}
+        try:
+            menu_dict.get(selection)()
```
robalford: An alternative would be to use the bracket syntax to access dict values and then catch a `KeyError` for bad input

robalford:
@jorauk remember to submit the assignment on the Course website as well so we can grade.

PR #555, opened 2018-05-13
lesson09; please advise on next steps

johnpharmd:


natasha-aleksandrova:
@johnpharmd could you clarify your comment about advising on next steps please?

students/johnpharmd/lesson09/mailroom_OO.py
```diff
@@ -0,0 +1,120 @@
...
+                sum(self.donations), 'num_donations': len(self.donations)}}
+
+    @property
+    def title(self):
+        return self.new_title
+
+    @title.setter
+    def title(self, new_title):
+        """enables title change for donor"""
+        self.new_title = new_title
```
natasha-aleksandrova: if you're not doing anything special with an attribute then you don't need a property (it is also discouraged to defined self attributes outside of `__init__` method like `self.new_title`)
you can access title directly.

students/johnpharmd/lesson09/mailroom_OO.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @property
+    def donation(self):
+        """getter for most recent donation amount"""
+        return self.donations[-1]
+
+    @donation.setter
+    def donation(self, donation):
+        """enables addition of new donation"""
+        self.donations.append(donation)
```
natasha-aleksandrova: this would probably be more clear as a `add_donation` method

johnpharmd:
@natasha-aleksandrova, sorry; meant additional required functionality and code changes

students/johnpharmd/lesson09/mailroom_OO.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class DonorGroup(Donor):
```
natasha-aleksandrova: this class shouldn't extend from `Donor`

`Donor` represents a single donor object (which you have defined very well above)
`DonorGroup` should be a collection of donor objects and any action on that. Think of Dog is a Pet relationship that makes sense for inheritance, but DonorGroup is Donor doesn't really make sense.

students/johnpharmd/lesson09/mailroom_OO.py
```diff
@@ -0,0 +1,120 @@
...
+        while True:
+            print('\nMain Menu:')
+            response = input(self.main_text)
+            print()
+            try:
+                if response == 'q':
+                    print('Program execution completed.')
+                self.menu_dict[response](self)
+            except KeyError:
+                print('\nThat selection is invalid. Please try again.')
```
natasha-aleksandrova: the main program CLI does not need to be a class, you just need to represent a single donor and a collection of donors as objects.

for your menu switch, you could do something like this:
```
donor_group = DonorGroup()  # pass whatever is required into constructor
def send_thank_you():
    # your function to handle user input and interact with your class objects
menu_dict = {"1": thank_you, "2": create_report}
```

students/johnpharmd/lesson09/testmailroomOO.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python3
+import unittest
+import mailroom_OO as mrOO
```
natasha-aleksandrova: note that modules should be lower case

PR #558, opened 2018-05-14
lesson08 submission

natrod22:


students/NatalieRodriguez/Lesson08/Circle.py
```diff
@@ -0,0 +1,116 @@
...
+
+#imports
+
+import math
+
+
+class Circle:
+
+    def __init__(self, in_radius=1):
+        self._radius = in_radius
```
robalford: A leading underscore conventionally denotes a 'protected' attribute in Python, and I don't think you want to indicate that here. It's also conventional to use the same name for the input parameter and instance attribute that it sets in the constructor.

students/NatalieRodriguez/Lesson08/Circle.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __str__(self):
+        return "Circle with radius: "+str(self.radius)
```
robalford: String formatting is generally preferred to concatenation when dynamically building strings

students/NatalieRodriguez/Lesson08/Circle.py
```diff
@@ -0,0 +1,116 @@
...
+    def from_diameter(cls, in_diameter):
+        return cls(in_diameter / 2)
+
+    @property
+    def radius(self):
+        return self._radius
+
+    @radius.setter
+    def radius(self, value):
+        self._radius = value
```
robalford: You only need to use getters and setters when you're adding some custom logic, so these are not needed for your `radius` attribute

students/NatalieRodriguez/Lesson08/Circle.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @property
+    def diameter(self):
+        return 2 * self._radius
+
+    @diameter.setter
+    def diameter(self, value):
+        self._radius = value / 2
```
robalford: This, however, is a perfect use case for a getter `@property` and setter method, since you need to add custom logic related to the `radius` value upon setting, or initially accessing the `diameter` attribute

PR #559, opened 2018-05-14
Finishing Lesson04 Assignments

TinaBrownfield:
Adding and Updating Lesson Assignments.


students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+"""
+$ chmod +x mailroom_part2.py
+"""
+import datetime
+import os
+
+# list of donors
+donors_list_dictionary = [{"firstname": "William", "lastname": "Gates", "donations": [326892.24, 122, 22, 12], "donate_times": [4]},
+                          {"firstname": "Mark", "lastname": "Zuckerberg", "donations": [
+                              30, 60, 65982.55], "donate_times": [3]},
```
robalford: I don't think you need this "donate_times" item, see my note below

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+                          {"firstname": "Jeff", "lastname": "Bezos",
+                              "donations": [52636.27], "donate_times": [1]},
+                          {"firstname": "Paul", "lastname": "Allen",
+                              "donations": [877.33, 22], "donate_times": [2]},
+                          {"firstname": "Steven", "lastname": "Hawking", "donations": [
+                              326892.24, 123, 123.33, 123, 123], "donate_times": [5]},
+                          {"firstname": "Justin", "lastname": "Timberlake", "donations": [999658.25, 1233, 123], "donate_times": [3]}]
+
```
robalford: I would recommend using a dict as your main data structure with full names as keys, rather than a list of dicts. This should simplify a lot of your code for looking up and adding donors and donations.

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+        response = input(prompt)
+        if dispatch_dict[response]() == "exit menu":
+            break
+
+
+def single_print_sub_menu():
+    menu_selection(single_print_sub_prompt, single_print_sub_dispatch)
+
+
+def quit():
```
robalford: This function is overwriting the Python built-in `quit()` function (see how GitHub highlights the text in the same color as built-in `print` below?). I'd recommend naming something that doesn't clash with the built in

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+def get_name_input():
+    # Function to select user input to return to print function
+    user_input = input("\nLookup donor by: \n"
+                       "\t1. Firstname: \n"
+                       "\t2. Lastname:\n"
+                       "Please enter 1 or 2 >>  ")
+
+    donor_return = []
+    donor_check = 0
+    if int(user_input) == 1:
```
robalford: Why not just reference the string '1'?

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+        fname = input(
+            "Type the first name of the donor you are looking for:  ")
+    else:
```
robalford: Probably better to go with `elif user_input == '2'` here, since this will grab any input other than '1'. You could have an additional `else` clause to handle bad input

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+        lname = input("Type the last name of the donor you are looking for:  ")
+
+    for donor in donors_list_dictionary:
+        # print(donor["firstname"].lower())
```
robalford: It's always a good idea to remove 'dead' code before committing to git

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+        if int(user_input) == 1:
+            if fname.lower() in donor["firstname"].lower():
```
robalford: Careful here, this will jump to the else clause if the first name doesn't match for the first donor item in the dict without checking if it matches any of the other donors.

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+                last_name = input(
+                    "No donor with that first name found in donor list. Please provide a last name to add donor to list: ")
+                donor_return.append(first_name)
+                donor_return.append(last_name)
+                break
+
+        else:  # user input = 2
+            if lname.lower() in donor["lastname"].lower():
+                donor_check = input(
+                    "Is this the donor you are looking for:  {} {}?   Please enter yes or no:  ".format(donor["firstname"], donor["lastname"]))
```
robalford: Watch line length for Pep8 here. It's acceptable to add line break at opening parens like this:
```
donor_check = input(
                    "Is this the donor you are looking for:  {} {}?   Please enter yes or no:  ".format(
                        donor["firstname"], donor["lastname"]))
```

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+                    break
+                else:
+                    last_name = lname
+                    first_name = input(
+                        "No donor with that last name found in donor list. Please provide a first name for addition to donor list")
+                    donor_return.append(first_name)
+                    donor_return.append(last_name)
+                    break
+
+    return donor_return, donor_check
```
robalford: Why not return the entire donor item along with their donation history so you don't have to look the donor up again in the `send_single_thank_you` function? Also, I don't think `donor_check` isn't needed since this function should always either return a new or existing donor

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+    donor_name, donor_check = get_name_input()
+    donor_amount = float(input('Please enter a donation amount for {} {}: '.format(
+        donor_name[0], donor_name[1])))
+
+    # search through donors with returned name - 1 means name is in list, 0 means not in list
+    if donor_check == 1:
+        for donor in donors_list_dictionary:
+            if donor["firstname"] == donor_name[0] and donor["lastname"] == donor_name[1]:
+                donor["donations"].append(donor_amount)
+                donor["donate_times"]+=1
```
robalford: `donor["donate_times"]` is a list, so you'd need to access the item that you want to increment, but you don't appear to be using the "donate_times" value anywhere in your code so I would just get rid of it and use `len("donations")` where you need the total number of donations

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+    print()
+
+
+def print_thank_you(donor_fname, amount):
+    """ prints thank you message"""
+    d = {'name': donor_fname, 'donation': amount}
+    thank_you = '\nDear {},\n '.format(donor_fname)
+    thank_you += '\tThank you for your generous donation of ${donation:,.2f}\n'.format(
+        **d)
+    thank_you += 'Sincerely, \nThe ChickTech Donations Department\n'
```
robalford: It's generally preferred to use string formatting on the entire string (like you're doing in the `print_thank_you_total` function below) rather than concatenate

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+    return thank_you
+
+
+def print_thank_you_total(donor):
+    """ prints thank you message"""
+    #d = {'name': donor_fname, 'donation': amount}
+    donor_output = {"firstname": donor["firstname"],"lastname": donor["lastname"] }
```
robalford: Why create a separate data structure here when you can refer to the `donor` values directly?

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+    all_donations = donor["donations"]
+    donor_output['last_donation'] = all_donations[len(all_donations) - 1]
```
robalford: Remember that you can use negative indexing to access the last item in a list

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+    total = 0
+    for gift in all_donations:
+        total += gift
```
robalford: You can use the built in `sum` function to sum all items in list

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+    print()
+    # Creating list to hold donors info for printing
+    donor_list = list()
+    for donor in donors_list_dictionary:
+        donor_fullname = "{firstname} {lastname}".format(**donor)
+        # donor object will hold fullname, donation total, donation times, average donation
+        donor_info = [donor_fullname, 0, 0, 0]
+        for donor_amount in donor["donations"]:
+            donor_info[1] += donor_amount
+            donor_info[2] += 1
```
robalford: See built in functions `len` and `sum` for a more concise way to do this

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+    print()
+
+
+def send_letters_everyone():
+    """Creates a letter for everyone in the database, and writes them to file."""
+    letters_count = 0
+    new_folder = str(datetime.datetime.now())
+    try:
+        os.mkdir(new_folder)
+    except OSError as oserr:
```
robalford: You don't need to cast this as `oserr` unless you're going to use the `osser` variable in your `except` block

students/TinaB/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,231 @@
...
+        print("\nError with directory creation.Something must have gone wrong!\n")
+        return
+    for donor in donors_list_dictionary:
+        # create file in date folder titled with donor name
+        filename = "./{dir}/{firstname}_{lastname}.txt".format(**donor, dir=new_folder)
+        with open(filename, 'w') as donor_thanks:
+            letter_output = print_thank_you_total(donor)
+            donor_thanks.write(''.join(letter_output))
```
robalford: Don't need to call join, since `letter_output` is already a string

robalford:
@TinaBrownfield your Lesson 4 mailroom assignment looks good. I made some suggestions to incorporate as you're working on the next version of the assignment for lesson 5. The biggest thing is that you should be using a dict for the main data structure at this point, rather than a list of dicts and that will simplify a lot of the other code in your implementation.

I started reviewing your Kata Fourteen assignment here, however it doesn't appear to be complete at this point. Were you hoping to get some suggestions on it, or should we merge this PR, and you can resubmit that assignment when it's finished?

robalford:
@TinaBrownfield I see that you submitted the Kata Fourteen assignment on the course website, so I will go ahead and review it for you here

students/TinaB/lesson04/katafourteen.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+
+import random
+
+# book_text_file = "sherlock.txt"
+#"sherlock_small.txt"
+text_file = open("sherlock_small.txt", 'r')
```
robalford: It's recommended to use `with` statement for file reading and writing, and it will automatically close the file for you when you exit the `with` block

students/TinaB/lesson04/katafourteen.py
```diff
@@ -0,0 +1,63 @@
...
+text = text_file.read()
+# first option takes text as is, second
+words_in_text = text.split()
+#words_in_text = text.replace('--', ' ').replace('.', '').replace(',', '').replace(
+#    '"', '').replace("'", '').replace(":", ' ').replace('?', '').replace('!', '').replace('\n', ' ').replace('(', '').replace(')', '').split()
```
robalford: It's a best practice to remove all commented code from your files before committing to git. I had assumed this assignment was still a WIP partly because of all the commented code.

students/TinaB/lesson04/katafourteen.py
```diff
@@ -0,0 +1,63 @@
...
+words_in_text = [str(i) for i in words_in_text]
```
robalford: You don't need this comp since the values are already strings

students/TinaB/lesson04/katafourteen.py
```diff
@@ -0,0 +1,63 @@
...
+text_file.close()
+# print(words_in_text)
+
+
+def create_trigram(trigram_text):
+    '''Creates the trigram dictionary from text argument when called'''
+    trigram_dictionary = {}
+    # populates trigram by running a loop
+    for tri_word in range(0, len(trigram_text)-2):
```
robalford: `range` function starts at `0` index by default so you don't need to pass in the `0`

students/TinaB/lesson04/katafourteen.py
```diff
@@ -0,0 +1,63 @@
...
+        # setdefault checks to see if key exists in dictionary and if not sets it with value of []
+        trigram_dictionary.setdefault(
+            '{} {}'.format(trigram_text[tri_word], trigram_text[tri_word+1]), [])
+
+        trigram_dictionary['{} {}'.format(trigram_text[tri_word], trigram_text[tri_word+1])
+                           ].append('{}'.format(trigram_text[tri_word+2]))
```
robalford: You can chain the `append()` method to the `set default()` method on the previous line to make this a bit more concise

students/TinaB/lesson04/katafourteen.py
```diff
@@ -0,0 +1,63 @@
...
+        # creates and prints trigram dictionary
+        trigram_dictionary = create_trigram(words_in_text)
+
+        # creates the output
+        starting_point = random.choice(list(trigram_dictionary))
+        # print(starting_point)
+        word_one,word_two = starting_point.split()
+        output_kata = []
+        # print(len(trigram_dictionary)-1)
+        word_key = word_one + ' ' + word_two
```
robalford: String formatting is generally preferred to concatenation in Python for something like this

students/TinaB/lesson04/katafourteen.py
```diff
@@ -0,0 +1,63 @@
...
+        print(word_key)
+        print(len(trigram_dictionary)-1)
```
robalford: If these print statements are here for debugging, you should go ahead and remove those before committing as well

students/TinaB/lesson04/katafourteen.py
```diff
@@ -0,0 +1,63 @@
...
+        for word in range(0, len(trigram_dictionary)-1):
+            if word_key in trigram_dictionary:
+                output_kata.append(trigram_dictionary[word_key])
```
robalford: This should append the `word_key` word pair to `output_kata`, rather than the value, which is the entire list of following words

students/TinaB/lesson04/katafourteen.py
```diff
@@ -0,0 +1,63 @@
...
+
+                next_word = random.choice(trigram_dictionary[word_key])
```
robalford: After appending the previous word pair, you should append the randomly selected following word before combining with the last word to form a new pair as you're doing below

students/TinaB/lesson04/katafourteen.py
```diff
@@ -0,0 +1,63 @@
...
+                word_one = word_two
+                word_two = next_word
+                word_key = word_one + ' ' + word_two
+
+        print(" ".join(str(r) for v in output_kata for r in v))
+
```
robalford: This is way too much code to include in a single try block, you should only include the code that you think might raise the specific error that you're catching.

robalford:
@TinaBrownfield you're off to a good start with the trigrams assignment. Take a look at my notes and if you time see if you can incorporate them, but we will go ahead and grade this as passing so that you can move on to the next assignment.

TinaBrownfield:
@robalford - Thank you for all of your comments. I have worked hard to incorporate all of the mailroom recommendations into my lesson05 assignment. Maybe if you are grading that one you can (hopefully) see that. 
I will also work on fixing my trigrams assignment - after I catch up on my other assignments as I am still behind. :)

PR #561, opened 2018-05-14
Fizz-Buzz added

mcbraunie:
Created and submitted Fizz-Buzz program from lesson 02.


students/MicahBraun/Lesson 02/Series.py
```diff
@@ -0,0 +1,89 @@
...
+#--------------------------------- Display ----------------------------
+print(header)
+print()
+
+
+print("                        ----  Fibonacci Numbers ---- \n")
+fib_number = int(input("Input a value to see the Fibonacci series at index n (your entry): "))
+for i in range(fib_number):
+    print(fibonacci(i))
+    print()
```
natasha-aleksandrova: it's a good practice to use asserts to verify your code, for example:
```
assert fibonacci(9) == 34
```
this will become even more relevant when we do unit testing in lesson 6

PR #562, opened 2018-05-14
Completed Lesson06 exercises. Mailroom.py has been updated to keep user input in __main__ while utilizing the defined functions for code work only, making them much easier to test.

letoui:


students/ian_letourneau/Lesson06/mailroom.py
```diff
@@ -0,0 +1,169 @@
...
+    dictionary. Finally, the script prints a thank you letter to that donor.
+    To quit to the main menu, 'q' can be typed at any user prompt."""
+
+    # Once amount is input, add the donation to the donor key if existing
+    # or add both the donor and the donation to the dictionary if new
+    while True:
+        try:
+            amount = float(amount)
+        except ValueError:
+            raise ValueError
```
natasha-aleksandrova: if you raise an exception as soon as you catch it, will it ever get to lines 32, 33?

students/ian_letourneau/Lesson06/mailroom.py
```diff
@@ -0,0 +1,169 @@
...
+                "Please enter the amount that was donated('q' to quit"
+                " to menu): ")
+            if amount == 'q':
+                continue
+            # Call the send Thank You latter function passing user input
+            send_thank_you(name, amount)
+        elif response == '2':
+            create_report()
+        elif response == '3':
+            send_all()
```
natasha-aleksandrova: you should be using a dict switch to navigate to main functions (that was part of Lesson 4).

example:
```
main_menu = {"1": thank_you, "2": create_report, "3": exit_menu}
main_menu[response]()
```

students/ian_letourneau/Lesson06/mailroom.py
```diff
@@ -0,0 +1,169 @@
...
+    if name in donors.keys():
+        donors[name].append(amount)
+    else:
+        donors.setdefault(name, [])
+        donors[name].append(amount)
+
+    print('\nDear {}, we wanted to say thank you for your generous'
+          ' donation of ${:.2f}!'.format(name, amount))
+    return('Dear {}, we wanted to say thank you for your generous'
+          ' donation of ${:.2f}!'.format(name, amount))
```
natasha-aleksandrova: you want to avoid adding returning value in a void function just for the sake of testing.

students/ian_letourneau/Lesson06/test_mailroom.py
```diff
@@ -0,0 +1,53 @@
...
+def test_1():
+    assert send_thank_you("Angelina Jolie", 3.47) == ('Dear Angelina Jolie,'
+                                                      ' we wanted to say thank'
+                                                      ' you for your generous'
+                                                      ' donation of $3.47!')
+
+# Test the send_thank_you function: long name
+
+
+def test_2():
```
natasha-aleksandrova: you should avoid naming your functions with vague names like this, when you get a good amount of tests debugging becomes difficult since names mean nothing.

your test names should have some indication what it is testing, for example
```
def test_send_thank_you_two_param(): pass
def test_send_thank_you_three_param(): pass
```

students/ian_letourneau/Lesson06/test_mailroom.py
```diff
@@ -0,0 +1,53 @@
...
+
+def test_4():
+    with pytest.raises(ValueError):
+        assert send_thank_you('Angelina Jolie', 'donation')
+
+# Test the create_report function
+
+
+def test_5():
+    assert create_report() == ('Report has been created.')
```
natasha-aleksandrova: this test isn't testing that report was created correctly, it is only testing that the function run to the end without error. 

one option is to have `create_report` and separate `display_report` functions, where create_report returns a string representation of report and display_report just prints it. that way you can test create_report function to ensure correct report is generated.

students/ian_letourneau/Lesson06/test_mailroom.py
```diff
@@ -0,0 +1,53 @@
...
+
+# Test the send_all function
+
+
+def test_6():
+    assert send_all() == ('Letters have been sent to all donors.')
```
natasha-aleksandrova: same comment as for test_5, you can assert that files were in fact created and even assert that contents of the file is what you expect (`os.path.isfile` check)

PR #566, opened 2018-05-15
Lesson 08: circle.py and test cases

prgrover:


students/prgrover/lesson08/circle.py
```diff
@@ -0,0 +1,48 @@
...
+        self.radius = diameter / 2
+
+    @property
+    def area(self):
+        return self.radius**2 * math.pi
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        cls.radius = diameter / 2
+        return cls(cls.radius)
```
natasha-aleksandrova: you would just return `return cls(diameter / 2)`

students/prgrover/lesson08/circle_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+import unittest
+import math
+import random
+from circle import Circle
+
+class CircleTest(unittest.TestCase):
+
+    def test_init_radius(self):
+        my_radius = 10
```
natasha-aleksandrova: it's common to use `expected` and `actual` vars in tests for clarity

PR #567, opened 2018-05-15
[DennisLee][L8][Circle class] Ready for submission.

DennisLee-DennisLee:


students/DennisLee/lesson08/circle_class.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        Set the radius of the circle. Only nonnegative numbers are
+        allowed.
+
+        :value:  The new radius value.
+        """
+        if value < 0:
+            raise ValueError(f"Negative radius of '{value}' not allowed.")
+        else:
+            self._radius = value
```
robalford: It's great that you're thinking about how to handle faulty user input, but I would use a single instance attribute for `radius` and do the error handling directly in the `__init__` method, rather than using a getter and setter. The approach you're taking ends up giving your class two separate instance attributes for `radius` and `_radius`, which could be a bit confusing for the user. 

It's also worth noting that the leading underscore on `_radius` conventionally denotes a 'protected' attribute (only available for modification in a Class and its subclasses) and I'm not sure if that's what you want to indicate here. Generally, this convention is used fairly sparingly in Python, as it is meant to be an 'open access' language that permits developers to make their own decisions about how they want to use and/or modify their objects.

It's also conventional in Python to define all instance attributes in the `__init__` method, and some linters (like PyCharm's) will complain if you try to define an instance attribute outside of `__init__`, as you're doing for `_radius`. I believe this convention is mainly for code readability, so that other devs can quickly scan the object constructor and know what instance attributes are available.

PR #568, opened 2018-05-15
Andrew Park Lesson09

nDruP:
Completed lesson 9 mailroom. I was unable to find a way to incorporate multiple inheritance into this project so I hope I can get some practice in a future lesson.

### DIFF
--- a/students/nDruP/lesson09/donor.py
+++ b/students/nDruP/lesson09/donor.py
@@ -1,0 +1,31 @@
+class Donor():
+    def __init__(self, title, gift):
+        self._name = title
+        self._history = []
+        if type(gift) == int or type(gift) == float:
+            self._history.append(round(gift, 2))
+        elif type(gift) == list:
+            for x in gift:
+                self._history.append(round(x, 2))
+
+    def __len__(self):
+        return len(self._history)
+    
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def history(self):
+        return self._history
+
+    @property
+    def avg_gift(self):
+        return sum(self._history)/len(self._history)
+
+    @property
+    def sum_gift(self):
+        return sum(self._history)
+
+    def add_gift(self, value):
+        self._history.append(round(value,2))
--- a/students/nDruP/lesson09/donor.py
+++ b/students/nDruP/lesson09/donor.py
@@ -27,5 +27,28 @@ class Donor():
     def sum_gift(self):
         return sum(self._history)
 
+    @property
+    def info(self):
+        return (self.name, len(self.history), self.avg_gift, self.sum_gift)
+    
     def add_gift(self, value):
         self._history.append(round(value,2))
+
+
+    def thank_u_letter_str(self, new_gift = False):
+        divider = "\n" + "*" * 50 + "\n"
+        thank_u_letter = divider + f"Dearest {self.name},\n"
+        if new_gift:
+            thank_gift = self.history[len(self.history)-1]
+            thank_u_letter += "\tThank you for your most recent donation of"
+        else:
+            thank_gift = self.sum_gift
+            thank_u_letter += "\tThank you for your lifetime total of "
+        thank_u_letter += (f" ${thank_gift:.2f}!\n"
+                           "We will use your donation(s) to create real "
+                           "living Pokemon.\n"
+                           "You now have our eternal loyalty. Use it wisely.\n"
+                           "Sincerely,\n"
+                           f"We're a Pyramid Scheme & so is {self.name}"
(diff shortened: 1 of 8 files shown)


students/nDruP/lesson09/donor.py
```diff
@@ -0,0 +1,54 @@
+class Donor():
+    def __init__(self, title, gift):
+        self._name = title
+        self._history = []
```
robalford: The leading underscore denotes a 'protected' attribute (only available for modification by class and subclass) which is not the case here, since these are being accessed and modified in `DonorDict`

students/nDruP/lesson09/donor.py
```diff
@@ -0,0 +1,54 @@
...
+    def __len__(self):
+        return len(self._history)
+    
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def history(self):
+        return self._history
```
robalford: You don't need to use a `@property` for these attributes since you aren't doing any extra work beyond defining them it's more appropriate to set them directly in the constructor

students/nDruP/lesson09/donor_dict.py
```diff
@@ -0,0 +1,159 @@
+from donor import Donor
+
+class Donor_Dict():
+    def __init__(self, donor, *args):
+        self._dict = {donor.name.lower(): donor}
+        for x in args:
+            self._dict[x.name.lower()] = x
+
+    def __contains__(self, donor):
+        return (donor.lower() in self._dict.keys()) or (donor in self.donors)
```
robalford: `in` checks for key membership in a dict

students/nDruP/lesson09/donor_dict.py
```diff
@@ -0,0 +1,159 @@
...
+                    donor_list.append(Donor(donor_init[0], history))
+                end = text.tell()
+        return cls(first_donor, *donor_list)
+    
+    @property
+    def names(self):
+        d_names = []
+        for x in self.donors:
+            d_names.append(x.name)
+        return d_names
```
robalford: These could be one liners:
```
return [x.name for x in self.donors]
```

students/nDruP/lesson09/mailroom.py
```diff
@@ -0,0 +1,223 @@
...
+import datetime
+from donor import Donor
+from donor_dict import Donor_Dict
+
+
+d = Donor_Dict.from_file("dict_init.txt")
+divider = "\n" + "*" * 50 + "\n"
+
+
+def main_menu(user_prompt = None):
```
robalford: No space around equal sign in parameter assignment for Pep8

students/nDruP/lesson09/mailroom.py
```diff
@@ -0,0 +1,223 @@
...
+
+def conv_str(conv_str, conv_type=int):
+    """
+    Convert string to given conv_type.
+    If it's unable to convert, return original string.
+    """
+    try:
+        conv_yes = conv_type(conv_str)
+        return conv_yes
+    except:
```
robalford: It's almost always a good idea to hone in on a specific exception like `ValueError`

students/nDruP/lesson09/mailroom.py
```diff
@@ -0,0 +1,223 @@
...
+    Compose and print a thank you letter to the donor for their donation.
+    Return to main
+    """
+    print(divider)
+    print("Let's craft a very personal thank you note for our donor!")
+
+    d_name = input_donor_name()
+    if d_name:
+       gift_amt = input_donor_amt()
+       if gift_amt:
```
robalford: Looks like the indentation is off by a space here

students/nDruP/lesson09/mailroom.py
```diff
@@ -0,0 +1,223 @@
...
+    if write_dir:
+        for donor in d.keys:
+            print(write_txt_to_dir(d[donor].name,
+                                   d[donor].thank_u_letter_str(),
+                                   write_dir))
+        print("Finished writing the letters")
+    return
+
+
+def exit():
```
robalford: This overwrites the Python built-in `exit` function, so I'd suggest naming it something else

students/nDruP/lesson09/run_mailroom.py
```diff
@@ -0,0 +1,5 @@
+from mailroom import *
```
robalford: It's a best practice to avoid `import *`, with few exceptions. Here's a Stack Overflow [post](https://stackoverflow.com/a/2360808) on the reasoning behind this that also includes a link to the official Python docs on the subject of imports.

PR #569, opened 2018-05-15
adding lesson06

iamAndrewKim:


students/AndrewKim/lesson06/mailroom_part4.py
```diff
@@ -0,0 +1,56 @@
...
+    
+    #add donation amount on the history
+    donation_amount = input("Please enter the donation amount: ")
+    try:
+        donor_history[name].append(float(donation_amount))
+        #print thank you
+        print(f'Thank you {name}, for your generous donation of {donation_amount}!') 
+    except ValueError:
+        print("Not a valid number! Please enter a valid number\n")
+
```
robalford: To make this function testable, I would refactor it to return the thank you letter as a string. Then you can print the string to the console wherever the function is called

robalford:
@iamAndrewKim we are going to temporarily grade this assignment as 'Incomplete/Poor', so that we can grade other students' assignments, but we will adjust your grade to 'Complete/Good', as soon as you are able to resubmit the completed assignment.

iamAndrewKim:
hello here is the revised pull request.

https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/597

I cant submit to the UW website directly since it is already graded.

PR #571, opened 2018-05-16
Lesson 2 is complete. Added Grid.py, FizzBuzz.py, and series.py

rhamersky:
Lesson 2 is ready for review.

### DIFF
--- a/students/rhamersky/Lesson_2/FizzBuzz.py
+++ b/students/rhamersky/Lesson_2/FizzBuzz.py
@@ -1,0 +1,23 @@
+# Description: This program will print Fizz if the number is divisible by 3,
+# Buzz if the number is divisible by 5, and FizzBuzz if the number is divisible by 3 & 5.
+# Developer: Ryan Hamersky
+# Date: 05/01/2018
+# Rev: A - 05/15/2018 add comment to the code.
+
+# -----Presentation Section-----
+
+for num in range(1, 101):
+    # Is the number divisible by 3 and 5?
+    if num % 3 == 0 & num % 5 == 0:
+        print("FizzBuzz")
+    # Is the number divisible by 3?
+    elif num % 3 == 0:
+        print("Fizz")
+    # Is the number divisible by 5?
+    elif num % 5 == 0:
+        print("Buzz")
+    else:
+        print(num)
+    num =+ 1
+
+
--- a/students/rhamersky/Lesson_2/Grid.py
+++ b/students/rhamersky/Lesson_2/Grid.py
@@ -1,0 +1,54 @@
+# Description: This program will create a grid.
+# Developer: Ryan Hamersky
+# Date: 04/29/2018
+# Rev: A - 05/15/2018 add comment to the code.
+
+# -----Data Section-----
+
+# -----Process Section-----
+
+def row(row, units):
+    '''
+    Creates a row (+ & -) for the grid.
+    :param row: How many rows the user would like for the grid.
+    :param units: The width of the columns.
+    :return: Returns gridrow which is a row of + and -.
+    '''
+    gridrow = "+ "
+    for row in range(0, row):
+        for unit in range(0, units):
+            gridrow = gridrow + "- "
+        gridrow = gridrow + "+ "
+    return gridrow
+
+def column(column, units_pipe):
+    '''
+    Creates the column edges with the pipe "|" character.
+    :param column: How many columns the user would like for the grid.
+    :param units_pipe: The width of the columns.
+    :return: Returns gridcolumn which is a column edge using the pipe "|" character.
+    '''
+    gridcolumn = "|"
+    for column in range(0, column):
+        for unit in range(0, units_pipe * 2 + 1):
+            gridcolumn = gridcolumn + " "
+        gridcolumn = gridcolumn + "|"
+    return gridcolumn
+
+
+# -----Presentation Section-----
+
+if __name__ == '__main__':
+
+    # User input
+    intRowCloumn = int(input("How many rows and columns for your grid? "))
+    intUnits = int(input("How many units for your grid cell? "))
+
+    # Creates grid without the bottom row.
+    for i in range(0, intRowCloumn):
+        print(row(intRowCloumn, intUnits))
+        for j in range(0, intUnits):
+            print(column(intRowCloumn, intUnits))
+
+    # Prints bottom row to complete grid
+    print(row(intRowCloumn, intUnits))
--- a/students/rhamersky/Lesson_2/series.py
+++ b/students/rhamersky/Lesson_2/series.py
@@ -1,0 +1,102 @@
+# Description: This program returns the value from the fibonacci and lucus series.
+# Developer: Ryan Hamersky
+# Date: 05/01/2018
+# Rev: A - 05/15/2018 add comment to the code.
+
+# -----Presentation Section-----
+
+def fibonacci(n):
+    '''
+    Returns a number from the fibonacci series.
+    :param n: Number input
+    :return:
+    '''
+    number1 = 0
+    number2 = 1
+    counter = 0
+
+    while True:
+        newnumber = number1 + number2
+        if n is counter:
+            print(number1)
+            break
+        else:
+            number1 = number2
+            number2 = newnumber
+            counter = counter + 1
+            continue
+    return
+
+fibonacci(6)  # --> Testing fibonacci()
+
+def lucus(n):
+    '''
+    Returns a number from the fibonacci series.
+    :param n: Number input
+    :return:
+    '''
+    number1 = 2
+    number2 = 1
+    counter = 0
+
+    while True:
+        newnumber = number1 + number2
+        if n is counter:
+            print(number1)
+            break
+        else:
+            number1 = number2
+            number2 = newnumber
+            counter = counter + 1
+            continue
+    return
+
+lucus(7)  # --> Testing lucus()
+
+def sum_series(n,number1 = 0,number2 = 1):
+    '''
+    Returns a number from the fibonacci or lucus series depending on the parameter inputs number1
+    and number 2.
+    :param n: Number inout from the user.
+    :return:
+    '''
+    counter = 0
+    while True:
+        newnumber = number1 + number2
+        if n - 1 is counter:
+            print(number1)
+            break
+        else:
+            number1 = number2
+            number2 = newnumber
+            counter = counter + 1
+            continue
+    return
+
+number = int(input("Please, enter a number greater than 0. Type 0 to exit. "))
+while number > 0:
+    sum_series(number,2,1)
+    number = int(input("Please, enter a number greater than 0. Type 0 to exit. "))
+    continue
+
+def fibonacci_new(n):
+    '''
+    Returns a number from the fibonacci series.
+    :param n: Number input from the user.
+    :return:
+    '''
+    sum_series(n)
+    return
+
+fibonacci_new(6)  # --> Testing fibonacci_new()
+
+def lucus_new(n):
+    '''
+    Returns a number from the lucus series.
+    :param n: Number input from the user.
+    :return:
+    '''
+    sum_series(n,2,1)
+    return
+
+lucus_new(7)  # --> Testing lucus_new()


rhamersky:
Lesson 3 is ready for review.

students/rhamersky/Lesson_3/list_lab.py
```diff
@@ -0,0 +1,127 @@
...
+print(l)  # --> Prints new list.
+
+# Asks user for a fruit and inserts the fruit at the beginning of the list.
+l.insert(0,input("Please, add another fruit to the beginning of the list. ").title())
+
+print(l)  # --> Prints new list.
+
+# Prints all the fruits that begin with the letter 'P'.
+for fruits in l:
+    if "P" is fruits[0] or "p" is fruits[0]:
```
natasha-aleksandrova: you could do `if fruits[0].lower() == "p"` or `if fruits.lower().startswith('p')`

again, should be using `==` operator instead of `is`

is checks that 2 arguments refer to the same object, == checks that 2 arguments have the same value.  see:
```
>>> 1000 is 10**3
False
>>> 1000 == 10**3
True
```

you would use `is` when checking for None, True or False -> `var is None`, `var is True`.. and even then it is more pythonic to check for "truthiness" or "falsiness" -> `if var` or `if not var` unless there is specific distinction for a var to be boolean type.

students/rhamersky/Lesson_3/mailroom_part1.py
```diff
@@ -0,0 +1,269 @@
+#!/usr/bin/env python3
+#  Description: This program will allow...........
+# Developer: Ryan Hamersky
+# Date: 05/10/2018
+# Rev: A - 05/15/2018 updated tup_donor_names so it would fit on one page and changed 'is' to '==' when comparing the
+#      variable intChoice.
+
+# -----Data Section-----
+strName = None     # --> User's name
+intOption = int()  # --> User's option.
```
natasha-aleksandrova: make sure to use snake_case var names

students/rhamersky/Lesson_3/mailroom_part1.py
```diff
@@ -0,0 +1,269 @@
...
+tup_donor_names = (["William Gates, III",1,10000.0],["Mark Zuckerberg",1,20000.0],
+                   ["Jeff Bezos",1,1200.0],["Paul Allen",1,5000.0],
+                   ["Mike Scott",1,25.0])  # --> Tuple that stores all donor's information
+
+
+
+# -----Process Section-----
+#  Methods defined in the menu class are related to general menu checks or options based on user inputs.
+
+class menu(object):
```
natasha-aleksandrova: should not be using classes until OO lessons, these should be simple module level functions

students/rhamersky/Lesson_3/mailroom_part1.py
```diff
@@ -0,0 +1,269 @@
...
+# Date: 05/10/2018
+# Rev: A - 05/15/2018 updated tup_donor_names so it would fit on one page and changed 'is' to '==' when comparing the
+#      variable intChoice.
+
+# -----Data Section-----
+strName = None     # --> User's name
+intOption = int()  # --> User's option.
+tup_donor_names = (["William Gates, III",1,10000.0],["Mark Zuckerberg",1,20000.0],
+                   ["Jeff Bezos",1,1200.0],["Paul Allen",1,5000.0],
+                   ["Mike Scott",1,25.0])  # --> Tuple that stores all donor's information
```
natasha-aleksandrova: not sure I really understand the purpose of this tuple yet.. you should use a list data structure as your global DB store, not a tuple which is immutable.

students/rhamersky/Lesson_3/mailroom_part1.py
```diff
@@ -0,0 +1,269 @@
...
+                print(strName + ", not a valid option. \n")
+                continue
+            # Verify that the user selected an option 1 through option 3.
+            else:
+                if 1 <= intChoice <= 3:
+                    break
+                else:
+                    print("Integer selected is not a valid option. "
+                          "Please, enter an integer and corresponds to the options given. \n")
+                    continue
```
natasha-aleksandrova: you don't need the two `continue`

students/rhamersky/Lesson_3/mailroom_part1.py
```diff
@@ -0,0 +1,269 @@
...
+        # Error handles user's option choice.
+        intOption = menu.menuoptions()
+        # User selects option 1. Can display a list of the donors and where the user can add donors and donations.
+        if intOption == 1:
+            print()  # --> Formatting
+            sub_menu.send_thankyou()
+            if menu.menureturn(strName) == "Yes":
+                continue
+            else:
+                break
```
natasha-aleksandrova: you technically only need to check:
```
if menu.menureturn(str_name) != "Yes":
    break
```

students/rhamersky/Lesson_3/mailroom_part1.py
```diff
@@ -0,0 +1,269 @@
...
+        :return: This method does not return anything. Return is just used to clearly show where this method ends.
+        '''
+        bool = True  # --> Used as a flag.
+        donor_name = input("Please, enter full name of donor. ").title()
+
+        # Creates a donor list.
+        while donor_name == "List":
+            if len(tup_donor_names) > 0:  # --> Verifies that there is a list one stored as a list row in the table tuple.
+                donor_list_creation()
+                bool = False  # --> Is used as a flag on the while loop below
```
natasha-aleksandrova: `bool` is a built-in, do *not* overwrite it

students/rhamersky/Lesson_3/mailroom_part1.py
```diff
@@ -0,0 +1,269 @@
...
+    if donor_name in lst_donor_name:
+        index = lst_donor_name.index(donor_name)  # --> Finds index of donor in  list
+        tup_donor_names[index][1] = tup_donor_names[index][
+                                          1] + 1  # --> Increase amount of times donor donated
+        tup_donor_names[index].append(donation_amount)  # --> Adds donation to history
+    # Adds new donor to the donor list
+    else:
+        tup_new_donor = ([donor_name, 1, donation_amount],)
+        tup_donor_names += tup_new_donor
+    return
```
natasha-aleksandrova: don't need the returns in void functions unless you're exiting early (it is not pythonic)

students/rhamersky/Lesson_3/mailroom_part1.py
```diff
@@ -0,0 +1,269 @@
...
+        for row in lst_table:
+            print(("{0:{padding}}".format(row[0], padding=max_length)), end=" ")
+            print(("{0:{padding}}".format(row[1], padding=max_length)), end=" ")
+            print(("{0:{padding}}".format(row[2], padding=max_length)), end=" ")
+            print(("{0:{padding}}".format(row[3], padding=max_length)), end=" ")
+            print()
+        return
+
+
+def donor_history(donor_name, donation_amount):
```
natasha-aleksandrova: it is important for function names to closely reflect what it does, in this case this should be something like `update_donors`

`donor_history` is a bit misleading

students/rhamersky/Lesson_3/mailroom_part1.py
```diff
@@ -0,0 +1,269 @@
...
+        while bool:
+            try:
+                flo_donation_amount = float(input(strName + ", please enter donation amount. "))
+
+                # Updates donor tuple
+                donor_history(donor_name, flo_donation_amount)
+
+                # String formatting for a brief thank-you to donor.
+                print()  # --> Formatting
+                thankyou_message = "{}, thank you for donating ${:,.2f}.".format(donor_name, flo_donation_amount)
```
natasha-aleksandrova: if you want thank you note to have a new line I would recommend adding `\n` to the beginning of the `thankyou_message` to reduce number of code lines

students/rhamersky/Lesson_3/mailroom_part1.py
rhamersky: Natasha,

Once I learned the basic of class in the Python Fundamental class from UW, I got in a habit of using them. I also like my code organized as it helps me find certain areas of my code faster.

If you would really like me to get rid of the classes and make them functions I can in Lesson 6.

Let me know.

Thanks,

Ryan

students/rhamersky/Lesson_3/mailroom_part1.py
rhamersky: Corrected in Lesson 4 and Lesson 5 of the vars that I saw.

students/rhamersky/Lesson_3/mailroom_part1.py
rhamersky: Natasha,

I choose a tuple because it was immutable as I did not want to accidentally change the first element of the table when coding (list[0] = new_element). I only wanted to make the donor information mutable that is why I made the table a tuple and the row a list.

If this is not best practice in the programming field I can work on it, but being relative new to programming I did not want to make a costly mistake like this in this type of program.

Thanks,

Ryan

students/rhamersky/Lesson_3/mailroom_part1.py
rhamersky: Corrected in Lesson 4 and Lesson 5

students/rhamersky/Lesson_3/list_lab.py
rhamersky: Thanks for this. I never really knew between the 'is' and '=='.  

Ah thanks for the tip with the method startswith.

students/rhamersky/Lesson_3/Sequence.py
```diff
@@ -0,0 +1,92 @@
...
+
+def third_mix_up(seq):
+    '''
+    This function rearranges the sequence in the followig order: middle third, last third,
+    and finally first third.
+    :param seq: The seq that is passed into the function.
+    :return: Returns the newly created sequence.
+    '''
+    length = len(seq)  # --> Gets length of sequence.
+    a_new_sequence = seq[length//3:2*length//3]+seq[2*length//3:length]+seq[0:length//3]
```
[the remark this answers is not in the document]
rhamersky: I am not sure what I was thinking with I wrote that line of code. It should have been a variable as you mentioned and yours looks a lot cleaner and easier to read/debug

students/rhamersky/Lesson_3/Sequence.py
```diff
@@ -0,0 +1,92 @@
...
+    :return: Returns the newly created sequence.
+    '''
+    length = len(seq)  # --> Gets length of sequence.
+    a_new_sequence = seq[-1:]+seq[1:length-1]+seq[:1]  # --> Exchanges first and last
+    return a_new_sequence
+
+# Tests to see if the function is working properly.
+print()  # --> Formatting
+print(exchange_first_last(seq))
+print(exchange_first_last(seq1))
```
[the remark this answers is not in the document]
rhamersky: I will have to use assert sometime as I have been looking for better ways to verify code can using print() statements.

students/rhamersky/Lesson_2/Grid.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # User input
+    intRowCloumn = int(input("How many rows and columns for your grid? "))
+    intUnits = int(input("How many units for your grid cell? "))
+
+    # Creates grid without the bottom row.
+    for i in range(0, intRowCloumn):
+        print(row(intRowCloumn, intUnits))
+        for j in range(0, intUnits):
+            print(column(intRowCloumn, intUnits))
```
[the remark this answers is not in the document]
rhamersky: Thanks for the information.

students/rhamersky/Lesson_2/FizzBuzz.py
```diff
@@ -0,0 +1,23 @@
...
+        print("FizzBuzz")
+    # Is the number divisible by 3?
+    elif num % 3 == 0:
+        print("Fizz")
+    # Is the number divisible by 5?
+    elif num % 5 == 0:
+        print("Buzz")
+    else:
+        print(num)
+    num =+ 1
```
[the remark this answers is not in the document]
rhamersky: This is another place where I do not know what I was thinking. I got my for and while loop mixed up and the num =+ 1 was going to be my counter. So, it should not be in there.

students/rhamersky/Lesson_3/mailroom_part1.py
natasha-aleksandrova: `tuple` is a great data structure for something that needs to stay immutable. the donor data structure _is_ mutable as we are modifying it to add more donors.

it is pretty much expected for this to be a list (and a dictionary in next version of mailroom).

students/rhamersky/Lesson_3/mailroom_part1.py
natasha-aleksandrova: `bool_flag` is a good var name 👍

PR #573, opened 2018-05-16
Lesson 7  and 8 commits

tbrack:


students/tbrackney/Lesson7/html_render.py
```diff
@@ -0,0 +1,144 @@
...
+Author: Travis Brackney
+Class: Python 201 - Self paced online
+Date Created 5/12/2018
+Python Version: 3.6.4
+"""
+
+''' '''
+
+
+class Element():
```
robalford: No parens needed for class definition

students/tbrackney/Lesson7/html_render.py
```diff
@@ -0,0 +1,144 @@
...
+    """Class for rendering html page"""
+    tag = None
+    i_string = '    '
+    indent = 0
+    new_lines = True
+    keywords = ''
```
robalford: I don' t think you need to define these last three as class attributes. You don't appear to be using the `indent` class attribute, but rather passing the value in as an instance method parameter to your tag and render methods. The `new_lines` attribute could be a method level variable inside the render method, since you're overriding render in the SelfClosing and OneLine subclasses. And you're setting `keywords` as an instance attribute in the `__init__` method which is more appropriate than making it class level. Remember class attribute are for values that will be shared between all instances of the class.

students/tbrackney/Lesson7/html_render.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def __init__(self, *args, **kwargs):
+        self.content = []
+        if len(args) is not 0:
```
robalford: This could be `if args:`, but it seems like you should be using a single `content` parameter rather than variable `*args`

students/tbrackney/Lesson7/html_render.py
```diff
@@ -0,0 +1,144 @@
...
+                c.render(file_out, (curr_ind + 1))
+            else:
+                file_out.write(c + ' ')
+        file_out.write(self.close_tag(curr_ind))
+
+
+class Html(Element):
+    tag = 'html'
+
+    def render(self, file_out):
```
robalford: Need to add the `curr_ind` parameter here to avoid this:
```
TypeError: render() takes 2 positional arguments but 3 were given
```

students/tbrackney/Lesson7/html_render.py
```diff
@@ -0,0 +1,144 @@
...
+class Br(SelfClosingTag):
+    tag = 'br'
+
+
+class A(OneLineTag):
+    tag = 'a'
+
+    def __init__(self, *args, **kwargs):
+        kwargs['href'] = args[0]
+        Element.__init__(self, *args[1:], **kwargs)
```
robalford: You can call the base class constructor with `super().__init()`

students/tbrackney/Lesson8/circle_class.py
```diff
@@ -0,0 +1,74 @@
...
+Date Created 5/16/2018
+Python Version: 3.6.4
+"""
+import math
+
+
+class Circle():
+    """Implements a class of circle objects"""
+    def __init__(self, the_radius):
+        self._r = the_radius
```
robalford: I would use `self.radius` for the attribute name for readability. The leading underscore denotes a 'protected' attribute and I'm not sure that's what you want to indicate here.

students/tbrackney/Lesson8/circle_class.py
```diff
@@ -0,0 +1,74 @@
...
+            return Circle(self.radius * factor)
+
+    def __rmul__(self, factor):
+        return Circle(self.radius * factor)
+
+    def __eq__(self, other):
+        return (self.radius == other.radius)
+
+    def __lt__(self, other):
+        return (self.radius < other.radius)
```
robalford: Remember to wrap your class with the `@functools.total_ordering` decorator so you only need to define `__eq__` and `__lt__` and not all of the other comparison operators

students/tbrackney/Lesson8/circle_class.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @property
+    def radius(self):
+        return self._r
+
+    @radius.setter
+    def radius(self, val):
+        if val < 0:
+            raise ValueError('Radius must be greater than 0')
+        self._r = val
```
robalford: I would do the error handling in the `__init__` method rather than use getter and setter for this attribute

PR #576, opened 2018-05-16
Lesson6 - Refactored mailroom.py and unit testing

csdotson:
**Notes:** Mailroom required a pretty heavy refactor to allow for unit testing. Aside from printing functions and user input, most functionality is covered by testing. Let me know if you see areas to fix/improve.

**Files**:
- mailroom_4.py
- test_mailroom_4.py

### DIFF
--- a/students/csdotson/lesson06/test_mailroom_4.py
+++ b/students/csdotson/lesson06/test_mailroom_4.py
@@ -4,13 +4,6 @@ Unit testing for mailroom
 
 import mailroom_4 as mr
 import pytest
-import sys
-
-donations = {"Bill Gates": [10.50, 123.45, 1111.11],
-             "Jeff Bezos": [7.65, 1000.00],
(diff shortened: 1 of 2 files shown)


students/csdotson/lesson06/test_mailroom_4.py
```diff
@@ -0,0 +1,79 @@
...
+        "q": mr.quit_menu,
+    }
+
+
+def test_thank_you_menu():
+    assert mr.thank_you_menu == {
+        "1": mr.prompt_for_donation,
+        "2": mr.print_donors,
+        "q": mr.quit_menu,
+    }
```
robalford: I would focus on testing the parts of your program that are dynamically generated, so you probably don't need these last couple of tests since they're really just testing the initial variable assignment in your code

robalford:
@csdotson remember to submit the assignment on the course website as well for grading.

PR #577, opened 2018-05-16
Mailroom4 for Lesson06 & html_render for Lesson07

julienjass:


julienjass:
Also added to this PR: html_render for lesson7

julienjass:
@robalford Thanks. Yes, Lesson 7 is ready. I just submitted it to the class website. Thanks!

students/julienjass/Lesson07/html_render.py
```diff
@@ -0,0 +1,153 @@
...
+class Element:
+    """Base Class of the HTML Render"""
+    tag = 'html'
+    ind = 0
+
+    def __init__(self, content=None, **kwargs):
+        if content is None:
+            self.content = []
+        if content != None:
+            self.content = [content]
```
robalford: This could be a one-liner:
```
self.content = [content] if content is not None else []
```

students/julienjass/Lesson07/html_render.py
```diff
@@ -0,0 +1,153 @@
...
+        if kwargs:
+            self.attrs = kwargs
+        else:
+            self.attrs = None
```
robalford: This could be simplified to this (and would be set to empty dict if there were no kwargs passed in):
```
self.attrs = kwargs
```

students/julienjass/Lesson07/html_render.py
```diff
@@ -0,0 +1,153 @@
...
+        if any(isinstance(line, str) for line in self.content):
+            content_copy = self.content[:]
+            for i in range(len(content_copy)-1):
+                next_index = i + 1
+                if isinstance(content_copy[i], object) and not isinstance(content_copy[i], str):
+                    pass
+                elif not isinstance(content_copy[next_index], str):
+                    self.content.insert(next_index, '\n')
+        if isinstance(self.content[-1], str):
+            self.content.append('\n')
```
robalford: This method seems overly complicated. See my note below for a simpler solution.

students/julienjass/Lesson07/html_render.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def render(self, out_file, **kwargs):
```
robalford: You need to include the indentation level as a positional argument in order for the run_html_render script to work. I"m currently seeing the following error when I try to run the script and generate the html:
```
TypeError: render() takes 2 positional arguments but 3 were given
```

students/julienjass/Lesson07/html_render.py
```diff
@@ -0,0 +1,153 @@
...
+        if self.attrs:
+            attr_string = ''
+            for key, value in self.attrs.items():
+                attr_string += str('{}="{}" '.format(key, value))
+            out_file.write(f'{" "*indent}<{self.tag} {attr_string.strip()}>\n')
+        else:
+            out_file.write(f'{" "*indent}<{self.tag}>\n')
+        for line in self.content:
+            if isinstance(line, str):
+                out_file.write(f'{" "*(indent+4)}{line}')
```
robalford: I think you can get rid of the `add_newlines` method if you add this here (there might be some other places where you'd need the new line characters as well, but seems simpler to add them in `render` where you need them):
```
out_file.write('\n')
```

students/julienjass/Lesson07/html_render.py
```diff
@@ -0,0 +1,153 @@
...
+            elif isinstance(line, object):
```
robalford: Everything is an object in Python so this isn't really doing anything, I would just go with an `else` clause to catch objects that aren't strings

students/julienjass/Lesson07/html_render.py
```diff
@@ -0,0 +1,153 @@
...
+    def render(self, out_file, **kwargs):
+        if 'indent' in kwargs:
+            indent = kwargs['indent']
+        else:
+            indent = self.ind
+        if self.attrs:
+            attr_string = ''
+            for key, value in self.attrs.items():
+                attr_string += str('{}="{}" '.format(key, value))
+            out_file.write(f'{" "*indent}<{self.tag} {attr_string.strip()}>')
```
robalford: There's a lot of repeated code between the different `render` implementations, I would look at refactoring some of this out into a separate method on the base class

robalford:
@julienjass you're off to a great start with html renderer as well. I made some a few notes for you to incorporate if you have the time, but this is ready to merge.

PR #578, opened 2018-05-17
Lesson 10 with the separation of concerns feedback

MikeS-1:
Functional programming mailroom with the separation of concerns feedback from lesson 09 included


students/MikeShand/lesson10/mailroom_fp_2.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+class Roster:
+
+    def __init__(self, donors=None):
+        self.donor_list = donors
+
+    def donor_roster(self):
+
+        ''' Returns a list of the names from the donor objects'''
```
natasha-aleksandrova: note that docstrings use triple double quotes not single double quotes:
```
def donor_roster(self):
    """Returns a list of the names from the donor objects"""
    pass
```

students/MikeShand/lesson10/mailroom_fp_2.py
```diff
@@ -0,0 +1,245 @@
...
+
+        donor_names = []
+        for donor in self.donor_list:
+            donor_names.append(donor.name)
+        return donor_names
```
natasha-aleksandrova: remember those list comprehensions 😉

students/MikeShand/lesson10/mailroom_fp_2.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+        elif max:
+            for donor in money:
+                return list(map(lambda x: x * factor, filter(lambda y: y <= max, money)))
+
+
+        else:
+            for donor in money:
+                return list(map(lambda x: x * factor, money))
```
natasha-aleksandrova: you want to avoid using `min` and `max` as var names, those are built-in functions that are being overwritten here.

PR #579, opened 2018-05-17
Updated mailroom.py with OO principles and generated tests

shibinmathew:
mailroom.py and test_mailroom.py for lesson 9.

### DIFF
--- a/students/shibin_mathew/lesson9/mailroom.py
+++ b/students/shibin_mathew/lesson9/mailroom.py
@@ -1,0 +1,152 @@
+#!/usr/bin/env python3
+import sys
+from collections import defaultdict
+
+
+class Donor:
+    def __init__(self, names, amount):
+        self.names = names
+        self.amount = amount
+
+    @property
+    def names(self):
+        return "{}".format(self._names)
+
+    @names.setter
+    def names(self, names):
+        self._names = names
+
+    @property
+    def amount(self):
+        return self._amount
+
+    @amount.setter
+    def amount(self, amount):
+        self._amount = amount
+
+
+class Collection:
+    def __init__(self, donors2=None):
+        self.donors2 = donors2
+
+    @property
+    def donors2(self):
+        return self._donors2
+
+    @donors2.setter
+    def donors2(self, donors2):
+        self._donors2 = donors2
+
+    def print_donor_names(self, donors2):
+        list_test = []
+        for names in donors2:
+            list_test.append(names)
+        return list_test
+
+    def send_thank_you(self, donors2):
+        #  Function to Send a Thank You Email
+        c1 = Collection()
+        program_quit = False
+        while not program_quit:
+            name = input("Please input your full name\n")
+            if name == "list":
+                print("list of names")
+                print(c1.print_donor_names(donors2))
+            elif name == 'quit' or name == 'Quit':
+                break
+            else:
+                break
+        while True:
+            donation = input("Please enter a donation amount\n")
+            if donation == 'quit' or donation == 'Quit':
+                return
+            else:
+                try:
+                    donation = float(donation)
+                    donors2[name].append(donation)
+                    break  # break out of the while loop
+                except ValueError:
+                    print("please input a valid amount")
+
+    def create_report(self, donors2):
+        #  Function to create a report
+        list_test = []
+        print("hello")
+        print("{: <20s}{: ^4s}{: <10s}{: ^4s}{: <10s}{: ^4s}{: <10s}"
+              .format('Donor Name', '|', 'Total Given', '|',
+                      'Num Gifts', '|', 'Average Gift'))
+        print('-' * 70)
+        print(' ')
+        for names in donors2:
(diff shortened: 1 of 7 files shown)


students/shibin_mathew/lesson9/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+    def names(self, names):
+        self._names = names
+
+    @property
+    def amount(self):
+        return self._amount
+
+    @amount.setter
+    def amount(self, amount):
+        self._amount = amount
```
natasha-aleksandrova: if you are not doing anything special you don't really need to use properties, just regular class attributes should work.

students/shibin_mathew/lesson9/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+        for names in donors2:
+            list_test.append(names)
+        return list_test
+
+    def send_thank_you(self, donors2):
+        #  Function to Send a Thank You Email
+        c1 = Collection()
+        program_quit = False
+        while not program_quit:
+            name = input("Please input your full name\n")
```
natasha-aleksandrova: remember the separation of concern from the assignment page, you should not have any `input` functions in your data handling code

students/shibin_mathew/lesson9/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+        return self._amount
+
+    @amount.setter
+    def amount(self, amount):
+        self._amount = amount
+
+
+class Collection:
+    def __init__(self, donors2=None):
+        self.donors2 = donors2
```
natasha-aleksandrova: what does `donors2` mean?

students/shibin_mathew/lesson9/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+    def create_report(self, donors2):
+        #  Function to create a report
+        list_test = []
+        print("hello")
+        print("{: <20s}{: ^4s}{: <10s}{: ^4s}{: <10s}{: ^4s}{: <10s}"
+              .format('Donor Name', '|', 'Total Given', '|',
+                      'Num Gifts', '|', 'Average Gift'))
+        print('-' * 70)
+        print(' ')
+        for names in donors2:
```
natasha-aleksandrova: the main point of the class is to use the attribute (`self.donors2`), why is this accepting it as a param instead?

students/shibin_mathew/lesson9/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+    d3 = Donor("Jordy", 12)
+    d4 = Donor("Andreck", 14)
+    list_ = [(d1.names, d1.amount), (d2.names, d2.amount), (d3.names, d3.amount), (d4.names, d4.amount)]
+    donors2 = defaultdict(list)
+    for name, amount in list_:
+        donors2[name].append(amount)
+
+    q = Collection()
+    q.donors2 = donors2
+    q.question_module(q.donors2)
```
natasha-aleksandrova: couple points here:
* `Donor` class should have methods that deal with single donor like letter text for example
* Your collection class should deal with multiple donors, so it should have access to your donor objects something like `collection = Collection([Donor("name1", 100), Donor("name2", 200)])`, you should be doing all of the operations on that collection of donors (not creating more instances of collection class within the collection class?)
* actual data structure for donors should be hidden in your `Collection` class (lines 141 - 144)
* input functions should be outside of data handling classes (Donor and Collection)

students/shibin_mathew/lesson9/mailroom.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class Donor:
+    def __init__(self, names, amount):
+        self.names = names
+        self.amount = amount
+
+    @property
+    def names(self):
+        return "{}".format(self._names)
```
natasha-aleksandrova: don't need this as a property, `_names` is already a string so no need to format.

also, it shouldn't be `names`, but a single `name`

students/shibin_mathew/lesson9/mailroom.py
shibinmathew: Good point - I made the change.

students/shibin_mathew/lesson9/mailroom.py
shibinmathew: I created a class to handle the question module which will separate the functions that take in the inputs with the functions that handle the data.

students/shibin_mathew/lesson9/mailroom.py
shibinmathew: passing in a donors list - I would assume in the industry you can access this from a database that holds the donor information.

students/shibin_mathew/lesson9/mailroom.py
shibinmathew: Good point - this was my mistake. I made the change.

shibinmathew:
I refactored a lot of the code to implement your suggestions.

PR #581, opened 2018-05-17
Lesson 4 & Lesson 5

rhamersky:
Lesson 4 assignment 1 and lesson 5 complete.

### DIFF
--- a/students/rhamersky/Lesson_4/mailroom_part2.py
+++ b/students/rhamersky/Lesson_4/mailroom_part2.py
@@ -1,0 +1,335 @@
+#!/usr/bin/env python3
+# Description: This program will allow a user to input a donor and a donation.
+# The user will also be able to send out thank you notes to the donors.
+# Developer: Ryan Hamersky
+# Date: 05/13/2018
+# Rev: A - 05/16/2018 updated code to reflect comments made by the instructor in mailroom_part.py.
+#      Update variables to snake_case. Removed unnecessary 'continues' and 'returns'.
+#      Changed 'bool' variable to 'bool_flag'. Updated misleading function names.
+#      Reduced the amount of unnecessary print() statements.
+
+# -----Data Section-----
+import os
+str_name = None     # --> User's name
+int_option = int()  # --> User's option.
+tup_donor_names = (["William Gates, III",1,10000.0],["Mark Zuckerberg",1,20000.0],
+                   ["Jeff Bezos",1,1200.0],["Paul Allen",1,5000.0],
+                   ["Mike Scott",1,25.0])  # --> Tuple that stores all donor's information
+str_salutation = ""   # --> Email salutation
+str_body = ""         # --> Email body
+str_valediction = ""  # --> Email valediction
+
+# -----Process Section-----
+#  Methods defined in the menu class are related to general menu checks or options based on user inputs.
+
+class menu(object):
+    @staticmethod
+    def menu_options():
+        '''
+        Try and except block used on User's option choice to check for integers.
+        :return: Returns int data (1-4) which is used to determine what menu option will be executed.
+        '''
+        while True:
+            try:
+                int_choice = int(input(str_name + ", please select an option above. "))
+            except ValueError:
+                print(str_name + ", not a valid option. \n")
+            # Verify that the user selected an option 1 through option 4.
+            else:
+                if 1 <= int_choice <= 4:
+                    break
+                else:
+                    print("Integer selected is not a valid option. "
+                          "Please, enter an integer and corresponds to the options given. \n")
+        return int_choice
+############## END OF MENU CLASS ######################
+
+# Methods in the sub_menu class will execute code based off of the user's sub-menu choice.
+
+
+class sub_menu(object):
+    @staticmethod
+    def send_thankyou():
+        '''
+        This method will create a list of donors, allow new donors to be added, store donations, and print a thank you
+        statement to send to the donor by mail.
+        :return:
+        '''
+        bool_flag = True  # --> Used as a flag.
+        donor_name = input("Please, enter full name of donor. ").title()
+
+        # Creates a donor list.
+        while donor_name == "List":
+            # --> Verifies that there is at least a donor is stored as a list row in the table tuple.
+            if len(tup_donor_names) > 0:
+                donor_list_creation()  # --> Creates a list of donors
+                bool_flag = False  # --> Is used as a flag on the while loop below
+            else:
+                print("\n"+"There are no donors in list.")  # --> Let's the user know that a donor list does not exist.
+                bool_flag = False  # --> Is used as a flag on the while loop below.
+            break
+
+        # Gathers donation information from the user and prints a thank-you message for the donor.
+        while bool_flag:
+            try:
+                flo_donation_amount = float(input(str_name + ", please enter donation amount. "))
+
+                # Updates donor tuple
+                update_donor(donor_name, flo_donation_amount)
+
+                # String formatting for a brief thank-you to donor.
+                thankyou_message = "\n"+"{}, thank you for donating ${:,.2f}.".format(donor_name, flo_donation_amount)
+                print(thankyou_message)
+
+            except ValueError:  # --> Allows the program to continue if the user does not enter the right data type.
+                print(str_name + ", not a valid amount. \n")
+                continue
+
+            else:
+                break
+
+    @staticmethod
+    def create_report():
+        '''
+        This method takes the table tuple and creates a summary report for the user (name, number of gifts, total amount
+        in gifts, and the average amount given. The code sorts the list rows into the largest amount amount give to the
+        least amount given.
+        :return:
+        '''
+        total = int()
+        donation_count = int()
+        lst_table = []
+        # Creates a new list [name, number of gifts, average amount given, total amount in gifts] to be used to create
+        # the summary chart.
+        for row in tup_donor_names:
+            for value in range(2, len(row)):
+                total += row[value]
+                donation_count += 1
+            lst_table.append([row[0], row[1], total/donation_count, total])
+            total = 0
+            donation_count = 0
+        # Reverses the list
+        for row in lst_table:
+            row.reverse()
+        # Shorts the list form largest to smallest on the first element of each row.
+        lst_table.sort(reverse=True)
+
+        index = 0
+        # Formatting the money elements and changing all int() & float() data to str() data
+        # to use for creating the chart.
+        for row in range(len(lst_table)):
+            lst_table[row][index] = "${:,.2f}".format(lst_table[row][index])  # --> Total amount for gifts
+            lst_table[row][index + 1] = "${:,.2f}".format(lst_table[row][index + 1])  # --> Avg. gift amount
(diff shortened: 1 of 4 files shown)


students/rhamersky/Lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,335 @@
...
+#      Changed 'bool' variable to 'bool_flag'. Updated misleading function names.
+#      Reduced the amount of unnecessary print() statements.
+
+# -----Data Section-----
+import os
+str_name = None     # --> User's name
+int_option = int()  # --> User's option.
+tup_donor_names = (["William Gates, III",1,10000.0],["Mark Zuckerberg",1,20000.0],
+                   ["Jeff Bezos",1,1200.0],["Paul Allen",1,5000.0],
+                   ["Mike Scott",1,25.0])  # --> Tuple that stores all donor's information
```
natasha-aleksandrova: you should be using a dictionary for donors data structure

students/rhamersky/Lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,335 @@
...
+                else:
+                    print("Integer selected is not a valid option. "
+                          "Please, enter an integer and corresponds to the options given. \n")
+        return int_choice
+############## END OF MENU CLASS ######################
+
+# Methods in the sub_menu class will execute code based off of the user's sub-menu choice.
+
+
+class sub_menu(object):
```
natasha-aleksandrova: classes do *not* use snake case style but upper camel case -> `SubMenu`
and you don't need to extend from object in python3:
`class SubMenu:`

students/rhamersky/Lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,335 @@
...
+    @staticmethod
+    def write_email(donors):
+        '''
+        Writes emails to the donors and saves them to text files.
+        :param donors: Tuple table that stores all the donor's donation information.
+        :return:
+        '''
+        for donor_name in donors:
+            # Open text file and assign a mode to the file.
+            newFile = open(donor_name[0] + ".txt", mode="w")
```
natasha-aleksandrova: per pep8 snake_case for var names not camelCase

it is preferred to use `with` statement when working with files.

students/rhamersky/Lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,335 @@
...
+            break
+        else:
+            continue
+    return
+
+# -----Dispatch Dictionary Section-----
+menu_dispatch = {1 : choice1,
+                 2 : choice2,
+                 3 : choice3,
+                 4 : choice4
```
natasha-aleksandrova: it is better to name those functions with the actual action instead of numbered for readability. ex:
```
def send_thank_you: pass
def create_report: pass
def send_letters: pass
```

students/rhamersky/Lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,335 @@
...
+    while True:
+        # User menu to update the priority list.
+        print()  # --> Formatting
+        print("\t\tMenu of Options\n\n"
+              "\t\t1) Send a Thank You\n\n"
+              "\t\t2) Create a Report\n\n"
+              "\t\t3) Send letters to everyone\n\n"
+              "\t\t4) Quit\n")
+        # Error handles user's option choice.
+        global int_option
```
natasha-aleksandrova: this is confusing.. you don't need a `global` here and if you wanted to call a function then call it directly instead of using `thankyou_dispacth`

students/rhamersky/Lesson_5/mailroom_part3.py
```diff
@@ -0,0 +1,326 @@
...
+        :return: Returns int data (1-4) which is used to determine what menu option will be executed.
+        '''
+        while True:
+            try:
+                int_choice = int(input(str_name + ", please select an option above. "))
+            except ValueError:
+                print(str_name + ", not a valid option. \n")
+            # Verify that the user selected an option 1 through option 4.
+            else:
+                if 1 <= int_choice <= 4:
```
natasha-aleksandrova: you can check if `int_choice in menu_dispatch`

students/rhamersky/Lesson_5/mailroom_part3.py
```diff
@@ -0,0 +1,326 @@
...
+        # Creates a donor list.
+        while donor_name == "List":
+            # --> Verifies that there is at least a donor is stored as a list row in the table tuple.
+            if len(tup_donor_names) > 0:
+                donor_list_creation()  # --> Creates a list of donors
+                bool_flag = False  # --> Is used as a flag on the while loop below
+            else:
+                print("\n"+"There are no donors in list.")  # --> Let's the user know that a donor list does not exist.
+                bool_flag = False  # --> Is used as a flag on the while loop below.
+            break
```
natasha-aleksandrova: why is this a while loop if breaking out of it after single run?

students/rhamersky/Lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,335 @@
...
+
+        # Creates a donor list.
+        while donor_name == "List":
+            # --> Verifies that there is at least a donor is stored as a list row in the table tuple.
+            if len(tup_donor_names) > 0:
+                donor_list_creation()  # --> Creates a list of donors
+                bool_flag = False  # --> Is used as a flag on the while loop below
+            else:
+                print("\n"+"There are no donors in list.")  # --> Let's the user know that a donor list does not exist.
+                bool_flag = False  # --> Is used as a flag on the while loop below.
```
natasha-aleksandrova: when would it not have a list? you are always starting with a preset data

students/rhamersky/Lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,335 @@
...
+            break
+
+        # Gathers donation information from the user and prints a thank-you message for the donor.
+        while bool_flag:
```
natasha-aleksandrova: what is the purpose of the bool_flag?

students/rhamersky/Lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,335 @@
...
+        index = 0
+        # Formatting the money elements and changing all int() & float() data to str() data
+        # to use for creating the chart.
+        for row in range(len(lst_table)):
+            lst_table[row][index] = "${:,.2f}".format(lst_table[row][index])  # --> Total amount for gifts
+            lst_table[row][index + 1] = "${:,.2f}".format(lst_table[row][index + 1])  # --> Avg. gift amount
+            lst_table[row][index + 2] = str(lst_table[row][index + 2])  # --> Number of gifts
+        # Reverses list
+        for row in lst_table:
+            row.reverse()
```
natasha-aleksandrova: are above 4 loops necessary? can this be done in a single loop?

students/rhamersky/Lesson_4/mailroom_part2.py
rhamersky: Thanks for the input will be updated in lesson 6

students/rhamersky/Lesson_4/mailroom_part2.py
rhamersky: That makes sense, updated functions for readability.

students/rhamersky/Lesson_5/mailroom_part3.py
rhamersky: Thanks for the tip. This line I always had to remember to update if I changed the number of options in the menu. This will take care of it since I am using the dictionary dispatch to go into my functions.

students/rhamersky/Lesson_5/mailroom_part3.py
rhamersky: Oh it should be an if statement. When I originally was writing this code I wanted the program to give the user another shot at putting in a user prior to calling for a list that was empty. As the program was originally designed for an empty donors list and the user would put in the information. Through revisions I made the decision to kick the user back to the main menu if the donor list is empty and the user tries to call for a list.

students/rhamersky/Lesson_4/mailroom_part2.py
rhamersky: The would call the preset data 'test data'. My original code had an empty donor list and I wanted a way to let the user know that there were no donors in the list. The reason why this data is in my code because the lesson asked for 5 donors to be in the list. 

If I was doing to use this in a real life application I would have the 'test data' to prove to myself that it works. And I would also use that data in examples of how out puts are displayed for the client.

students/rhamersky/Lesson_4/mailroom_part2.py
rhamersky: The purpose of the bool_flag variable is that I wanted to enter the while loop only if the user did not want to display a list. I wanted the names of the donors to be displayed and then kick the user back to the main menu. 

If the user inputted a donor then the program would enter the while loop until the user entered data that can be cast as a float. 

That is why the bool_flag variable exist. It could be refactored, but I just wanted Boolean logic and this variable name made the most sense to me.

students/rhamersky/Lesson_4/mailroom_part2.py
rhamersky: I am not sure how to condense this section of code.

students/rhamersky/Lesson_4/mailroom_part2.py
rhamersky: Yes, that is confusing. I removed thankyou_dispatch in Lesson 6 and global int_option. Were thankyou_dipatch was called out I replaced it with the function call.

students/rhamersky/Lesson_4/mailroom_part2.py
natasha-aleksandrova: The way the mailroom lessons are set up is there is always pre-set data structure with donors data (this simulates a program reading from an existing Database and always having some data already there). 

This allows getting practice working with that specific data structure and keeping things fairly simple and clear to concentrate on language features added each mailroom version.

Lesson 3 expects list data structure for donors
Lessons 4, 5, 6 expects dictionary
Lessons 9, 10 expect a classes

which is why I comment every time about your data structure being a tuple 🙂

students/rhamersky/Lesson_4/mailroom_part2.py
natasha-aleksandrova: this should replace 99 through 144:
```
        summary = []
        # {"Jeff Bezos": [100, 200]}
        for donor, data in donors.items():
            # name, total, number_of_donations, avg
            summary.append(
                donor, sum(data), len(data), sum(data) / len(data)
            )
            # you can find your max value in this loop as well

        # do sorting
        sorted_summary = sorted_operation_here # this is just a single line using sorted()

        # print heading

        for row in sorted_summary:
            # format your row
            print("".format(*row))
```

you now only have 3 loops and a cleaner flow to follow

students/rhamersky/Lesson_4/mailroom_part2.py
natasha-aleksandrova: sounds like you wanted to exit the function, a simple return should do it in this case.

PR #585, opened 2018-05-18
Lesson 9

BrandonHenson:
Finished Lesson 9. Added a total to the report as well.


students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,138 @@
+# Brandon Henson
+# 5/17/18
+# Lesson 09 mailroom_5.py
+# Refactoring
+
+
+class Donor(object):
```
robalford: In Python 3, this can just be
```
class Donor:
```
(no need to explicitly inherit from `object` when defining a class)

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,138 @@
...
+
+    donor_history = {'Brandon Henson': [1005.49, 3116.72, 5200],
+                        'Alicia Henson': [21.47, 1500],
+                        'Michael Green': [2400.54],
+                        'Brandon Henson Jr': [355.42, 579.31],
+                        'Kaiya Henson': [636.9, 850.13, 125.23]}
```
robalford: In the assignment instructions, it is recommended to have a `Donor` class for logic and data related to a single donor, and a separate class for the collection of donors, this will give you a better separation of concerns and allow your code to be more encapsulated, flexible and extensible.

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,138 @@
...
+
+    name_list = [i for i in donor_history]
```
robalford: Calculated or dynamically generated data should go in a `@property` or method

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def __init__(self, name, *donation):
```
robalford: Don't think you need the `*` here, since this indicates a variable number of args

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,138 @@
...
+    def total(self):
+        return sum(self.donations)
+
+    @property
+    def average(self):
+        return self.total/self.num_of_donations
+
+    @property
+    def num_of_donations(self):
+        return len(self.donations)
```
robalford: These would be great use cases for properties on the donor collection class, but you don't appear to be using them (or the `donations` instance attribute) anywhere in your code as this is currently implemented

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class All(Donor):
+    def __init__(self):
+        pass
```
robalford: This is overriding the `__init__` method  of the `Donor` class to do nothing, which is a good indication that you don't need to be subclassing here

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,138 @@
...
+            for i in v:
+                count = count+1
+            total = sum(v)
+            list_of_sums.append(total)
+        gt = (sum(list_of_sums))
+        av = gt/count
+        gt = str(round(gt, 2))
+        av = str(round(av, 2))
+        print('The grand total of all donations is ${}.\
+ The average amount per donation is ${}.'.format(gt, av))
```
robalford: This method is static (doesn't reference instantiated `self` object) which is also a good indication that you don't need this class at all, since it's the only method on the class. I don't see any reason to subclass `Donor` here, this would be more appropriately defined as an additional method on the `Donor` class

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,138 @@
...
+def menu_1():
+    try:
+        fullname = input("Enter a full name or 'list' to view all\n")
+        if fullname.lower() == 'list':
+            print(Donor.name_list)
+
+        elif str(fullname) in Donor.donor_history:
+            amount = int(input("Donation amount? \n"))
+            newvalue = Donor.donor_history.get(fullname) + [amount]
+            donor_history[fullname] = newvalue
```
robalford: Unresolved reference here (should this be `Donor.donor_history`?)

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,138 @@
...
+        print("------------------------------------------------------------\n")
+        report = list()
+        for donor, total in Donor.donor_history.items():
+            report.append([donor, sum(total), len(total), sum(total)/len(total)])
+        sorted_report = sorted(report, key=lambda x: -x[1])
+
+        for i in sorted_report:
+            print("{:<20}   ${:>8} {:>8}        ${:<1}".format(
+                i[0], i[1], i[2], i[3]))
+        print(All.all_total(self=All))
```
robalford: The `self` parameter is a special parameter provided by Python to give access to the instance of a class in a method definition. It will always be the first positional argument of an instance method, and you never want/need to explicitly set it. However, in this case, you're not instantiating the object or using the `self` parameter in the method definition, which is why this would be more appropriately defined as a static method

BrandonHenson:
I reworked the assignment and Natasha closed the request. Do I get a zero
on the assignment?

natasha-aleksandrova:
@BrandonHenson no worries, you can submit a new PR and will grade it!

robalford:
@BrandonHenson closing the PRs when we request changes is the best way for us to keep your repo clean and keep things organized on our end and because of the limitations of the online course platform, we can only grade one submission per assignment at a time, that's why we have to temporarily grade the as incomplete but we will always regrade your assignment as 'Complete/Good' after you incorporate the requested revisions and resubmit.

natasha-aleksandrova:
your assignment was approved via `https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/607` 4 days ago

this is the old PR that was closed prior to PR #607

PR #588, opened 2018-05-18
Lesson 03 - list_lab.py

mcbraunie:
Final version for review.


students/MicahBraun/Lesson 03/list_lab.py
```diff
@@ -0,0 +1,138 @@
...
+print(sec1_list)                         # display updated list
+print()
+
+add_new_insert = "Strawberries"           # new value to be added to list
+sec1_list.insert(0,add_new_insert)        # add value to beginning of list using .insert()
+print(sec1_list)                          # display updated list
+print()
+
+for element in sec1_list:                 # print only items in list that begin with "P"
+    if element[:1] == "P":
```
robalford: I would just access directly without slicing: `element[0]`

PR #590, opened 2018-05-18
lesson 04 mailroom

rgpag:
lesson04 mailroom for review


students/rgpag/lesson04/mailroom.py
```diff
@@ -0,0 +1,111 @@
...
+    don_amt = int(input('What was the donation amount?'))
+    if don_amt == 0:
+        return ret_don, cnt_dict[ret_don], amt_dict[ret_don]
+    cnt_dict[ret_don] += 1
+    amt_dict[ret_don] += don_amt
+    avg_dict[ret_don] = amt_dict[ret_don]/cnt_dict[ret_don]
+    return ret_don, cnt_dict[ret_don], amt_dict[ret_don]
+
+
+def getKey(item):
```
robalford: Remember to use `snake_case` for functions per Pep8

students/rgpag/lesson04/mailroom.py
```diff
@@ -0,0 +1,111 @@
...
+    return item[1]
+
+
+def menu_sel_1():
+    name = input('Please provide full name ')
+    if name == 'quit':
+        quit = 1
```
robalford: This is overriding the Python built-in `quit` function (see how GitHub highlights in blue like other built-ins, `input`, `print` etc.)

students/rgpag/lesson04/mailroom.py
```diff
@@ -0,0 +1,111 @@
...
+        return quit
+    elif name == 'list':
+        for z in donors:
+            print(z)
+    elif name not in (donors):
```
robalford: Don't need the parens around `donors` here

students/rgpag/lesson04/trigrams.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python3
+import random
+
+# txt file to be used
+text_in = 'sherlock_full.txt'
+f = open(text_in)
```
robalford: It's recommended to use a `with` for opening files. This will automatically close the file when you exit the block.

students/rgpag/lesson04/trigrams.py
```diff
@@ -0,0 +1,68 @@
...
+msg = f.read()
+f.close
```
robalford: Missing parens on the call to `close()`, so this never closes the file

students/rgpag/lesson04/trigrams.py
```diff
@@ -0,0 +1,68 @@
...
+string = string.replace("\n", " ")
+string = string.replace("--", " ")
+string = string.replace("-", " ")
+spl_string = string.split(" ")
+
+# init vals
+str_dict = {}
+i = 0
+j = 0
+quit = 0
```
robalford: Overwriting built-in `quit` here, also it's good to define a flag variable like this closer to the point in code at which you are using it

students/rgpag/lesson04/trigrams.py
```diff
@@ -0,0 +1,68 @@
...
+
+# build trigram dictionary of keys
+for i in range(0, len(spl_string)-2):
+    key = spl_string[i] + " " + spl_string[i+1]
+    str_dict[key] = []
+
+# append values
+for i in range(0, len(spl_string)-2):
+    key = spl_string[i] + " " + spl_string[i+1]
+    str_dict[key].append(spl_string[i+2])
```
robalford: Check out the `dict.setdefault()` method for a way to accomplish the same thing without looping through twice

students/rgpag/lesson04/trigrams.py
```diff
@@ -0,0 +1,68 @@
...
+
+def trigrams(x):
+    start = rand_start()
+    print(get_next(x, start))
+
+
+if __name__ == "__main__":
+    print("you are in trigram main")
+    print("you are currently using text from {}".format(text_in))
+    while quit == 0:
```
robalford: I would use `while True` (sets up an infinite loop until some exit condition is met), or use a Boolean variable (set explicitly to `true` or `false`) for readability like `while_asking_for_size`

rgpag:
Hey Rob,

Thanks for reviewing and appreciate the feedback. I'll incorporate before
moving on.

-Robert

PR #592, opened 2018-05-19
Submitting Lesson05 Assignments

TinaBrownfield:
Finished except_exercise, and updated mailroom_part03.
Can you please look at my menu code on mailroom_part03 as I still can't get the script to exit out of the program, but all other parts of the menu (including quitting from sub to main menu work).


students/TinaB/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,216 @@
...
+# Menus
+
+
+def menu_selection(prompt, dispatch_dict):
+    """menu function"""
+    try:
+        while True:  # loop forever until they quit
+            response = input(prompt)
+            if dispatch_dict[response]() == "exit menu":
+                break
```
natasha-aleksandrova: you should put your KeyError around line 23 only

students/TinaB/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,216 @@
...
+               "Steven Hawking": [326892.24, 123, 123.33, 123, 123],
+               "Justin Timberlake": [999658.25, 1233, 123]}
+
+# Menus
+
+
+def menu_selection(prompt, dispatch_dict):
+    """menu function"""
+    try:
+        while True:  # loop forever until they quit
```
natasha-aleksandrova: The reason your program does not quit is because you have two while True loops, one above here and another on line 216.

students/TinaB/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,216 @@
...
+"""
+import datetime
+import os
+
+DONORS_DICT = {"William Gates": [326892.24, 122, 22, 12],
+               "Mark Zuckerberg": [30, 60, 65982.55],
+               "Jeff Bezos": [52636.27],
+               "Paul Allen": [877.33, 22],
+               "Steven Hawking": [326892.24, 123, 123.33, 123, 123],
+               "Justin Timberlake": [999658.25, 1233, 123]}
```
natasha-aleksandrova: all caps are usually used for constants, since we are modifying donors dictionary it should be all lower case variable.

students/TinaB/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+# functions for prompts ----
+
+def get_name_input():
+    ''' Function to select user input to return to print function'''
+    name_input = input(
+        "Please enter the name of the donor:  ")
+
+    for donor, donations in DONORS_DICT.items():
```
natasha-aleksandrova: you never use `donations` var, if you only need keys: `for donor in donors_dict:`

students/TinaB/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,216 @@
...
+
+        thank_you_letter = print_thank_you(donor_name, donor_amount)
+        print(thank_you_letter)
+
+
+def print_donors_names():
+    """ prints list of donors"""
+    print("\nDonors")
+    print("-"*20)
+    [print(d, "\n") for d in DONORS_DICT]
```
natasha-aleksandrova: comprehension is not a good pattern here since the output gets thrown out, how about this version:
```
print("\n".join(list(donors_dict)))
```

students/TinaB/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,216 @@
...
+def quit_menu():
+    '''Quit menu function and method'''
+    print("Quitting this menu now")
+    return "exit menu"
+
+
+# functions for prompts ----
+
+def get_name_input():
+    ''' Function to select user input to return to print function'''
```
natasha-aleksandrova: docstrings should be using triple double quotes

natasha-aleksandrova:
@TinaBrownfield I just saw your question on comprehensions on the submission form.

You are correct it is better to use regular for loop for something that does not need to consume a value  (with like print statements) since you are throwing away generated list.

students/TinaB/lesson05/mailroom_part3.py
TinaBrownfield: Thanks for finding that. I discovered it this weekend - and honestly could've kicked myself for the stupidity of that line! I was so mad at myself for one - writing it and two not catching it right away!

natasha-aleksandrova: it happens to all of us 🙂

PR #593, opened 2018-05-19
lesson 09 -- OO Mailroom

natrod22:
I have a try-except block, but it seems that the code at lines 89-92 never gets run, even when the donor name is found in the list. Not sure what I'm missing to get it to run this chunk.

### DIFF
--- a/students/NatalieRodriguez/Lesson09/OO_Mailroom.py
+++ b/students/NatalieRodriguez/Lesson09/OO_Mailroom.py
@@ -1,0 +1,172 @@
+#Natalie Rodriguez
+#Lesson 09: Object Oriented Mailroom
+# May 15, 2018
+
+#Hints
+'''
+You’ll want a Donor class – this will hold all the information about the donor, and have attributes, properties,
+and methods to provide access to the donor specific information that is needed.
+
+
+You’ll then want a class that handles the collection of donors. This will hold all the donor objects,
+but also methods to add a new donor, search for a given donor, etc. If you want a way to save and re-load your data,
+this class would have that too.
+
+Your class for the collection of donors will also hold the code that generates reports about multiple donors.
+
+Remember that the user interaction code (anything with an input function) should be outside of these
+“logic” or “model” classes.
+
+In general, you will have a method for each of the functions in your non-OO version.
+Which class they go it will depend on whether the method only needs the information from one donor,
+or from the whole collection.
+
+Rules of thumb for where to put methods:
+
+Hopefully, once you made your code testable, all the user-interaction code (with input()) is self contained
+in functions that don’t have any logic (data manipulation) in them. If not, then this is a good time to refactor.
+If a function does something with a single donor – it should be a method in the Donor class.
+If a function works with multiple donors – it should be in the class that handles a collection of donors.
+If a function contains a call to input() – it belongs outside of the logic classes – either stand alone in
+the module (like they are already) or perhaps all in a CLI class.'''
+
+#!/usr/bin/env python3
+import sys
+from collections import defaultdict
+
+class Donor:
+
+    def __init__(self, name, donations=None):
+
+        self.name = name
+        if donations is None:
+            self.donations = []
+        else:
+            self.donations = donations
+
+    def add_donation(self, amount):
+        try:
+            self.donations.append(int(amount))
+        except ValueError:
+            print("Enter a donation amount.")
+
+    def donation_amount(self):
+        return len(self.donations)
+
+    def total_donation(self):
+        try:
+            return sum(self.donations)
+        except TypeError:
+            return self.donations
+
+    def donation_avg(self):
+        try:
+            return self.total_donation()/ self.donation_amount()
+        except TypeError:
+            return self.donations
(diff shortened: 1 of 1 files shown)


students/NatalieRodriguez/Lesson09/OO_Mailroom.py
```diff
@@ -0,0 +1,172 @@
...
+            donor_names = []
+            for donor in self.donor_list:
+                donor_names.append(donor.name)
+            print(donor_names)
+
+        elif new_donor.lower() == 'dash':
+            return
+
+        else:
+            if new_donor in self.donor_list:
```
natasha-aleksandrova: the simplest way to debug is to add some `print` statements 🙂 

if you print `new_donor` and `self.donor_list` you will see that new_donor is a string (name) and `self.donor_list` is a list of Donor objects. So your if statement will always evaluate to False.

students/NatalieRodriguez/Lesson09/OO_Mailroom.py
```diff
@@ -0,0 +1,172 @@
...
+        try:
+            return sum(self.donations)
+        except TypeError:
+            return self.donations
+
+    def donation_avg(self):
+        try:
+            return self.total_donation()/ self.donation_amount()
+        except TypeError:
+            return self.donations
```
natasha-aleksandrova: you don't really want to sometimes return an integer and then other times return a list? especially this would essentially return wrong information.

students/NatalieRodriguez/Lesson09/OO_Mailroom.py
```diff
@@ -0,0 +1,172 @@
...
+            print("Enter a donation amount.")
+
+    def donation_amount(self):
+        return len(self.donations)
+
+    def total_donation(self):
+        try:
+            return sum(self.donations)
+        except TypeError:
+            return self.donations
```
natasha-aleksandrova: when would you get a type error?

students/NatalieRodriguez/Lesson09/OO_Mailroom.py
```diff
@@ -0,0 +1,172 @@
...
+class DonorHistory:
+
+    def __init__(self, donors=None):
+        self.donor_list = donors
+
+    def thank_you(self):
+        '''user inputs donor and amounts to dict.
+        '''
+        print('Type the full donor name (Enter "list" to show all donors or enter "dash" to return to the Donor Dashboard.')
+        new_donor = input(':')
```
natasha-aleksandrova: remember the separation of concern, your two main Data classes should have not `input` functions.

students/NatalieRodriguez/Lesson09/OO_Mailroom.py
```diff
@@ -0,0 +1,172 @@
...
+                print('''            Your contribution to the purchase of land for conservation
+            and habitat restoration will make a profound difference in allowing
+            the environment to recover from the detrimental impact of human
+            life. With noble individuals like yourself, we can make large strides
+            to allow the North American continent to exist without the stressors of 
+            agriculture and other human impacts. We appreciate and look forward to
+            your continued support.
+            
+            Best,
+            The Nature Conservancy''')
```
natasha-aleksandrova: this letter text would be better in Donor class since it only deals with single donor, ie:
```
new_donor_object.get_letter()
```

students/NatalieRodriguez/Lesson09/OO_Mailroom.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def write_file(self):
+        '''Creates a text file with a thank you message for each of the donors in the dictionary'''
+        for donor_data in self.donor_list:
+            filename = donor_data.name.replace(" ", "_") + ".txt"
+            total_donation = donor_data.total_donation()
+            letter = ('Thank you {} for you generous contributions in the amount of {:.2f}!'.format(donor_data.name, total_donation))
+            with open(filename, 'w') as file:
+                file.write(letter)
+            print(f"{donor_data.name}'s thank you has been saved at: " + filename)
```
natasha-aleksandrova: this can also live in Donor class; also, this function should be called `write_files` since you would be writing multiple files.

students/NatalieRodriguez/Lesson09/OO_Mailroom.py
```diff
@@ -0,0 +1,172 @@
...
+def quit_dashboard():
+    sys.exit("Exiting the Nature Conservancy Donor Dashboard. Goodbye!")
+
+if __name__ == '__main__':
+
+    Luke = Donor('Luke Rodriguez',[5512.75, 3250.50, 42.50])
+    River = Donor('River Tails',[63.00, 1200.00, 300.00, 450.00, 4000.00])
+    Virgil = Donor('Virgil Ferdinand',[350.00, 5000.00])
+    Jokib = Donor('Joseph Kibson',[3498.00, 5.50])
+    mailroom = DonorHistory([Luke, River, Virgil, Jokib])
```
natasha-aleksandrova: 👍

PR #596, opened 2018-05-19
Lesson09 & Lesson10

julienjass:


students/julienjass/Lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,234 @@
...
+        return "{}: {}".format(self._name, self._donations)  
+
+class DonorDB:
+    
+    def __init__(self):
+        self._donors = {}
+
+    def add_donor(self, donor):
+        if donor.name.lower() in self._donors:
+            self._donors[donor.name.lower()].add_donation(*(donor.donations))
```
robalford: Don't need the inner parens around `donor.donations`

students/julienjass/Lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,234 @@
...
+        return {k:v[1] for k, v in self._menu.items()}
+
+    
+def print_report():
+    print(donor_db.report())
+
+
+def return_to_main():
+    return
+
```
robalford: I'd look for another way to represent this instead of creating a function with nothing but a bare `return`

students/julienjass/Lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,234 @@
...
+
+def write_letters_on_disk():
+    donor_db.letters_on_disk()
+    return
+
```
robalford: Why not call the method directly instead of wrapping in another function?

students/julienjass/Lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,234 @@
...
+    check = 0
+    while check == 0:
+        try:
+            amount = float(amount)
+        except ValueError:
+            print('Please enter an amount which is valid')
+            amount = input('Amount: ')
+            continue
+        else:
+            ex = 1
```
robalford: You don't appear to be doing anything with`ex`

students/julienjass/Lesson10/mailroom_fp.py
```diff
@@ -0,0 +1,297 @@
...
+        return str([d for d in self._donors.values()])
+
+    @property
+    def projection(self):
+        return {d[0]:d[1] for d in self._projection}
+
+    def challenge(self, x=1, mn=0, mx=float("inf")):
+        self._projection = []
+        for k, v in self.export.items():
+            self._projection.append((k, list(map(lambda y: y * x, list(filter(lambda z: z >= mn and z <= mx, v))))))
```
robalford: Although it's common to use 'throw away' variable names for things like lambdas, this code is pretty obscure. I would suggest using meaningful variable names to make it more readable

julienjass:
@robalford Thanks for your feedback during the course. It was very helpful.

PR #600, opened 2018-05-19
Lesson 7: HTML Render exercise

itsabhister:
This was probably my favorite exercise..I found myself struggling with the whole Duck typing concept (Step 2) particularly, but hopefully I think I'm getting a hang of it.

### DIFF
--- a/students/itsabhi/Lesson7/html_render.py
+++ b/students/itsabhi/Lesson7/html_render.py
@@ -1,0 +1,164 @@
+#!/usr/bin/env python3
+
+
+class Element:
+
+    tag = ""
+    indent = "   "
+
+    def __init__(self, content=None, **kwargs):
+        # Takes optional keyword arguments that can be used to initialize additional element parameters like font etc.
+        if isinstance(content, str):
+            # Initializes a list with the content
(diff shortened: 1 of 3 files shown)


students/itsabhi/Lesson7/html_render.py
```diff
@@ -0,0 +1,164 @@
...
+        return tag_str
+
+
+class Html(Element):
+    tag = "html"
+
+    def render(self, file_out, cur_ind=""):
+        f = file_out
+        f.write('<!DOCTYPE html>'+'\n')
+        Element.render(self, f)
```
natasha-aleksandrova: note that it is better to use `super().render(f)`

you can read more about this here: https://stackoverflow.com/questions/21639788/difference-between-super-and-calling-superclass-directly

students/itsabhi/Lesson7/html_render.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class Li(Element):
+
+    tag = "li"
+
+
+class H(OneLineTag):
+
+    tag = "h"
```
natasha-aleksandrova: only need one line space

PR #601, opened 2018-05-20
Lesson07 and Lesson08

DeFeNdog:


students/carlos_novoa/lesson07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+                    file_out.write('{}{}\n'.format(cur_ind, node))
+                else:
+                    file_out.write(self.indent)
+                    file_out.write('{}{}\n'.format(cur_ind, node))
+        file_out.write("{}</{}>\n".format(cur_ind, self.tag))
+
+
+class Html(Element):
+    def __init__(self, content=None, **kwargs):
+        self.tag = 'html'
```
robalford: `tag` should be a class attribute (keep it how it was initially defined at class level on `Element`, overriding the value for each subclass) , since it will be shared by all instances of the `Html` class

students/carlos_novoa/lesson07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def render(self, file_out, cur_ind=""):
+        file_out.write("<!DOCTYPE html>\n")
+        super().render(file_out, cur_ind)
+
+
+class Body(Element):
+    def __init__(self, content=None, **kwargs):
+        self.tag = 'body'
+        super().__init__(**kwargs)
```
robalford: See my note above on making `tag` a class attribute and then you won't need to override `__init__` for many of the `Element` subclasses (unless you need to add some custom logic or pass in some extra args to object instances)

students/carlos_novoa/lesson07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+    def __init__(self, content=None, **kwargs):
+        self.tag = 'head'
+        super().__init__(**kwargs)
+
+
+class OneLineTag(Element):
+    tag = ''
+
+    def __init__(self, content=None, **kwargs):
+        super().__init__(content)
```
robalford: Don't think you need to override `__init__ here

students/carlos_novoa/lesson07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+
+class SelfClosingTag(Element):
+    def __init__(self, content=None, **kwargs):
+        self.content = content
+        self.atts = kwargs if kwargs else ''
+        try:
+            if self.content:
+                raise TypeError
+        except TypeError as e:
+            raise
```
robalford: You don't need the try/except block here, you can just check for content and raise the error. Note that this could alternatively be handled a bit more concisely in the `append` method where `content` is your only method parameter

students/carlos_novoa/lesson08/circle_class.py
```diff
@@ -0,0 +1,95 @@
...
+
+"""
+Lesson8 - Circle Class
+"""
+
+
+class Circle(object):
+    version = '0.1'
+
+    def __init__(self, radius=None, diameter=None):
```
robalford: Note that the assignment only asks for a single required radius parameter for the constructor method. This should create a `radius` instance attribute (no getter and setter required)

students/carlos_novoa/lesson08/circle_class.py
```diff
@@ -0,0 +1,95 @@
...
+        if radius is not None:
+            try:
+                if radius < 0:
+                    raise ValueError()
+            except ValueError:
+                print('Radius must be positive.')
+                return
+            else:
+                self._radius = radius
+                self._diameter = radius * 2
```
robalford: You don't need to calculate this here, since you are doing it below in the diameter `@property`. The way properties work, is that they run each time the attribute that they define is accessed on the object instance. So every time that your program accesses `circle.diameter`, the diameter will be calculated in your `@property` based on the current radius value. This is why the `@property` decorator is often used for values that are calculated from the values of other attributes.

students/carlos_novoa/lesson08/circle_class.py
```diff
@@ -0,0 +1,95 @@
...
+                self._area = round(math.pi * radius ** 2.0, 6)
```
robalford: This calculation should also be moved to the `area` property. That way the area can be recalculated based on the current radius value every time the area attribute is accessed. As this is written, your area will only be defined based on the initial radius value, and will not stay in sync if the radius value is changed

students/carlos_novoa/lesson08/circle_class.py
```diff
@@ -0,0 +1,95 @@
...
+                self._area = round(math.pi * self._radius ** 2.0, 6)
+
+    @property
+    def diameter(self):
+        return self.radius * 2
+
+    @diameter.setter
+    def diameter(self, diameter=None):
+        self._diameter = diameter
+        self._radius = diameter / 2
```
robalford: This is the correct approach to using getters and setters here for diameter

students/carlos_novoa/lesson08/circle_class.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @property
+    def radius(self):
+        return self._radius
+
+    @radius.setter
+    def radius(self, radius=None):
+        self._radius = radius
+        self._diameter = radius * 2
```
robalford: Don't need to repeat the diameter calculation here, since it will run in the diameter property every time the diameter attribute is accessed. But you also don't need the getter and setter for radius at all per my earlier comment, radius should be defined and set as instance attribute with no extra logic needed (the extra logic lives in the diameter and area getters/setters)

robalford:
@DeFeNdog I don't see html renderer assignment submitted on the course website. Remember to submit it there as well so we can grade. Thanks!

PR #603, opened 2018-05-20
Lesson 6 Complete

rhamersky:
Lesson 6 is ready for review

### DIFF
--- a/students/rhamersky/Lesson_6/mailroom_part4.py
+++ b/students/rhamersky/Lesson_6/mailroom_part4.py
@@ -9,44 +9,40 @@
 #      Reduced the amount of unnecessary print() statements.
 #
 # Rev: B - 05/19/2018 updated code to reflect comments made by the instructor in mailroom_part3.py.
+# Rev: C - 05/21/2018 updated code to reflect comments made on 05/20/2018.
+#      Changed tuple data set to dictionary data set. Refactored program.
 
 # -----Data Section-----
 import os
 str_name = str()    # --> User's name
-int_option = int()  # --> User's option.
-tup_donor_names = (["William Gates, III",1,10000.0],["Mark Zuckerberg",1,20000.0],
-                   ["Jeff Bezos",1,1200.0],["Paul Allen",1,5000.0],
-                   ["Mike Scott",1,25.0])  # --> Tuple that stores all donor's information
+int_user_input = int()  # --> User's option.
+dic_donors = {"William Gates, III":[10000.0, 1], "Mark Zuckerberg":[20000.0, 1],
+                   "Jeff Bezos":[1200.0,1],"Paul Allen":[5000.0,1],
+                   "Mike Scott":[25.0,1]} # --> Dic that stores all donor's information
 str_salutation = ""   # --> Email salutation
 str_body = ""         # --> Email body
 str_valediction = ""  # --> Email valediction
 
 
 # -----Process Section-----
-#  Methods defined in the menu class are related to general menu checks or options based on user inputs.
-
-class Menu():
-    @staticmethod
-    def menu_options():
-        '''
-        Try and except block used on User's option choice to check for integers.
-        :return: Returns int data (1-4) which is used to determine what menu option will be executed.
-        '''
-        while True:
-            try:
-                int_choice = int(input(str_name + ", please select an option above. "))
-            except ValueError:
-                print(str_name + ", not a valid option. \n")
-            # Verify that the user selected an option 1 through option 4.
+def menu_options():
+    '''
+    Try and except block used on User's option choice to check for integers.
+    :return: Returns int data (1-4) which is used to determine what menu option will be executed.
+    '''
+    while True:
+        try:
+            int_choice = int(input(str_name + ", please select an option above. "))
+        except ValueError:
+            print(str_name + ", not a valid option. \n")
+        # Verify that the user selected an option 1 through option 4.
+        else:
+            if 1 <= int_choice <= 4:
+                break
             else:
-                if int_choice in menu_dispatch:
-                    break
-                else:
-                    print("Integer selected is not a valid option. "
-                          "Please, enter an integer and corresponds to the options given. \n")
-        return int_choice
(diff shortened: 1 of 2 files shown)


students/rhamersky/Lesson_6/mailroom_part4.py
```diff
@@ -0,0 +1,325 @@
...
+#
+# Rev: B - 05/19/2018 updated code to reflect comments made by the instructor in mailroom_part3.py.
+
+# -----Data Section-----
+import os
+str_name = str()    # --> User's name
+int_option = int()  # --> User's option.
+tup_donor_names = (["William Gates, III",1,10000.0],["Mark Zuckerberg",1,20000.0],
+                   ["Jeff Bezos",1,1200.0],["Paul Allen",1,5000.0],
+                   ["Mike Scott",1,25.0])  # --> Tuple that stores all donor's information
```
robalford: You should be using a `dict` for your main data structure at this point in the course and that will simplify a lot of your other code that deals with looking up and adding donors and donations

students/rhamersky/Lesson_6/mailroom_part4.py
```diff
@@ -0,0 +1,325 @@
...
+str_salutation = ""   # --> Email salutation
+str_body = ""         # --> Email body
+str_valediction = ""  # --> Email valediction
+
+
+# -----Process Section-----
+#  Methods defined in the menu class are related to general menu checks or options based on user inputs.
+
+class Menu():
```
robalford: No need for parens in class definition

students/rhamersky/Lesson_6/mailroom_part4.py
```diff
@@ -0,0 +1,325 @@
...
+    @staticmethod
+    def menu_options():
```
robalford: If your class only includes a single static method, that's a good indication that you don't need to be using a class at all. We'll learn a lot more about classes and object-oriented Python later in the course, but generally you only need to use a class in Python if you have a collection of related data (attributes) and logic (methods) that act upon that data. 

In this case, the only reason I can see for using a class is to provide some organizational separation/hierarchy for your program, but there are more appropriate options like module level functions and modules themselves for accomplishing this in Python.

students/rhamersky/Lesson_6/test_mailroom_part4.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+import unittest
+from mailroom_part4 import tup_donor_names, Menu, SubMenu, Files, update_donor, donor_list_creation,send_thankyou_option, \
+    create_report_option, exit_program_option,lifetime_donations,menu_dispatch
+
+class MyTestCase(unittest.TestCase):
+    def test_creatReport(self):
```
robalford: Method names should be written in `snake_case` (I know the unittest library itself breaks this rule for its assert methods, probably because it was originally ported from Java, but the current Python/Pep8 guideline is `snake_case` for variables, functions and methods and `CamelCase` for class names and this is the common convention among Python devs).

students/rhamersky/Lesson_6/test_mailroom_part4.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def test_exitProgram(self):
+        self.assertEqual(exit_program_option(),"exit")
+
+    def test_lifetimeDonations(self):
+        self.assertDictEqual(lifetime_donations(tup_donor_names), lifetime_donations(tup_donor_names))
+
+    def test_menu_dispatch(self):
+        self.assertDictEqual(menu_dispatch,menu_dispatch)
+
```
robalford: Most of these tests are not actually testing anything other than the unittest assert statements and Python itself are working as expected. Since frameworks and languages are already well tested, you never need to write these kinds of tests yourself. 

The only actual test that I can see here is `test_donorEmail` where you are testing your own program's dynamic logic against an expected result/value. 

It seems as though you may need to refactor your mailroom code so that more of your functions return testable values. This will also have the positive effect of making your code more tightly focused, flexible and extensible, which will be important as we move into more advanced lessons and mailroom assignments.

Some ways to do this would be to have your functions return strings to be printed elsewhere, rather than printing as a side effect of the function. You could also use a test donor data structure for your unit tests, and write tests for your functions that modify that data structure, with assert statements that test against expected results/values.

rhamersky: I will work on this and send it back in as soon as possible.

rhamersky:
I will work on making Lesson 6 more testable and resubmit it.

I also wanted to let you know that I was already working on Lesson 8 and I pushed my completed Lesson 8.

robalford:
@rhamersky sounds good, I'll review your lesson 8 assignment here, and we'll look for your updates to lesson 6.

natasha-aleksandrova:
@rhamersky can you please let us know when you add updates for Lesson 6 and if possible hold off on committing next assignment until we merge this?

rhamersky:
Lesson 6 has been updated.

Yes, I will hold off until merge. Sorry about that.

natasha-aleksandrova:
@rhamersky thanks! we should be able to merge this today (@robalford )

students/rhamersky/Lesson_6/mailroom_part4.py
```diff
@@ -0,0 +1,263 @@
...
+
+        # Creates the donor rows for the chart.
+        sorted_dic = dic_sort()
+        for name in sorted_dic:
+            donation_avg = dic_donors[name][0] / dic_donors[name][1]
+            print(("{0:{padding}}".format(name, padding=longest_entry)), end=" ")
+            print(("{0:{padding}}".format("{}".format(dic_donors[name][1]), padding=longest_entry)), end=" ")
+            print(("{0:{padding}}".format("${:,.2f}".format(donation_avg), padding=longest_entry)), end=" ")
+            print(("{0:{padding}}".format("${:,.2f}".format(dic_donors[name][0]), padding=longest_entry)), end=" ")
+            print()
```
robalford: You could build up a string to return, rather than printing line by line to test the content of the report

PR #607, opened 2018-05-20
Lesson 9 addressed

BrandonHenson:
I've really been struggling with this one. I hope I understand the encapsulation a little better now.


students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,151 @@
...
+        self._name = name
+        self._donations = []
+
+    @property
+    def name(self):
+        return self._name
+
+    @name.setter
+    def name(self, val):
+        self._name = val
```
natasha-aleksandrova: you don't need to use properties if you aren't doing anything special to the attributes (like we did for `area` in Circle assignment) or you have to protect access to the attribute (here single underscore will not protect the attribute); as it is here it provides no value.

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,151 @@
...
+    @property
+    def donors(self):
+        return self._donors
+
+    @donors.setter
+    def donors(self, val):
+        self._donors = val
+
+    def add_donor(self, donor):
+        self.donors[donor.name] = donor.donations
```
natasha-aleksandrova: Great choice to use a dictionary for easy search functionality to find your donor, however I would recommend value to be a Donor object so you can access anything you want about donor.

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,151 @@
...
+
+if __name__ == '__main__':
+    donor_history = {}
+    donor_history['Brandon Henson'] = [1005.49, 3116.72, 5200]
+    donor_history['Alicia Henson'] = [21.47, 1500]
+    donor_history['Michael Green'] = [2400.54]
+    donor_history['Brandon Henson Jr'] = [355.42, 579.31]
+    donor_history['Kaiya Henson'] = [636.9, 850.13, 125.23]
+
+    DonorDict = DonorAll(donor_history)
```
natasha-aleksandrova: I would recommend passing in a list of Donor objects:

```
donors = DonorAll(Donor("alex", [100]), Donor("chris", [20, 100]))
```

`DonorAll` instance variable should follow snake_case convention -> `donor_dict = DonorAll(donor_history)`

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,151 @@
...
+             '[1] Send a Thank You\n'
+             '[2] Create a Report\n'
+             '[3] Send letters to everyone\n'
+             '[4] Exit\n')
+        else:
+            response = input('\nSelect an option:\n'
+             '[1] Send a Thank You\n'
+             '[2] Create a Report\n'
+             '[3] Send letters to everyone\n'
+             '[4] Exit\n')
```
natasha-aleksandrova: you can combine lines 118 to 151 to simplify and reduce duplicated code:
```
while True:
    response = input('>...')
    try:
        args_dict[int(response)](donor_dict)
    except:
        print("some error message")
```

I think this will satisfy all of your needs in those lines.

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,151 @@
...
+    def createfile(self, name):
+        for key, values in donor_history.items():
+            filename = str(key)+'.txt'
+            with open(filename, 'w') as fileobj:
+                total = sum(values)
+                numgifts = len(values)
+                fileobj.write("Dear {}\nThank you for your {} generous donations \
+totaling ${}\nThe money will be put to good use.\n\n\
+            Sincerely, \n                -The Team".format(key, numgifts, total))
+            fileobj.close()
```
natasha-aleksandrova: you don't need to worry about closing the file when you're using `with` statement.

students/BrandonHenson/Lesson9/Mailroom_5.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def menu_2(self):
+        try:
+            print(f'{"Donor Name":18s} {"|  Total Given":12s} \
+    {"|  Num Gifts  |":14s} {"Average Gift":18s}')
+            print(f'{"-"*62}')
+            [print(f'{name:18s} ${sum(self.donors[name]):10.2f} {len(self.donors[name]):8d}        \
```
natasha-aleksandrova: it is preferred to use use regular for loop when you are discarding the return value.

PR #608, opened 2018-05-20
lesson09 mailroom with classes

kmsnyde:
Had a difficult time with mostly with setting it up and finally used a dictionary after much research but I think I took the long way around some areas.  Slim on being able to use TDD because of my troubles I sort of troubleshot along the way hitting F5 a thousand time and making updates.

### DIFF
--- a/students/kmsnyde/lesson09/mailroom5_classes.py
+++ b/students/kmsnyde/lesson09/mailroom5_classes.py
@@ -1,0 +1,135 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Sun May 20 07:03:34 2018
+
+@author: Karl M. Snyder
+"""
+
+class Main:
+    def __init__(self):
+        self.menu = {1: 'Send a Thank You',
+                     2: 'Create a Report',
+                     3: 'Send letters to everyone',
+                     4: 'Quit'}
+    
+    def main_menu(self):
+        print('\n', 'Please select a number from the following choices:\n')
+        return {(print(str(k) + ':', v)) for k, v in self.menu.items()}
+    
+    def selection(self):
+        while True:
+            input1 = input("Selection: ")
+            #print(type(input1), input1)
+            try:
+                if int(input1) in range(1, 4):
+                    if int(input1) == 1:
+                        print('\nType a user\'s name or "list" to show names.')
+                        input2 = input('-> ')
+                        if input2 == 'list':
+                            our_donors.donors()
+                            continue
+                        else:
+                            input3 = input('Donation amount: ')
+                            user = Donor(input2, input3)
+                            user.send_thanks()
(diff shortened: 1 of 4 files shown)


students/kmsnyde/lesson09/mailroom5_classes.py
```diff
@@ -0,0 +1,135 @@
...
+                        our_donors.create_report()
+                    if int(input1) == 3:
+                        our_donors.send_letters_all()
+                elif int(input1) == 4:
+                    print("Exiting program...")
+                    break
+            except ValueError:
+                print("You must use a menu number between 1-4; try again!")
+
+class Donor:
```
natasha-aleksandrova: your `Donor` class should represent a single `Donor` with properties like `name` (string) and `donations` (list) at the least.

your `Donors` should handle a collection of `Donor` objects, internally it may create a dictionary to make it easier.

here is an example of interaction:
```
donor1 = Donor("Cat", [100])
donor2 = Donor("Bob", [200])
donors = Donors(donor1, donor2)
donors.create_report()
```
----
your current `Donor` has knowledge of many donors which shouldn't be the case.

students/kmsnyde/lesson09/mailroom5_classes.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __repr__(self):
+        return "Name: {}, Sum Donated: ${:.2f}".format(
+                self.name, Donor.donor_dict[self.name]['total'])
+        
+    def send_thanks(self):
+        letter = 'Thank you {} for your donation in the amount of ${}; it is very generous.'.format(self.name, self.donations)
+        with open('Thank_You - {}.txt'.format(self.name.lower().
+                     replace(' ', '_')), 'w') as f:
+            f.write(letter)
```
natasha-aleksandrova: 👍

students/kmsnyde/lesson09/mailroom5_classes.py
```diff
@@ -0,0 +1,135 @@
...
+    
+    donors_list = [(name, Donor.donor_dict[name]['total'], Donor.donor_dict[name]['num_donations'], Donor.donor_dict[name]['avg']) for name in Donor.donor_dict]
+    
+    donors_list_sort= sorted(donors_list,key=lambda i: i[1], 
+                                reverse=True)
+    
+    letters = 'Dear {},\n\n\tThank you for your total contributions in the amount of ${}.\n\n\tYou are making a difference in the lives of others.\n\n\t\tSincerely,\n\t\t"Working for America"'
+    
+    def donors(self):
+        {print(k) for (k, v) in Donor.donor_dict.items()}
```
natasha-aleksandrova: it is better to use regular for loop when you are discarding the output

PR #613, opened 2018-05-20
Refactored with classes

mmsunday1:


students/M_Sunday/lesson09/Mailroom_Part_5.py
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/python
+from operator import itemgetter
+
+# Lesson 09, Mailroom Part 5 Assignment
+
+
+class UserInputs(object):
```
natasha-aleksandrova: in python3 no need to extend from `object` -> `class UserInputs:`

students/M_Sunday/lesson09/Mailroom_Part_5.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @staticmethod
+    def get_menu_option():
+        sel_ = input("------> "
+                     "Select an option from the menu: ")
+        return sel_
```
natasha-aleksandrova: using underscores at the end of variable names is _not_ a common patter in python

students/M_Sunday/lesson09/Mailroom_Part_5.py
```diff
@@ -0,0 +1,173 @@
...
+    def add_donation(self, name, donation):
+        if not(name in self.donor_data_):
+            self.donor_data_[name] = donation
+        else:
+            self.donor_data_[name].append(donation)
+            print("adding printer")
+
+    def list_display(self):
+        print("\n---- List of all Donors ----")
+        [print("- {}".format(item)) for item in self.donor_data_]
```
natasha-aleksandrova: it is preferred to use regular for loop if you are not consuming the return value

PR #614, opened 2018-05-20
modified mailroom_part3.py Need help with a few problems

seaneightysix:
Hello,

I am having trouble with my mailroomPart3.py script. I have two problems that I cannot seem to figure out. Any assistance would be appreciated.

The first problem occurs somewhere between line 46-47 in the thank_you() function. When called, the function runs fine all the way until line 46. It then outputs to the terminal “Enter a number between 1-4.” (Which is the string for the except TypeError I have on line 25) and prompts the user with the main_prompt which is called in the main_menu_response() function on line 15. For some reason the script does not finish the else block from lines 43-48. It cuts off before line 47.  Any ideas why this is happening?

My second problem occurs at line 72 which is suppose to sort the list I have by the total given amount. Is there something wrong with my code on line 72? I copied and pasted this code directly from my mailroom_part2.py script where it ran fine there. The only difference is that the list is different, but otherwise it should run fine. Please help.

Thank you,
Sean


natasha-aleksandrova:
@seaneightysix your line 72 question:
you sorting `lis` list on line 72 but when you are generating the report on lines 76,77 you are not using that list, you are using a dictionary which isn't sorted.

natasha-aleksandrova:
I'm having a hard time following and understanding your first problem, you say that "It cuts off before line 47"..but there is no code left to run besides a call to `thank_you()` function.

is there a visual bug like when you run a program a bug happens? if so, can you provide me the repro steps so I can run it myself.

Otherwise, could you try clarifying your question?

natasha-aleksandrova:
@seaneightysix I'm not sure what the intent of this PR is? It looks like to be updates to existing mailroom that's been merged?

please comment and let us know and we can reopen.

seaneightysix:
Hi Natasha,

Sorry for the confusion. Actually, I'm still a little bit unclear as to the
intent of a PR. My understanding is that anytime I would like an instructor
to review my code, I have to submit a PR of that code. Is that correct? I
submitted PR #614 because I had questions about that code and wanted help
from an instructor. Or do I submit a PR only for my final versions of
assignments that are ready to be graded?

Anyway, I've figured out what was wrong with my mailroom_part3.py script.
Thanks for your help.

 I will submit a PR later this week with my final versions of Lesson05
assignments. Thank you.

PR #615, opened 2018-05-21
Exceptions Assignment

UWTestEngr:
Added exceptions to mailroom assignment

### DIFF
--- a/students/Matt_Hudgins/Lesson05/mailroom_3.py
+++ b/students/Matt_Hudgins/Lesson05/mailroom_3.py
@@ -1,0 +1,131 @@
+#!/usr/bin/env python3
+
+# chmod +x mailroom.py needs to be performed before executable
+
+'''
+    File Name: mailroom_3.py
+    Author: Matt Hudgins
+    Date created: 5/20/18
+    Date last modified: 5/20/18
+    Python Version 3.6.4
+'''
+import sys
+
+# This is the donor list
+donors = {}
+donors['Rick Grimes'] = [5.00, 10.00, 2.00]
+donors['Shane Walsh'] = [4.00, 10.00, 9.00]
+donors['Carl Grimes'] = [72.00, 10.00, 88.00]
+donors['Morgan Jones'] = [68.00, 10.00, 98.00]
+
+
+def donor_list():
+    '''This function prints a list of all current donors'''
+    for i in donors:
+        print(i)
+    return
+
+
+def adding_a_donor(newname):
+    '''This function adds a name to the donor list'''
+    donors[newname] = []
+    adding_a_donation(newname)
+    return
+
+
+def adding_a_donation(newname):
+    '''This function adds a donation to the the list'''
+    donation_amount = round(float(input(f'Enter a Donation amount for\n'
+    '{newname}\n')), 2)
+    if newname in donors.keys():
+        donors[newname].append(donation_amount)
+    elif newname not in donors:
+        donors[newname] = [donation_amount]
+    print(donors)
+    return
+
+
+def send_letter(newname, amount):
+        letter = 'Dear {},\n Thank you for your donation of ${}'
+        print(letter.format(newname, amount))
+        print('\n')
+        return
+
+
+def create_report():
+    '''This function creates a report'''
+    headers = ['Donor Name', 'Total Given', 'Num Gifts', 'Average Gift']
+    print('{:20} | {:20} | {:20} | {:20}'.format(*headers))
+    for name, donations in donors.items():
+        total = sum(donations)
+        num = len(donations)
+        avg = total / num
+        spacing = '{:20} | {:20} | {:20} | {:20}'
+        print(spacing.format(name, total, num, avg))
+    print('\n')
+    return
+
+def letter_to_all():
+        big_letter = ('Dear {},\n\n Thank you for you donation of {}\n'
+        'your money will go a long way\n')
+        for donor, donations in donors.items():
+            print(big_letter.format(donor, sum(donations)))
+            filename = donor.replace(' ', ' ') + '.txt'
+            save = open(filename, 'w')
+            save.write(big_letter.format(donor, sum(donations)))
+            save.close()
+        return
+
+
+def main_menu():
+    """main block"""
+    while True:
+        prompt = 'Please enter a donor\'s name or type list:\n'
+        entry = input(prompt)
+        if entry.lower() == "list":
+            donor_list()
+        elif entry.lower() == "quit":
+            break
+        else:
+            try:
+                assert entry in donors
+                adding_a_donation(entry)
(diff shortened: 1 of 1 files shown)


students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+
+'''
+    File Name: mailroom_3.py
+    Author: Matt Hudgins
+    Date created: 5/20/18
+    Date last modified: 5/20/18
+    Python Version 3.6.4
+'''
+import sys
+from collections import deque
```
robalford: Doesn't look like you're using this import

students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+
+# This is the donor list
+donors = {}
+donors['Rick Grimes'] = [5.00, 10.00, 2.00]
+donors['Shane Walsh'] = [4.00, 10.00, 9.00]
+donors['Carl Grimes'] = [72.00, 10.00, 88.00]
+donors['Morgan Jones'] = [68.00, 10.00, 98.00]
+
```
robalford: This would generally be defined as a dict literal using this syntax:
```
donors = {
    'Rick Grimes': [5.00, 10.00, 2.00],
    'Shane Walsh': [4.00, 10.00, 9.00],
}
```

students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+
+def donor_list():
+    '''This function prints a list of all current donors'''
```
robalford: Use triple double quotes for docstrings for Pep8 (`"""`)

students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+def adding_a_donor(newname):
+    '''This function adds a name to the donor list'''
+    donors[newname] = []
+    adding_a_donation(newname)
+    return
+
+
+def adding_a_donation(newname):
+    '''This function adds a donation to the the list'''
+    donation_amount = round(float(input(f'Enter a Donation amount for\n'
```
robalford: This would be a good place for a try/except block where you try the float conversion and catch a `ValueError` for bad input

students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+    '{newname}\n')), 2)
+    if newname in donors.keys():
```
robalford: `in` checks for key membership in dict so no need to call `keys()`

students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+        donors[newname].append(donation_amount)
+    elif newname not in donors:
+        donors[newname] = [donation_amount]
```
robalford: Check out the `dict.setdefault()` function for a more concise way to do this

students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+    print(donors)
+    return
+
+
+def send_letter(newname, amount):
+        letter = 'Dear {},\n Thank you for your donation of ${}'
+        print(letter.format(newname, amount))
+        print('\n')
+        return
```
robalford: These bare `return` statements are not necessary. Python functions will return `None` by default if no return value is specified and the convention is to leave them off the function if you don't need them for anything.

students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+    print('\n')
+    return
+
+def letter_to_all():
+        big_letter = ('Dear {},\n\n Thank you for you donation of {}\n'
+        'your money will go a long way\n')
+        for donor, donations in donors.items():
+            print(big_letter.format(donor, sum(donations)))
+            filename = donor.replace(' ', ' ') + '.txt'
+            save = open(filename, 'w')
```
robalford: It's a best practice to use a `with` block for working with files, which will always automatically close the file for you when your program exits the block

students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+    while True:
+        prompt = 'Please enter a donor\'s name or type list:\n'
+        entry = input(prompt)
+        if entry.lower() == "list":
+            donor_list()
+        elif entry.lower() == "quit":
+            break
+        else:
+            try:
+                assert entry in donors
```
robalford: You dont generally see assert statement used in live/production code outside of tests. Here's a good [article](https://dbader.org/blog/python-assert-tutorial) on why.

Since except blocks are for catching and handling anticipated errors in your live code, it's better to use an existing exception such as `KeyError` or `Value Error` based on the expected error if you can. But in this case, I'm not sure if you should be raising or handling an exception at all since the intent is for new donors to be added to the donor dict.

students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+                 '3': letter_to_all,
+                 '0': quit
+                 }
+
+
+def menu(prompt, dispatch):
+    while True:
+        entry = input(prompt).lower()
+        try:
+            if dispatch.get(entry)() == "exit menu":
```
robalford: This `if` condition is never actually checking anything. Your `main_dispatch` functions are called in the `if` statement before the `==` operator, so it's never checking the return value against 'exit menu'. And when a user enters '0', it's calling Python built-in `quit` function, rather than your `quit` function defined below.

I would rename your `quit` function to something that doesn't clash with built in, and move it above the `main_dispatch` dict. Then you can use dict key bracket access in your `try` block to call the functions (e.g. `dispatch['1']()`) and catch a `KeyError` with your `except`clause if the user enters any value that isn't a key in your dict

students/Matt_Hudgins/Lesson05/mailroom_3.py
```diff
@@ -0,0 +1,132 @@
...
+                break
+        except TypeError:
+            print("Invalid response please enter a number 0-4")
+
+
+def quit():
+    """ This function exits the executable"""
+    print("Thanks for your time!")
+    return 'exit menu'
+    sys.exit()
```
robalford: `sys.exit()` should be above the return statement or it will be unreachable

PR #616, opened 2018-05-21
Mailroom V2

willhwhite:
Mailroom V2

### DIFF
--- a/students/WWhite/lesson04/Mailroom.py
+++ b/students/WWhite/lesson04/Mailroom.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python3
+
+# -------------------------------------#
+# Desc: Mailroom Part 1
+# Dev: Will White
+# Date: 5/8/2018
+# ChangeLog: (When,Who,What)
+# -------------------------------------#
+
+
+list_donations = [["bill gates", 25000.00, 3], ["monet holt", 50000.00, 2], ["jeff bezos", 123500.09, 3],
+                  ["john wick", 120000.00, 2], ["john snow", 10.56, 1]]  # List of donors
+
+
+def menu_options():  # Function to run the menu options
+    user_input = input('''
+Please enter a number from the following options:
+
+[1] Send a Thank You
+[2] Create a Report
+[3] Send Letters to Everyone
+[4] Quit the Program
+
+''')
+    return user_input  # Function returns user_input
+
+
+def prompt_for_name():  # Function to prompt the user for a name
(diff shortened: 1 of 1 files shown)


students/WWhite/lesson04/Mailroom.py
```diff
@@ -0,0 +1,112 @@
...
+# Desc: Mailroom Part 1
+# Dev: Will White
+# Date: 5/8/2018
+# ChangeLog: (When,Who,What)
+# -------------------------------------#
+
+
+list_donations = [["bill gates", 25000.00, 3], ["monet holt", 50000.00, 2], ["jeff bezos", 123500.09, 3],
+                  ["john wick", 120000.00, 2], ["john snow", 10.56, 1]]  # List of donors
+
```
robalford: This would be a great place to use a dict, as your main data structure, something like this:
```
donors = {
    'bill gates': [25000.00, 3],
}
```
I think you'll find that this simplifies a lot of your code for looking up and adding donors and donations

students/WWhite/lesson04/Mailroom.py
```diff
@@ -0,0 +1,112 @@
...
+    sorted_list = sorted(list_donations, key=lambda x: x[1], reverse=True)
+
+    for i in sorted_list:
+        i[0] = i[0].title()
+        print('{:<10}{:>20.2f}{:>20}{:>20.2f}'.format(i[0], i[1], i[2], (i[1] / i[2])))
+
+
+def send_letters():
+    for i in list_donations:
+        txt_file = open("{}.txt".format(i[0]), 'w')
```
robalford: It's a best practice to use `with` statement for working with files, which will automatically close the file when you exit the with block

students/WWhite/lesson04/Mailroom.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+switch_func_dict = {
+    0: menu_options,
+    1: prompt_for_name,
+    2: add_donation,
+    3: create_a_report,
+    4: send_letters()
+}
+
```
robalford: Doesn't look like you're using the `0` option but I'd say it's better to call that function directly. These functions in your dispatch dict should map direclty to the options the user is provided by your `menu_options` function so that you can get and call the associated functions directly without all the if/elif statements in your `if __name__ == "__main__"` block

robalford:
@willhwhite remember to submit the assignment on the course website as well so we can grade. Thanks!

PR #618, opened 2018-05-21
lesson 09 - Mailroom OO

prgrover:


students/prgrover/lesson09/mailroom.py
```diff
@@ -0,0 +1,167 @@
...
+    @property
+    def donation_number(self):
+        return len(self.donations)
+
+    @property
+    def donation_average(self):
+        return self.donation_total / self.donation_number
+
+    def add_donation(self, donation_amount):
+        self.donations.append(donation_amount)
```
natasha-aleksandrova: 👍

students/prgrover/lesson09/mailroom.py
```diff
@@ -0,0 +1,167 @@
...
+
+        print("{:30s} | {:11s} | {:9s} | {:12s}".format("Donor Name", "Total Given", "Num Donations", "Average Donation"))
+        print("-" * 79)
+
+        for index in range(len(sorted_donor_list)):
+            row = sorted_donor_list[index]
+            print("{:30s}  ${:11.2f}   {:12d}    ${:12.2f}".format(*row))
+
+
+    def send_mail(self, donor_name, donation_amount):
```
natasha-aleksandrova: suggested name -> `get_letter` and this would be better to live in `Donor` class since it is dealing with single donor only.

PR #620, opened 2018-05-21
Finished lessons 1 to 5

yixingxu:
Looks like I can only use git push on one laptop, not both of my laptops. I use dropbox to sync two folders but git push only work on one of my laptop. Is it possible to use two laptops to store two clones in local and use git push from both laptops?


students/yixingxu/lesson05/mailroom3.py
```diff
@@ -0,0 +1,145 @@
...
+            wf.write(Letter)
+    print("\nAll letters written to files")
+   
+    
+if __name__ == "__main__":
+    main_dispatch = {
+                '1': send_a_thank_you,
+                '2': create_a_report,
+                '3': send_letters_to_everyone,
+                '4': Quit
```
natasha-aleksandrova: should be lower case variable and to avoid colliding with built-in `quit` you can name it `quit_program`

students/yixingxu/lesson05/mailroom3.py
```diff
@@ -0,0 +1,145 @@
...
+    sorted_list = summary_list[:]
+    sort_list(sorted_list)
+    print_report(sorted_list)
+    
+# print a list
+def print_report(list):
+    Donor_name_width = 20
+    Total_given_width = 15
+    Num_gifts_width = 5
+    Average_gift_width = 15
```
natasha-aleksandrova: make sure to use all lower case for var names

students/yixingxu/lesson05/mailroom3.py
```diff
@@ -0,0 +1,145 @@
...
+    print(f"\nDear {name},\n"
+        f"Thanks for your generous donation of ${donation_amount}"
+        )
+
+# construct list of names
+def list_name(donation_history = donation_history):
+    namelist = []
+    for item in donation_history:
+        namelist.append(item['name'])
+    return namelist
```
natasha-aleksandrova: this would be a great candidate for list comprehension

students/yixingxu/lesson05/mailroom3.py
```diff
@@ -0,0 +1,145 @@
...
+
+import sys # import sys, so i can use sys.exit(0) later
+import io
+donation_history = [
+                    {'name':'Anna','donations':[100,200,300]},
+                    {'name':'Bob','donations':[1000,2000]},
+                    {'name':'Chuck','donations':[10000]},
+                    {'name':'David','donations':[1]},
+                    {'name':'Ethan','donations':[10,20]}
+                 ]
```
natasha-aleksandrova: another data structure to make things easier:
```
donors = {"Anna": [100, 200, 300], "Bob": [2000]}
```
donor's first name is a natural key in this case so it makes sense to make it the dict key since it will make it easier for us to find the donor.

students/yixingxu/lesson03/Slicing_Exercise.py
```diff
@@ -0,0 +1,57 @@
...
+	a_new_sequence = seq[-1::-1]
+	return a_new_sequence	
+	
+def mid_last_first(seq):
+	# with the middle third, then last third, then the first third in the new order.
+	first_third = seq[:len(seq)//3]
+	mid_third = seq[len(seq)//3:-len(seq)//3]
+	last_third = seq[-len(seq)//3:]
+	a_new_sequence = mid_third + last_third + first_third
+	return a_new_sequence
```
natasha-aleksandrova: make sure indent is set to 4 spaces (no tabs)

I noticed later assignments had good indentation so it might be something you have fixed already 🙂

yixingxu:
Thanks a lot for your feedback. I will fix them in the later assignments.
Yes, I noticed that my tab and spaces are not set correctly in my editor
and fixed it later.


Sincerely,
Xu, Yixing

PR #622, opened 2018-05-22
finished lesson 6

yixingxu:
used pytest to do the testing


natasha-aleksandrova:
@yixingxu please take a look at the files in this PR.

before running `git add .` it is always a good idea to run `git status`

I cannot merge this PR with the files outside of your student directory as this may cause issues for other students.

You may try `git revert` or `git rm`

students/yixingxu/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,46 @@
...
+    assert (expected == mailroom4.create_a_report()) is True
+    
+def test_create_summary():
+    donation_history = [
+                    {'name':'Anna','donations':[100,200,300]},
+                    {'name':'Bob','donations':[1000,2000]},
+                    {'name':'Chuck','donations':[10000]},
+                    {'name':'David','donations':[1]},
+                    {'name':'Ethan','donations':[10,20]}  
+                    ]
```
natasha-aleksandrova: this isn't used.

students/yixingxu/lesson06/mailroom4.py
```diff
@@ -0,0 +1,145 @@
...
+    return list[2]
+    
+# quit
+def quit_program():
+    print('Quit')
+    return 'exit menu'
+
+def send_letters_to_everyone(donation_history = donation_history):
+    for item in donation_history:     
+        Letter = "Dear {name},\n\n        Thank you for all your very kind donations of ${donations}\n\n        It will be put to very good use.\n\n                       Sincerely,\n                          -The Team".format(**item)
```
natasha-aleksandrova: make sure to use lower case for var names.

students/yixingxu/lesson06/mailroom4.py
```diff
@@ -0,0 +1,145 @@
...
+# get the total given amount 
+def sort_by_total_given(list):
+    return list[2]
+    
+# quit
+def quit_program():
+    print('Quit')
+    return 'exit menu'
+
+def send_letters_to_everyone(donation_history = donation_history):
```
natasha-aleksandrova: this would be a good function to test

students/yixingxu/lesson06/mailroom4.py
yixingxu: missed one. Thanks. fixed this one.

students/yixingxu/lesson06/mailroom4.py
yixingxu: can you give me a hint. How to test written out files? Thanks!

natasha-aleksandrova: in your test call `send_letters_to_everyone()`
then assert that files exists via `os.path.isfile`, in addition you can also open the file and assert contents of the file

students/yixingxu/lesson06/test_mailroom4.py
yixingxu: function create_summary() has a default input, which i why I put the default input donation_history in the test function. create_summary() is from line 99 to line 106 in mailroom4.py

students/yixingxu/lesson06/mailroom4.py
yixingxu: Thanks!

students/yixingxu/lesson06/test_mailroom4.py
natasha-aleksandrova: screen shot 2018-05-22 at 3 31 35 pm

PR #623, opened 2018-05-22
Lesson 7 Assignments

TrkCat:
Ready for review.

Thanks!


students/TrkCat/Lesson07/html_render.py
```diff
@@ -0,0 +1,116 @@
+class Element:
+    tag = 'html'
+    indent = 4
+
+    def __init__(self, content=None, **kwargs):
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
```
robalford: This could be a one liner, e.g:
```
self.content = [] if content is None else [content]
```

students/TrkCat/Lesson07/html_render.py
```diff
@@ -0,0 +1,116 @@
...
+        ind = 0 if cur_ind is None else cur_ind
+        for item in self.content:
+            if hasattr(item, 'render'):
+                item.render(file_out, ind + 1)
+            else:
+                file_out.write(self.conv_ind(ind + 1) + item + '\n')
+        file_out.write(self.conv_ind(cur_ind))
+        self.render_close_tag(file_out)
+
+    def render_open_tag(self, file_out, cur_ind=None, attr={}):
```
robalford: Be careful with mutable default arguments, better to set `attr` to `None` and handle with the same pattern you use for `content` argument on `__init__`

students/TrkCat/Lesson07/html_render.py
```diff
@@ -0,0 +1,116 @@
...
+        self.render_close_tag(file_out)
+
+
+class Title(OneLineTag):
+    tag = 'title'
+
+
+class SelfClosingTag(Element):
+    def __init__(self):
+        pass
```
robalford: It would be better to raise an error if the user passes in content (this could be done by overriding the `append` method), then you can still use the attribute `kwargs` and don't need to override `__init__` on `Meta`

PR #626, opened 2018-05-22
submit mailroom_fn.py

tricaonguyen:


students/tri_nguyen/lesson10/mailroom_fn.py
```diff
@@ -0,0 +1,296 @@
...
+
+def num_gifts(donations):
+    ''' return how many times a donor has donated '''
+    return len(donations)
+
+
+def average(donations):
+    ''' return the average of all the amount 
+        of money that a donor has donated '''
+    return sum(donations) / len(donations)
```
natasha-aleksandrova: when you have several functions with the same input param it is usually indication that this should be a class

PR #627, opened 2018-05-22
Lesson07, Lesson08, Lesson09

DeFeNdog:
- Lesson07 assignment and edits
- Lesson08 edits


students/carlos_novoa/lesson09/mailroom.py
```diff
@@ -0,0 +1,337 @@
...
+_donors = {
+    'Jim Halpert': [1.00],
+    'Pam Beesley': [1000.00, 2000.00, 3000.00],
+    'Dwight Shrute': [2.00, 3.00],
+    'Michael Scott': [10.00, 20.00, 30.00],
+    'Andy Bernard': [500.00]
+}
+
+
+class Helpers():
```
robalford: Don't need the parens for class definition

students/carlos_novoa/lesson09/mailroom.py
```diff
@@ -0,0 +1,337 @@
...
+    def format_currency_str(amount=None):
+        return "${0:.2f}".format(float(amount))
```
robalford: Instance methods automatically pass in the object instance as the first argument, and this is conventionally named 'self', so here `amount` is actually the `Helpers` instance. You can avoid this by making it a static method, but since this is the only method on the class, and there aren't any attributes this should really be a module level function or a method on one of your other classes

students/carlos_novoa/lesson09/mailroom.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+class Donors:
+    donors = None
```
robalford: You're defining this as an instance attribute below

students/carlos_novoa/lesson09/mailroom.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def __init__(self, donors=None):
+        self.donors = donors if donors else _donors
```
robalford: It would be better to make donors be always be an instance attribute created from the `__init__` parameter, rather than allowing the attribute to default to the value of a global variable. One of the main reasons for using classes is to encapsulate related data (attributes) and actions/behavior (methods) that use that data inside the class

students/carlos_novoa/lesson09/mailroom.py
```diff
@@ -0,0 +1,337 @@
...
+    def donations_count(self, name=None):
+        return len(self.donors[name])
+
+    def donations_average(self, name=None):
+        total = sum(self.donors[name])
+        count = len(self.donors[name])
+        return Helpers.format_currency_str(total / count)
+
+
+class Donor(Donors):
```
robalford: Subclassing should almost always be used for 'is a' relationships, but 'a donor is a donors' does not make sense so this is a good indication that inheritance isn't the right relationship to establish between these classes. 

I would recommend having a Donor class that encapsulates all the logic and data related to a single donor, and a Donors class for logic and data related to the collection of donors as a whole, so you're on the right track, but you shouldn't need Donor to inherit all the behavior and attributes from Donors

students/carlos_novoa/lesson09/mailroom.py
```diff
@@ -0,0 +1,337 @@
...
+            else:
+                print('Donor already exists, use append.')
+
+        donation = self.prompt_donation()
+        self.thank_donor(name, donation)
+        d.donors_update(name, donation)
+
+    def process_donation(self):
+        D = Donors()
+        d = Donor()
```
robalford: This is pretty confusing to use 'D' and 'd' as variable names and also a good indication that there isn't enough separating the roles of these classes in your code. See my comment above on reworking these classes for better encapsulation and separation of concerns

PR #630, opened 2018-05-23
Lesson 8: Circle

itsabhister:


students/itsabhi/Lesson8/circle.py
```diff
@@ -0,0 +1,116 @@
...
+class Circle:
+    # Circle class
+    _diameter = None  # This cannot be stored as its not an object attribute
+    _area = None  # This cannot be stored as its not an object attribute
+
+    @property
+    def diameter(self):
+        return self.radius*2
+
+    def __init__(self, the_radius):
```
natasha-aleksandrova: init should always be the first method you define

students/itsabhi/Lesson8/circle.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python3
+
+import math
+
+
+class Circle:
+    # Circle class
+    _diameter = None  # This cannot be stored as its not an object attribute
+    _area = None  # This cannot be stored as its not an object attribute
```
natasha-aleksandrova: what is the purpose of these?

students/itsabhi/Lesson8/circle.py
```diff
@@ -0,0 +1,116 @@
...
+    def area(self):
+        return (self.radius**2)*math.pi
+
+    @classmethod
+    def from_diameter(cls, diam):
+        # Returns a Circle class object
+        return cls(diam/2)
+
+    def __str__(self):
+        return("Circle with radius: "+str(self.radius))
```
natasha-aleksandrova: you don't need parens with return and don't forget proper spacing:
`return "Circle with radius: " + str(self.radius)`

students/itsabhi/Lesson8/circle.py
itsabhister: I might have copied these from the class lesson example:
class C:
    _x = None
    @property
    def x(self):
        return self._x

PR #635, opened 2018-05-24
Modified Trigram.py code

shibinmathew:
Modified trigram.py and tested with sherlock_small.txt file.


students/shibin_mathew/lesson4/trigram.py
```diff
@@ -1,34 +1,57 @@
...
+def generate_story(sample_story, word_dict):
+    # Generate a story based on the trigram dictionary
+    rand_start = int(random.random()*len(word_dict))  # random starting point
+    story = ''
+    for i in range(len(sample_story)-2):
+        try:
+            if sample_story[i] + ' ' + sample_story[i+1] in word_dict.keys():
+                # print(sample_story[i] + ' ' + sample_story[i+1])
+                if i == 0:
+                    story += sample_story[i] + ' ' + sample_story[i+1] + ' ' + word_dict[sample_story[i] + ' ' +
```
robalford: The assignment asks to begin with an arbitrary word pair, and then add a random following word from your `word_dict', but here (use random module), you're starting with the first three words of the input text

students/shibin_mathew/lesson4/trigram.py
```diff
@@ -1,34 +1,57 @@
...
+                                                                                         sample_story[i+1]][0] + ' '
+                    try:
+                        word_dict[sample_story[i] + ' ' + sample_story[i+1]].remove(word_dict[sample_story[i] + ' ' +
+                                                                                              sample_story[i+1]][0])
+                    except KeyError:
+                        print("{} is not in the list".format(word_dict[sample_story[i] + ' ' + sample_story[i+1]]))
+                else:
+                    story += word_dict[sample_story[i] + ' ' + sample_story[i+1]][0] + ' '
```
robalford: This adds the next word from the original input text to `story` each time through the loop, so your code is regenerating the exact same text that you're feeding it. Each time through the loop, you should use the last two words of your generated text to look up a new random following word in the `word_dict` each time through the loop and append that to `story`, so it builds a new randomly generated text.

PR #636, opened 2018-05-24
Lesson 9 complete and ready for review.

rhamersky:
Lesson 9 is ready for review.

### DIFF
--- a/students/rhamersky/Lesson_9/mailroom_part5.py
+++ b/students/rhamersky/Lesson_9/mailroom_part5.py
@@ -1,0 +1,279 @@
+#!/usr/bin/env python3
+# Description: This program will allow a user to input a donor and a donation.
+# The user will also be able to send out thank you notes to the donors.
+# Developer: Ryan Hamersky
+# Date: 05/21/2018
+# Rev: New
+
+# -----Data Section-----
+str_name = str()    # --> User's name
+int_user_input = int()  # --> User's option.
+dic_donors = {"William Gates, III": [10000.0, 1], "Mark Zuckerberg": [20000.0, 1],
+                   "Jeff Bezos": [1200.0,1],"Paul Allen": [5000.0,1],
+                   "Mike Scott": [25.0,1]} # --> Dic that stores all donor's information
+str_salutation = ""   # --> Email salutation
+str_body = ""         # --> Email body
+str_valediction = ""  # --> Email valediction
+
+
+# -----Process Section-----
+# This class creates an donor object.
+class Donor:
+    initial_donation = 1  # --> Field is used as the initial donation is constant with all donor objects.
+
+    def __init__(self, donor, donation):  # --> Constructs attributes donor and donation.
+        self.donor = donor
+        self.donation = donation
+
+########### END OF DONOR OBJECT CLASS###############
+
+# Methods in the Menu class will execute code based off of the user's menu choice.
+
+
+class Menu:
+    @staticmethod
+    def menu_options():
+        '''
+        Try and except block used on User's option choice to check for integers.
+        :return: Returns int data (1-4) which is used to determine what menu option will be executed.
+        '''
+        while True:
+            try:
+                int_choice = int(input(str_name + ", please select an option above. "))
+            except ValueError:
+                print(str_name + ", not a valid option. \n")
+            # Verify that the user selected an option 1 through option 4.
+            else:
+                if 1 <= int_choice <= 4:
+                    break
+                else:
+                    print("Integer selected is not a valid option. "
+                          "Please, enter an integer and corresponds to the options given. \n")
+        return int_choice
+
+    @staticmethod
+    def send_thankyou():
+        '''
+        This method will  allow new donors to be added, store donations, and print a thank you
+        statement to send to the donor by mail.
+        :return:
+        '''
+        donor_name = input("Please, enter full name of donor. ").title()
+        global dic_donors
+
+        # Gathers donation information from the user and prints a thank-you message for the donor.
+        while True:
+            # Appends the donor list.
+            if donor_name == "List":
+                # --> Verifies that there is at least a donor is stored as a list row in the table tuple.
+                if len(dic_donors) > 0:
+                    Data.print_donor_names()  # --> Prints a list of donors
+                    break
+                else:
+                    print(
+                        "\n" + "There are no donors in list.")  # --> Let's the user know that a donor list does not exist.
+                    break
+
+            try:
+                flo_donation_amount = float(input(str_name + ", please enter donation amount. "))
+
+                # Updates donor tuple
+                new_donor = Donor(donor_name,flo_donation_amount)
+                dic_donors = Data.update_donor(new_donor.donor, new_donor.donation, new_donor.initial_donation)
+
+                # String formatting for a brief thank-you to donor.
+                thankyou_message = "\n"+"{}, thank you for donating ${:,.2f}.".format(donor_name, flo_donation_amount)
+                print(thankyou_message)
+                break
+
+            except ValueError:  # --> Allows the program to continue if the user does not enter the right data type.
+                print(str_name + ", not a valid amount. \n")
+                continue
+
+########### END OF MENU CLASS###############
+
+# Methods defined in the files class are related to opening and renaming files.
+# Modifications to the methods are needed if the files are not for a task priority list.
+
+
+class Files:
+    @staticmethod
+    def write_email(donors):
+        '''
+        Writes emails to the donors and saves them to text files.
+        :param donors: Dictionary that stores all the donor's donation information.
+        :return:
+        '''
+        for donor_name in donors:
+            # Open text file and assign a mode to the file.
+            new_file = open(donor_name + ".txt", mode="w")
+            new_file.write(Files.donor_email(donor_name))
+            new_file.close()
+            print(donor_name + " saved to "+donor_name+".txt.")
+
+    @staticmethod
+    def donor_email(donor):
+        '''
+        Formats the email with donor's name and total lifetime donations.
+        :param donor: Dictionary that stores all the donor's donation information.
+        :param dictionary: A dictionary that stores the donor as the 'key' and the lifetime donation as the 'value'.
+        :return: The concatenation how the three strings.
+        '''
+        str_salutation = "Dear {},".format(donor)
+        str_body = "\n\n"+"Thank you for generosity! " \
+                   "\n\n"+"This donation will be put to great use!" \
+                   "\n\n"+"Your current lifetime donation is "+"${:,.02f}".format(dic_donors[donor][0]) + "."
+        str_valediction = "\n\n"+"Sincerely," \
+                          "\n\n"+"The Team"
+        thankyou_message = str_salutation+str_body+str_valediction
+        return thankyou_message
+
+############# END OF FILE CLASS #################
+
+# The Data class is where the data get manipulated.
+
+
+class Data:
+    @staticmethod
+    def update_donor(donor_name, donation_amount, initial_donation):
+        '''
+        Updates donor's donation history or creates a new donor.
+        :param donor_name: The name of the donor.
+        :param donation_amount: The amount of money the donor donated.
+        :param initial_donation: The initial donation of the donor.
+        :return:
+        '''
+        # Creates a donor name list
+        # lst_donor_name = [dic_donors[row][0] for row in range(len(dic_donors))]  # --> Added list comprehension
+        # Checks to see if donor has given before and updates history
+        dic_new_donor_names = dic_donors
+        if donor_name in dic_donors:
+            dic_new_donor_names[donor_name][1] = dic_donors[donor_name][
+                                                     1] + 1  # --> Increase amount of times donor donated
+            dic_new_donor_names[donor_name][0] = dic_donors[donor_name][
+                                                     0] + donation_amount  # --> Adds donation to history
+            dic_new_donor_names = Data.dic_sort(dic_donors)
+        # Adds new donor to the donor list
+        else:
+            dic_new_donor_names[donor_name] = [donation_amount, initial_donation]
+        return dic_new_donor_names
+
+    @staticmethod
+    def print_donor_names():
+        '''
+        Prints donors names if the user types "list" when the program prompts the user to input a donor.
+        :return:
+        '''
+        # Creates a list just of donor names
+        print("\n" + "Donor")
+        for name in dic_donors:
+            print(name)
+
+    @staticmethod
+    def dic_sort(dic_donors):
+        '''
+        Takes a dictionary and creates a sorted tuple using the first element of the value list.
+        It then puts the sorted data into a new dictionary.
+        :param dic_donors: The dictionary of donors.
+        :return:
+        '''
+        sorted_tuple = sorted(dic_donors.items(), key=lambda x: x[1], reverse=True)
+        sorted_dic = {}
+        for items in sorted_tuple:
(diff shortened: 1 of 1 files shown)


students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+# The user will also be able to send out thank you notes to the donors.
+# Developer: Ryan Hamersky
+# Date: 05/21/2018
+# Rev: New
+
+# -----Data Section-----
+str_name = str()    # --> User's name
+int_user_input = int()  # --> User's option.
+dic_donors_sorted = {}  # --> Sorted dictionary of donors.
+dic_donors = {"William Gates, III": [10000.0, 1], "Mark Zuckerberg": [20000.0, 1],
```
robalford: If you're going to indicate type in your variable names, I would use the actual type name rather than abbreviating further (dict to dic) for readability

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+                   "Jeff Bezos": [7800.0,4, 2500, 2000, 3000, 300],"Paul Allen": [5000.0,1],
+                   "Mike Scott": [25.0,1]} # --> Dic that stores all donor's information
+str_salutation = ""   # --> Email salutation
+str_body = ""         # --> Email body
+str_valediction = ""  # --> Email valediction
```
robalford: There's no reason to define these empty variables here in the global namespace, they should be defined closer to where they are used in your code, especially when they are only used within the scope of a specific function or method

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+# -----Process Section-----
+# This class creates an donor object.
+class Donor:
+    initial_donation = 1  # --> Field is used as the initial donation is constant with all donor objects.
+
+    def __init__(self, donor, donation):  # --> Constructs attributes donor and donation.
+        self.donor = donor
+        self.donation = donation
```
robalford: You could store the entire donation history here in a list as an attribute, and wouldn't need to track `initial_donation` separately

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+
+########### END OF DONOR OBJECT CLASS###############
+
+# Methods in the Menu class will execute code based off of the user's menu choice.
+
+
+class Menu:
```
robalford: In Python you would generally define these methods as module level functions rather than introducing the extra overhead of using a class with only static methods

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+        :return: Returns int data (1-4) which is used to determine what menu option will be executed.
+        '''
+        while True:
+            try:
+                int_choice = int(input(str_name + ", please select an option above. "))
+            except ValueError:
+                print(str_name + ", not a valid option. \n")
+            # Verify that the user selected an option 1 through option 4.
+            else:
+                if 1 <= int_choice <= 5:
```
robalford: You could be using a [dict as switch](https://uwpce-pythoncert.github.io/PythonCertDevel/modules/DictionaryAsSwitch.html#dict-as-switch) to map user input to dict keys and you wouldn't need this extra validation (you could just catch `Key_Error` on bad user input for any value that isn't a key in the dict)

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+
+    @staticmethod
+    def send_thankyou():
+        '''
+        This method will  allow new donors to be added, store donations, and print a thank you
+        statement to send to the donor by mail.
+        :return:
+        '''
+        donor_name = input("Please, enter full name of donor. ").title()
+        global dic_donors
```
robalford: Don't think you need to declare global here

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+
+        # Gathers donation information from the user and prints a thank-you message for the donor.
+        while True:
+            # Appends the donor list.
+            if donor_name == "List":
+                # --> Verifies that there is at least a donor is stored as a list row in the table tuple.
+                if len(dic_donors) > 0:
```
robalford: This should be `if dic_donors` (empty list evaluates to False)

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+                flo_donation_amount = float(input(str_name + ", please enter donation amount. "))
+
+                # Updates donor tuple
+                new_donor = Donor(donor_name,flo_donation_amount)
+                dic_donors = Data.update_donor(new_donor.donor, new_donor.donation, new_donor.initial_donation)
+
+                # String formatting for a brief thank-you to donor.
+                thankyou_message = "\n"+"{}, thank you for donating ${:,.2f}.".format(donor_name, flo_donation_amount)
+                print(thankyou_message)
+                break
```
robalford: You should include as little code as possible in a `try` block and only the code that might raise the expected exception, so you don't accidentally catch unanticipated errors which can lead to some difficult debugging

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+
+            except ValueError:  # --> Allows the program to continue if the user does not enter the right data type.
+                print(str_name + ", not a valid amount. \n")
+                continue
```
robalford: Don't need this `continue`

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+        :return: Returns donor name to main function.
+        '''
+        while True:
+            try:
+                str_user_input = input(str_name + ", please enter a donor name to do a projection on. ").title()
+            except ValueError:
+                print(str_name + ", not a valid option. \n")
+                # Verifies that the data type is part of the string class and that the donor name is in the donor dic.
+            else:
+                if str_user_input.__class__ == str().__class__ and str_user_input in dic_donors:
```
robalford: You can use the built-in `isinstance()` function to check for type, but `input()` will always return a string so you don't need to validate type here

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+
+    @staticmethod
+    def user_factor_input():
+        '''
+        Try and except block used on User's option choice to check for integers.
+        :return: Returns factor input to the main function.
+        '''
+        while True:
+            try:
+                int_user_input = float(input(str_name + ", please enter a positive integer. "))
```
robalford: It's misleading to name this `int_user_input` when the type is a `float`

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+            except ValueError:
+                print(str_name + ", not a valid option. \n")
+                # Verifies that the data type is part of the int class and that the user input a value greater than 0.
+            else:
+                if int_user_input.__class__ == float().__class__ and int_user_input > 0:
```
robalford: Don't need to check for `float` here since you converted in the `try` block and you will never reach this `else` clause unless the value was successfully converted to a `float`.

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+        '''
+        Try and except block used on User's option choice to check for integers.
+        :return: Returns filter data to the main function.
+        '''
+        while True:
+            try:
+                str_user_input = input(str_name + ", please enter max to filter for amounts greater than or equal "
+                                                  "to $2,500. \n or min to filter for "
+                                                  "amounts less than $2,500. ").capitalize()
+            except ValueError:
```
robalford: Not sure if it's necessary to catch a `ValueError` here since `input()` will always convert return value to a string

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+    @staticmethod
+    def write_email(donors):
+        '''
+        Writes emails to the donors and saves them to text files.
+        :param donors: Dictionary that stores all the donor's donation information.
+        :return:
+        '''
+        for donor_name in donors:
+            # Open text file and assign a mode to the file.
+            new_file = open(donor_name + ".txt", mode="w")
```
robalford: It's a best practice to use a `with` statement for dealing with files and this will automatically close the file for you when you're done with it

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+        :param dictionary: A dictionary that stores the donor as the 'key' and the lifetime donation as the 'value'.
+        :return: The concatenation how the three strings.
+        '''
+        str_salutation = "Dear {},".format(donor)
+        str_body = "\n\n"+"Thank you for generosity! " \
+                   "\n\n"+"This donation will be put to great use!" \
+                   "\n\n"+"Your current lifetime donation is "+"${:,.02f}".format(dic_donors[donor][0]) + "."
+        str_valediction = "\n\n"+"Sincerely," \
+                          "\n\n"+"The Team"
+        thankyou_message = str_salutation+str_body+str_valediction
```
robalford: I would use a single triple-double-quoted (multi-line) string here rather than concatenating multiple smaller strings

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+        return thankyou_message
+
+############# END OF FILE CLASS #################
+
+# The Data class is where the data get manipulated.
+
+
+class Data:
```
robalford: In general, it's better to use class/function/variable names that are more specific to their use case in your program e.g. `DonorsData`

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+        Updates donor's donation history or creates a new donor.
+        :param donor_name: The name of the donor.
+        :param donation_amount: The amount of money the donor donated.
+        :param initial_donation: The initial donation of the donor.
+        :return:
+        '''
+        # Creates a donor name list
+        # lst_donor_name = [dic_donors[row][0] for row in range(len(dic_donors))]  # --> Added list comprehension
+        # Checks to see if donor has given before and updates history
+        dic_new_donor_names = dic_donors
```
robalford: I'd recommend making the donors dict an attribute of this class rather than module level variable for better encapsulation

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+        if donor_name in dic_donors:
+            dic_new_donor_names[donor_name][1] = dic_donors[donor_name][
+                                                     1] + 1  # --> Increase amount of times donor donated
```
robalford: You don't need to keep track of the total and initial donations in the list. If you store only the individual donations in a list, you can use `sum()` for total and keep the donations sorted in the order they were donated to access the initial donation by `0` index. 

If you still want to keep track of these calculated values elsewhere on a per donor basis (which is a great idea), it would be best to use `@properties` on your Donor objects

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+            dic_new_donor_names[donor_name][0] = dic_donors[donor_name][
+                                                     0] + donation_amount  # --> Adds donation to history
+            lst_donor_data = dic_new_donor_names[donor_name]  # --> Creates a list of the values in the dictionary.
+            lst_donor_data.append(donation_amount)  # --> Appends donation amount to donor data list
+            dic_new_donor_names[donor_name] = lst_donor_data  # --> Replaces the old value list with lst_donor_data.
```
robalford: You can append the `donation_amount` to the original list (`dic_new_donor_names[donor_name].append(donation_amount)`) without copying, appending and replacing the list in the dict

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+        Takes a dictionary and creates a sorted tuple using the first element of the value list.
+        It then puts the sorted data into a new dictionary.
+        :param dic_donors: The dictionary of donors.
+        :return:
+        '''
+        dic_donors_total_donation = {}
+        # Removes all past donations from the value list.
+        for key in dic_donors:
+            lst_total_donation = dic_donors[key]
+            dic_donors_total_donation[key] = lst_total_donation
```
robalford: Is this loop making an exact copy of the dict? If so, you don't need a loop for that. Also not clear on why you need a copy here

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+        sorted_tuple = sorted(dic_donors_total_donation.items(), key=lambda x: x[1], reverse=True)
```
robalford: Is there a reason to convert to a `tuple` and back rather than sorting the `dict`?

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+        :param dic_donors: The dictionary of donors.
+        :return:
+        '''
+        longest_entry = float()
+        # Finds the max amount of characters for all items stored in dic_donors.
+        for name in dic_donors:
+            entry_length = len(name)
+            if longest_entry < entry_length:
+                longest_entry = entry_length
+            for index in range(0,1):
```
robalford: Don't need to use a loop with a single iteration, also the first arg of `range()` defaults to `0`

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+            lst_values = filter(lambda filter_value: filter_value >= 2500, lst_values)
+        else:
+            lst_values = filter(lambda filter_value: filter_value < 2500, lst_values)
+
+        # Adds a factor to the filtered donations.
+        factored_donations = map(lambda donation: donation * factor, lst_values)
+
+        # Sums factored donations
+        for past_donation in list(factored_donations):
+            new_factored_donation += past_donation
```
robalford: This would be a good place to use `sum() or `reduce()`

students/rhamersky/Lesson_10/mailroom_part6.py
```diff
@@ -0,0 +1,392 @@
...
+              "\t\t5) Quit\n")
+
+        # Error handles user's option choice.
+        int_user_input = Menu.menu_options()
+
+        # User selects option 1. Can display a list of the donors and where the user can add donors and donations.
+        if int_user_input == 1:
+            print()  # --> Formatting
+            Menu.send_thankyou()
+            continue
```
robalford: You should be using a dict to switch between user selections at this point per the instructions in Lesson 4. This would really simplify this section of code as well as some of the related validation logic for use input

PR #637, opened 2018-05-24
Lesson07 Assignment - HTML Render

TinaBrownfield:
Adding lesson 07 files: Test_html_render, html_render, and run_html_render files.
I don't believe my tests were robust enough - but I also didn't want to spend days just working on this testing file.

### DIFF
--- a/students/TinaB/lesson07/html_render.py
+++ b/students/TinaB/lesson07/html_render.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+
+class TextWrapper:
+    """
+    A simple wrapper that creates a class with a render method
+    for simple text
+    """
+    def __init__(self, text):
+        self.text = text
+
+    def render(self, file_out, current_ind=""):
+        file_out.write(current_ind)
+        file_out.write(self.text)
+
+class Element():
+    indent = '    '
+    tag = 'html'
+
+    def __init__(self, content=None, **attributes):
+        self.content = [content] if content else []
+        self.attributes = attributes
+
+    def append(self, content_string):
+        self.content.append(content_string)
+        # if hasattr(content_string, 'render'):
+        #     self.content.append(content_string)
+        # else:
+        #     self.content.append(TextWrapper(str(content_string)))
+
+    def render(self, file_out, cur_indent=''):
+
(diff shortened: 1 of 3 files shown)


students/TinaB/lesson07/html_render.py
```diff
@@ -0,0 +1,126 @@
...
+class Body(Element):
+    tag = 'body'
+
+class P(Element):
+    tag = 'p'
+
+class Head(Element):
+    tag = 'head'
+
+class OneLineElement(Element):
```
natasha-aleksandrova: need two lines space between classes (pep8)

students/TinaB/lesson07/test_html_render.py
```diff
@@ -0,0 +1,91 @@
...
+        self.assertEqual(test_element.content, [
+                         "this is some content", "Adding additional content"])
+
+    def test_element_tag(self):
+        """Testing override tag properties"""
+        self.assertEqual(self.body.tag, 'body')
+        self.assertEqual(self.head.tag, 'head')
+        self.assertEqual(self.paragraph.tag, 'p')
+        self.assertEqual(self.horizontal_rule.tag, 'hr')
+        self.assertEqual(self.anchor.tag, 'a')
```
natasha-aleksandrova: it would be better to test each tag as far as rendering vs just testing the attribute.

students/TinaB/lesson07/test_html_render.py
```diff
@@ -0,0 +1,91 @@
...
+    def setUp(self):
+        self.html = hr.Html()
+        self.head = hr.Head(hr.Title('The Title is Title'))
+        self.body = hr.Body()
+        self.paragraph = hr.P('This is my paragraph', style="color:blue;")
+        self.horizontal_rule = hr.Hr()
+        self.anchor = hr.A('http://chicktech.org', 'link')
+        self.unorderedlist = hr.Ul(id="UL_List", style="line-height:100%")
+        self.headertwo = hr.H(3, 'This is a Header 3 ')
+        self.meta = hr.Meta(charset="UTF-8")
```
natasha-aleksandrova: it probably makes sense to instantiate as part of each new test and not in setup here because you aren't actualy using this in each test.

PR #641, opened 2018-05-25
Lessons 6-8 

ChrisKenyon89:


students/Chris_Kenyon/Lesson_6/Test_Mailroom.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
+
+# Lesson_6 Activity 1 Test Mailroom Part 4
+
+from Mailroom_Pt4 import *
```
natasha-aleksandrova: it is preferred to use explicit imports rather than "star" imports, also python modules should be lower case -> `mailroom_pt4`

students/Chris_Kenyon/Lesson_6/Test_Mailroom.py
```diff
@@ -0,0 +1,39 @@
...
+    "Kenyon\\UWPython\\Testing_File_dump"
+    send_letters(test_dump)
+    for item in donor_chart.items():
+        assert os.path.isfile((os.path.join(test_dump, item[0] + ".txt")))
+
+
+def test_get_key():
+    for donor in donor_chart.items():
+        sumlist = get_key(donor)
+        assert sumlist == sum(donor[1])
```
natasha-aleksandrova: tests look good 👍

students/Chris_Kenyon/Lesson_7/html_render.py
```diff
@@ -0,0 +1,120 @@
...
+class H(OneLineTag):
+
+    def __init__(self, size, content, **kwargs):
+        OneLineTag.__init__(self, content, **kwargs)
+        self.size = size
+        self.tag = "h{}".format(size)
+
+
+class Meta(SelfClosingTag):
+    tag = 'meta'
```
natasha-aleksandrova: html renderer looks good too

students/Chris_Kenyon/Lesson_8/circle.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @diameter.setter
+    def diameter(self, value):
+        # make sure the radius changes when we change diameter
+        self.radius = value / 2
+
+    @property
+    def area(self):
+        """ Read Only """
+        return self._area
```
natasha-aleksandrova: user can technically overwrite `self._area` and mess up values; it might be better to exclusively use setter and getter properties and not define area attribute.

PR #647, opened 2018-05-27
Lesson05 assignments ready for review

seaneightysix:
Thank you :)


students/Sean_Tasaki/Lesson06/mailroom_part4.py
```diff
@@ -0,0 +1,106 @@
...
+        thank_you()
+    elif name.upper() == 'Q':
+        main_menu()
+    else:
+        add_donor_to_dict(name.title())
+    try:
+        if donor_dict[name.title()][1] == 0:
+            amount = input(f"{name.title()} is a new donor. Enter the donation amount.\n>> ")
+            add_donor_amount(name.title(), amount)
+            print(f"Thank you {name.title()} for becoming a new donor to our charity! Your genereous donation of ${float(amount):.2f} is much appreciated.")
```
robalford: I would include a separate function for generating the thank you letter as a string, and print the output of that function here -- that way you can write a unit test for the letter function to make sure that it generates the expected output for a specific donor

students/Sean_Tasaki/Lesson06/mailroom_part4.py
```diff
@@ -0,0 +1,106 @@
...
+        else:
+            amount = input(f"{name.title()} is a previous donor. Enter the donation amount.\n>> ")
+            add_donor_amount(name.title(), amount)
+            print(f"Thank you {name.title()} for your loyal support to our charity! Your genereous donation of ${float(amount):.2f} is much appreciated.")
+
```
robalford: You want to include as little code in your `try` block as possible, and only the code that you expect to raise the error that you are handling in your `except` block

students/Sean_Tasaki/Lesson06/mailroom_part4.py
```diff
@@ -0,0 +1,106 @@
...
+def create_report():
+    lis = []
+    lis = [[donor, values] for donor, values in donor_dict.items()]
+    # Sort lis by total given
+    lis.sort(key = lambda x: x[1][0], reverse = True) 
+    print("Donor Name           |   Total Given   |   Num Gifts  |    Average Gift")
+    print("-----------------------------------------------------------------------")
+    for donor in lis:
+        print(f"{donor[0]:20} ${donor[1][0]:>17.2f}    {donor[1][1]:>6}     ${donor[1][2]:>16.2f}")  
+    main_menu()
```
robalford: I would rework this function to return a string so that you can test it

students/Sean_Tasaki/Lesson06/mailroom_part4.py
```diff
@@ -0,0 +1,106 @@
...
+        main_menu()
+    finally: 
+        for file in file_object_for_donations:
+            file.close()
+def thank_you_letter_template(dk):
+    if donor_dict[dk][1] > 1:
+        return "Dear {},\nThank you for your {} generous donations of ${:.2f}. Your support helps our charity stay in business.\n\nSincerely,\n-The Team".format(dk, donor_dict[dk][1], donor_dict[dk][0])
+    else:
+        return "Dear {},\nThank you for your generous donation of ${:.2f}. Your support helps our charity stay in business.\n\nSincerely,\n-The Team".format(dk, donor_dict[dk][1], donor_dict[dk][0])
+
```
robalford: You should be able to write a test for this function that the files get written to disk with the expected content

students/Sean_Tasaki/Lesson06/test_mailroom_part4.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+import mailroom_part4 as mailroom 
+from collections import defaultdict
+import pytest
+
+@pytest.fixture()
+def resource_a():
+    s = 'Sean'
+    return s
```
robalford: Generally fixtures would be used for more complex data than a single string and would be shared between multiple test cases

robalford:
@seaneightysix looks like you may still be working on Lesson 6, in which case you can look at my current suggestions while you're finishing it up. Remember to submit the assignment on the course website when you're done, and if you want you can open a separate PR for a final review.

PR #650, opened 2018-05-28
Lesson 3 assignments

vvinodh:


students/vvinodh/Lesson3/list_lab.py
```diff
@@ -0,0 +1,58 @@
+#Series 1
+fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+print (fruits)
+add_fruit = input("What fruit do you want to add? ")
+fruits.append(add_fruit)
+print (fruits)
+response = int(input("Type in a number "))
+print(response," - ",fruits[response:response+1])
```
natasha-aleksandrova: instead of using slicing, you can use indexing -> `fruits[response + 1]`

I think this is also causing a bug, if I type in `1` it gives me "Pears" instead of "Apples"

students/vvinodh/Lesson3/list_lab.py
```diff
@@ -0,0 +1,58 @@
...
+fruits = ['Mango'] + fruits
+print (fruits)
+fruits.insert(0,'Grapes')
+print (fruits)
+for i in fruits:
+    if i[0][0] == 'P' or i[0][0] == 'p':
```
natasha-aleksandrova: `i[0][0]` is redundant, it only needs to be `i[0]`

students/vvinodh/Lesson3/list_lab.py
```diff
@@ -0,0 +1,58 @@
...
+        print(i)
+
+#Series 2
+print(fruits)
+del fruits[len(fruits)-1]
```
natasha-aleksandrova: this can be simplified to:
`del fruits[-1]`

students/vvinodh/Lesson3/mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+        if choice == '1':
+            thankyou()
+        elif choice == '2':
+            report()
+        elif choice == '3':
+            quit_program = True
+            break
+        else:
+            print("Please try again")
+        menu()
```
natasha-aleksandrova: no need to call `menu()` you are in a while loop, it will naturally loop through code again

students/vvinodh/Lesson3/mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+        print("Press 3 to quit the program\n")
+        choice = input("-> ")
+
+        if choice == '1':
+            thankyou()
+        elif choice == '2':
+            report()
+        elif choice == '3':
+            quit_program = True
+            break
```
natasha-aleksandrova: no need for both `break` AND boolean flag (`quit_program`)

recommended to change to `while True` and keep `break` statement.

students/vvinodh/Lesson3/mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+        menu()
+
+
+def thankyou():
+    global donors
+    print("~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Main Menu ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~")
+    print("- If the user types ‘list’, show them a list of the donor names and re-prompt -")
+    print("- type 'menu' at any time to return to the menu -\n")
+
+    def list():
```
natasha-aleksandrova: don't use `list` for var name or function name, it is overwriting built-in `list` so you won't be able to use that function in this scope

students/vvinodh/Lesson3/sequence.py
```diff
@@ -0,0 +1,18 @@
...
+    return seq[4:-4:2]
+#with the elements reversed (just with slicing).
+def reversed(seq):
+    return seq[::-1]
+#with the middle third, then last third, then the first third in the new order.
+def thirds(seq):
+    # Break the string into first, second and third portiona and then concatenate them
+    one_third = round(len(seq)/3)
+    two_third = round(2*len(seq)/3)
+    return seq[one_third:two_third] + seq[two_third:] + seq[:one_third]
```
natasha-aleksandrova: nice work here, remember to have two lines space between functions.

vvinodh:
Natasha,
When I try to do a pull request for specific assignment I get this message.


Please advice.
Thanks,
Vinodh

natasha-aleksandrova:
@vvinodh 
* navigate to the Pull Request with your submission
* click on the browser bar and copy link, which in this case looks like this: `https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/650`
* submit on the course website in the PR link form

you current submission has `Self_Paced-Online/students/vvinodh/Lesson3/list_lab.py`, please follow above steps next time.

thanks, 
natasha.

PR #651, opened 2018-05-28
Lesson 08 Assignments

TinaBrownfield:
Adding Lesson 08 Assignments.

### DIFF
--- a/students/TinaB/lesson08/circle.py
+++ b/students/TinaB/lesson08/circle.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python3
+"""
+Lesson 08 Assignment - Create a Circle Class
+"""
+from math import pi
+
+
+class Circle:
+    """Class for Circle - STEP ONE"""
+
+    def __init__(self, radius):
+        if radius is not None and radius < 0:
+            raise ValueError('Radius must be more than zero')
+
+        self._radius = radius
+        self._diameter = 2 * radius
+        self._area = pi * radius**2
+        self._perimeter = 2 * self._radius * pi
+
+    def __str__(self):
+        """Adding str method to circle class - STEP SIX"""
+        return f'Circle with radius: {self.radius}'
+
+    def __repr__(self):
+        """Adding repr method to circle class - STEP SIX"""
+        return 'Circle({})'.format(self.radius)
+
+    def __add__(self, other):
+        """Adding add method to circle class - STEP SEVEN"""
+        if isinstance(other, Circle):
+            return Circle(self.radius + other.radius)
+        return Circle(self.radius + other)
+
+    def __sub__(self, other):
+        """Adding add method to circle class - STEP SEVEN"""
+        if isinstance(other, Circle):
+            return Circle(self.radius - other.radius)
+        return Circle(self.radius - other)
+
+    def __mul__(self, other):
+        """Adding add method to circle class - STEP SEVEN"""
+        if isinstance(other, Circle):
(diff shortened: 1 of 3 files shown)


students/TinaB/lesson08/circle.py
```diff
@@ -0,0 +1,123 @@
...
+    """Class for Circle - STEP ONE"""
+
+    def __init__(self, radius):
+        if radius is not None and radius < 0:
+            raise ValueError('Radius must be more than zero')
+
+        self._radius = radius
+        self._diameter = 2 * radius
+        self._area = pi * radius**2
+        self._perimeter = 2 * self._radius * pi
```
robalford: Calculated values (diameter, radius, perimeter) should be handled in the getters and setters. Radius is best defined as as an instance attribute

students/TinaB/lesson08/circle.py
```diff
@@ -0,0 +1,123 @@
...
+    @radius.setter
+    def radius(self, value):
+        if value < 0:
+            raise ValueError
+
+        if self._radius != value:
+            self._diameter = value * 2
+            self._radius = value
+            self._area = pi * self._radius**2
+            self._perimeter = 2 * self._radius * pi
```
robalford: Radius doesn't need the getter and setter, it can be defined as instance attribute

students/TinaB/lesson08/circle.py
```diff
@@ -0,0 +1,123 @@
...
+        if diameter <= 0:
+            raise ValueError('Diameter needs to be a number greater then zero')
+        self = cls(diameter/2)
+        self.diameter = diameter
+        return self
+
+    @property
+    def area(self):
+        """Area property for Circle - STEP FOUR"""
+        return self._area
```
robalford: Move calculation from `__init__` to here

students/TinaB/lesson08/circle.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @property
+    def perimeter(self):
+        """setting perimeter property"""
+        return self._perimeter
+
+    @property
+    def diameter(self):
+        """diameter for Circle - STEP TWO"""
+        return self._diameter
```
robalford: Move calculation here from `__init__`

students/TinaB/lesson08/circle.py
```diff
@@ -0,0 +1,123 @@
...
+    @diameter.setter
+    def diameter(self, value):
+        """Set diameter for Circle - STEP THREE"""
+        if value < 0:
+            raise ValueError
+
+        if self._diameter != value:
+            self._diameter = value
+            self._radius = value / 2
+            self._area = pi * self._radius**2
```
robalford: This only needs to set the radius value, the area logic should be handled in the area `@property`

students/TinaB/lesson08/test_circle.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+class TestCircle(unittest.TestCase):
+
+    def __init__(self, *args, **kwargs):
+        super(TestCircle, self).__init__(*args, **kwargs)
+        self.t_radius = 4
+        self.t_diameter = 2 * self.t_radius
+        self.t_area = pi * self.t_radius**2
+        self.t_perimeter = 2 * self.t_radius * pi
```
robalford: You don't want to inherit from the `TestCase` class to set up test data in this way, instead you can create a circle instance for testing in the `setUp` method rather

PR #652, opened 2018-05-28
Finshed FP

BrandonHenson:
Thank you. Natasha and Rob for this class. I feel like a learned a lot.


natasha-aleksandrova:
@BrandonHenson great work! Glad you enjoyed the class and good luck on the next one!

PR #655, opened 2018-05-29
Please review lesson 07 only

mayc4t:
I partially "git push" lesson 08 in this thread... sorry. Please ignore those 2 files (Circle.py, CircleTest.py).  

Files for lesson 07:
     html_render.py
     html_render_test.py
    and the original file run_html_render.py

### DIFF
--- a/students/mayc4t/lesson07/html_render.py
+++ b/students/mayc4t/lesson07/html_render.py
@@ -11,12 +11,12 @@ class TextWrapper:
     def render(self, file_out, cur_ind='', new_line=True):
         out_str = self.text
         if new_line:
-            out_str = "%s\n" % out_str
-        file_out.write(out_str)
+            out_str = "{}\n".format(out_str)
+        file_out.write("{}{}".format(cur_ind, out_str))
 
 class Element:
     tag = 'html'
-    ind = 4
+    ind = '  '
     content =""
 
 
@@ -29,7 +29,7 @@ class Element:
         
         self.attributes = ''
         for key, value in kwargs.items():
-            self.attributes = '%s %s="%s"' % (self.attributes, key, value)
+            self.attributes = '{} {}="{}"'.format(self.attributes, key, value)
 
 
     def append( self, content ):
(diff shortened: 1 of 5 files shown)


students/mayc4t/lesson07/html_render.py
```diff
@@ -0,0 +1,125 @@
...
+        if new_line :
+            file_out.write("{}<{}{}>\n".format(cur_ind, self.tag, self.attributes))
+        else: 
+            file_out.write("{}<{}{}>".format(cur_ind, self.tag, self.attributes))
+        for idx_elm in self.sub_elements: 
+            idx_elm.render(file_out, '{}{}'.format(cur_ind, self.ind), new_line)
+        file_out.write("{}</{}>\n".format(cur_ind, self.tag))
+
+    def dump(self):
+        print('Element internals: tag={}, sub_elements={}'.format(self.tag, self.sub_elements))
```
natasha-aleksandrova: you could also define `__str__` method but I think you'll learn about that in Lesson 8 😉

students/mayc4t/lesson07/html_render.py
```diff
@@ -0,0 +1,125 @@
...
+
+class Body(Element):
+    tag = "body"
+
+class P(Element):
+    tag = "p"
+
+class Head(Element):
+    tag = "head"
+
```
natasha-aleksandrova: styling tip: should be two lines spaces between classes

students/mayc4t/lesson07/html_render_test.py
```diff
@@ -0,0 +1,422 @@
...
+    try:
+        if (CLEANUP_TEST_FILES):
+            os.remove(file_out)
+    except:
+        pass
+
+class TestTextWrapper(unittest.TestCase):
+    def setUp(self):
+        """setUp creates a file to back the TextWrapper class."""
+        self.file_out = 'text_wrapper_test.txt'
```
natasha-aleksandrova: you can use `StringIO` module which gives you a file like object without creating actual files.

students/mayc4t/lesson08/Circle.py
```diff
@@ -0,0 +1,104 @@
...
+import math
+
+class Circle:
+    def __init__(self, r):
+        self._r = r
+
+    @classmethod
+    def from_diameter(self, d):
+        #return cls(d/2)
+        return self(d/2)
```
natasha-aleksandrova: instead of self it should be `cls`, the name technically doesn't matter but using anything else is an anti-pattern.

students/mayc4t/lesson08/Circle.py
```diff
@@ -0,0 +1,104 @@
...
+            raise ValueError("Radius can't be negative")
+        self._r = val
+    
+    @property
+    def diameter(self):
+        return self.radius * 2
+    
+    @diameter.setter
+    def diameter(self, val):
+        self._d =val
```
natasha-aleksandrova: you're not using `self._d` in your diameter getter, no need for this var

students/mayc4t/lesson08/CircleTest.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python
+# #############################################################################
+# Written By: Mayc4t
+# May 28, 2018
+# Self-paced Python
+# lesson 8 - Properties and Magic Method
+# fileID: CircleTest.py
+# #############################################################################
+
+import Circle
```
natasha-aleksandrova: module names (file names) should be all lower snake_case

students/mayc4t/lesson08/Circle.py
mayc4t: I'm not sure how... but the reason I commented out the cls was I got compile error, and I didn't know to fix that so I changed it to self. I will try it again

PR #656, opened 2018-05-30
adding lesson 09 assignment and test files

TinaBrownfield:
Adding lesson 09 assignment and test files. Started out doing separate files and then combined it all in one script. Refactored the mailroom code to include the two recommended classes, and tried to remove all inputs/user interactions to outside the classes.

### DIFF
--- a/students/TinaB/lesson09/mailroom.py
+++ b/students/TinaB/lesson09/mailroom.py
@@ -1,0 +1,302 @@
+#!/usr/bin/env python3
+
+"""Updated Mailroom Script with classes
+$ chmod +x mailroom.py
+"""
+
+import datetime
+import os
+
+
+class Donor:
+
+    def __init__(self, firstname, lastname, donations=None):
+        self._firstname = firstname
+        self._lastname = lastname
+        if donations is None:
+            self._donations = []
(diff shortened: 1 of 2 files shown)


students/TinaB/lesson09/mailroom.py
```diff
@@ -0,0 +1,302 @@
...
+
+class Donor:
+
+    def __init__(self, firstname, lastname, donations=None):
+        self._firstname = firstname
+        self._lastname = lastname
+        if donations is None:
+            self._donations = []
+        else:
+            self._donations = donations
```
robalford: this can be a one liner:
```
self._donations = donations if donations else []
```

students/TinaB/lesson09/mailroom.py
```diff
@@ -0,0 +1,302 @@
...
+    def firstname(self):
+        return self._firstname
+
+    @property
+    def lastname(self):
+        return self._lastname
+
+    @property
+    def donations(self):
+        return self._donations
```
robalford: You don't need to use `@property`s if you aren't adding any custom logic or processing to the attribute

PR #657, opened 2018-05-30
Lesson 9 exercise: Mailroom object-oriented

itsabhister:


students/itsabhi/Lesson9/oo_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+    return True
+
+
+def sending_letters(d1):
+    for item in d1.donors:
+        fh = open(item.donor_name + ".txt", "w")
+        fh.write("Hello {}, \nJust wanted to drop a note of thanks for your recent donation of ${}.\n"
+                 .format(item.donor_name, item.donations[-1]))  #Includes the last donation amount, which is the last element of list of each dict value
+        fh.writelines(["It will be put to very good use.\n\n", "Sincerely,\n\n", "The Team\n"])  # Example of writelines
+        fh.close()
```
natasha-aleksandrova: prefer to use `with` statement

students/itsabhi/Lesson9/oo_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+    return True
```
natasha-aleksandrova: does this need to return True? it doesn't ever return any other value so it doesn't seem to make sense to have a return in this function

students/itsabhi/Lesson9/oo_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+        # Prints donor name for each Donor object in donors list
+        for d in self.donors:
+            print(d.donor_name)
+
+    def search_donors(self, donor):
+        for d in self.donors:
+            if donor == d.donor_name:
+                return d
+            else:
+                return False
```
natasha-aleksandrova: it's common to return a thing or None for functions like this; returning False makes sense for functions that normally would return a boolean value.

students/itsabhi/Lesson9/oo_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+    switch_func_dict = {'1': sending_thanksv2,
+                        '2': create_report,
+                        '3': sending_letters,
+                        '4': quitting
+                        }
+    while True:
+        option = input("Enter 1 to Send a Thank You!, 2 to Create a Report or 3 to Send TY Letters or 4 to Quit: ")
+        try:
+            if not switch_func_dict.get(option)(d1):
+                break
```
natasha-aleksandrova: it looks like all of the above functions are doing unnecessary return due to this logic.

to simplify, you can either:
* call sys.exit in your quit function
* raise custom exception or StopIterationException and catch here & break

PR #658, opened 2018-05-30
Completed Lesson 07 object assignment. Included test and run_html scr…

letoui:
…ipts for accessibility.

### DIFF
--- a/students/ian_letourneau/Lesson07/html_render.py
+++ b/students/ian_letourneau/Lesson07/html_render.py
@@ -1,0 +1,224 @@
+#!/usr/bin/env python3
+# Ian Letourneau
+# 5/30/2018
+
+from io import StringIO
+
+"""A script to define classes to later be initialized by the
+run_html_render.py script."""
+
+
+class Element:
+    """Parent class for all html elements to be created."""
+
+    def __init__(self, open_tag="<html>",
+                 close_tag="</html>", content=None, **kwargs):
+        """Constructor for Element objects.
+
+        :param open_tag="<html>": set opening tag parameter to html
+        should none be given 
+
+        :param close_tag="</html>": set closing tag parameter to html
+        should none be given 
+
+        :param content=None: set content value to NoneType if no 
+        content associated"""
+
+        self.open_tag = open_tag
+        self.close_tag = close_tag
+        self.content = [content]
+        # Loop through input keywords to incoporate into element
+        # attributes.
+        if kwargs:
+            self.open_tag = open_tag.strip(">")
+            for key in kwargs:
+                self.open_tag += ' {}="{}"'.format(key, kwargs[key])
+            self.open_tag += ">"
+
+    def append(self, content=None):
+        """A function to append objects or strings to higher level
+        html element objects."""
+        if hasattr(content, 'render') and self.content:
+            self.content.append(content)
+        elif hasattr(content, 'render'):
+            self.content = content
+        elif content and self.content:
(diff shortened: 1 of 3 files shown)


students/ian_letourneau/Lesson07/html_render.py
```diff
@@ -0,0 +1,224 @@
...
+
+"""A script to define classes to later be initialized by the
+run_html_render.py script."""
+
+
+class Element:
+    """Parent class for all html elements to be created."""
+
+    def __init__(self, open_tag="<html>",
+                 close_tag="</html>", content=None, **kwargs):
```
natasha-aleksandrova: it is better to define tag as class variable:
```
class Element:
    tag = "html" 

    def __init__(...):
        ....
```
then your subclass become very simple (this is all that would need to be defined):
```
class P(Element):
    tag = "p"
```

the rest of the logic like open / close tags etc should be in the function that is using it (`Element`/`render`)

students/ian_letourneau/Lesson07/html_render.py
```diff
@@ -0,0 +1,224 @@
...
+            self.content = content
+        elif content and self.content:
+            self.content += content
+        elif content:
+            self.content = content
+
+    def render(self, file_out, cur_ind=""):
+        """A function to render page elements into stringIO object, 
+        creating html file with associated elements and formatting."""
+        self.cur_ind = "\n" + cur_ind
```
natasha-aleksandrova: it is not recommended to define attributes outside of `__init__` method, `cur_ind` should be a local variable to `render()` function.

students/ian_letourneau/Lesson07/html_render.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class Html(Element):
+    """Html subclass of Element. Changes tags to 'html' tags."""
+
+    def __init__(self):
+        """Html object constructor."""
+        open_tag = "<html>"
+        close_tag = "</html>"
+        Element.__init__(self, open_tag, close_tag)
```
natasha-aleksandrova: refer to my comment above, this is not needed and could be accomplished by simply defining `tag` class variable in this class.

students/ian_letourneau/Lesson07/html_render.py
```diff
@@ -0,0 +1,224 @@
...
+        SelfClosingTag.__init__(self, tag)
+
+
+class Meta(SelfClosingTag):
+    """Meta subclass of SelfClosingTag. Changes tag to 
+    <meta /> with an added charset value if given for encoding."""
+
+    def __init__(self, charset):
+        """Meta object constructor."""
+        tag = '<meta charset="{}" />'.format(charset)
```
natasha-aleksandrova: another example of where class variable should be used.

students/ian_letourneau/Lesson07/test_html_render.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def test_doc_meta_tags():
+    """A function to test the added meta self-closing tag."""
+    page = hr.Html()
+    head = hr.Head()
+
+    head.append(hr.Meta(charset="UTF-8"))
+
+    assert head.content[1].tag == ('<meta charset="UTF-8" />')
```
natasha-aleksandrova: doesn't look like `render` function is tested for any of the classes?

natasha-aleksandrova:
@letoui sounds good!

PR #660, opened 2018-06-01
finished lesson 7

yixingxu:


students/yixingxu/lesson07/test_html_render.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+# tests for the new tags
+def test_html():
+    e = Html("this is some text")
+    e.append("and this is some more text")
+
+    file_contents = render_result(e).strip()
+
+    assert("this is some text") in file_contents
```
natasha-aleksandrova: no need for parens and needs a space:
`assert "this is some text" in file_contents`

PR #661, opened 2018-06-02
[DennisLee][L10][Functional Mailroom] Ready for submission.

DennisLee-DennisLee:
The tool can be executed by running mailroom_ui.py. There is also a test_mailroom.py file for testing lesson 10-related features, with full simulated input and output for the tool when used on the original donor database.


students/DennisLee/lesson10/mailroom.py
```diff
@@ -0,0 +1,343 @@
+#!/usr/bin/env python3
+
+import os
+from functools import reduce
+
+class Donor():
```
natasha-aleksandrova: note that you don't need parens for class def in python3

DennisLee-DennisLee: Okay, thanks.

Do you know how long we have access to a course after a person completes it? I'd like to be able to go back and reference some of the content, if necessary.

PR #662, opened 2018-06-02
Lesson 9 Assignment Ready For Review

TrkCat:
Thanks!


students/TrkCat/Lesson09/mailroom.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python
+from operator import itemgetter
+
+
+class Donor():
```
natasha-aleksandrova: note that in python3 you don't need to have parens in class def

students/TrkCat/Lesson09/mailroom.py
```diff
@@ -0,0 +1,182 @@
...
+    def __init__(self, first_name, last_name, donation):
+        if isinstance(donation, (list, tuple)):
+            self.donations = list(donation)
+        else:
+            self.donations = [donation]
+        self.first_name = first_name
+        self.last_name = last_name
+        self.full_name = ' '.join((first_name.lower(), last_name.lower()))
+        self.title_name = ' '.join((first_name.title(), last_name.title()))
+        self.last_don = self.donations[-1]
```
natasha-aleksandrova: this would be a good candidate for @property method

students/TrkCat/Lesson09/mailroom.py
```diff
@@ -0,0 +1,182 @@
...
+    def __str__(self):
+        return ('{}, {}'.format(self.last_name.title(),
+                                self.first_name.title()))
+
+    def __repr__(self):
+        return ('Donor({}, {}, {})'.format(self.first_name, self.last_name,
+                                           self.donations))
+
+
+class Donors_List():
```
natasha-aleksandrova: per pep8 class names are UpperCamelCase so no underscores.

PR #664, opened 2018-06-02
HTML Render and Test

UWTestEngr:
HTML assignment

### DIFF
--- a/students/Matt_Hudgins/Lesson07/html_render.py
+++ b/students/Matt_Hudgins/Lesson07/html_render.py
@@ -11,4 +11,114 @@
 '''
 
 
+class Element():
+    """Creating a class for rendering html page"""
+    tag = None
+    indent = '    '
 
+    def __init__(self, content=None, **kwargs):
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
+        self.kwargs = kwargs
+
+
+    def append(self, content):
+        self.content.append(content)
+
+
+    def render(self, file_out, cur_ind=''):
+        file_out.write(cur_ind + '<{}'.format(self.tag))
+        for key, value in self.kwargs.items():
+            file_out.write(' {}="{}"'.format(key, value))
+        file_out.write('>\n')
+        for item in self.content:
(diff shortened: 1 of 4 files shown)


students/Matt_Hudgins/Lesson07/html_render.py
```diff
@@ -0,0 +1,124 @@
...
+'''
+    File Name: html_render.py
+    Author: Matt Hudgins
+    Date created: 5/28/18
+    Date last modified: 5/28/18
+    Python Version 3.6.4
+'''
+
+
+class Element():
```
robalford: You don't need to include the parens for class definition

students/Matt_Hudgins/Lesson07/html_render.py
```diff
@@ -0,0 +1,124 @@
...
+    """Creating a class for rendering html page"""
+    tag = None
+    indent = '    '
+
+    def __init__(self, content=None, **kwargs):
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
```
robalford: This could be a one-liner:
```
```

students/Matt_Hudgins/Lesson07/html_render.py
```diff
@@ -0,0 +1,124 @@
...
+        self.kwargs = kwargs
```
robalford: I would name this something more meaningful like `attributes`

natasha-aleksandrova:
@robalford could you also do the circle one that Matt added here?

students/Matt_Hudgins/Lesson08/circle.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @property
+    def radius(self):
+        return self.radius
+
+    @radius.setter
+    def raduius(self, value):
+        if value < 0:
+            raise ValueError('Must have a value greater than 0')
+        self.radius = value
```
robalford: In this case, I think it'd be simpler to have `radius` be an instance attribute and handle the validation in the `__init__` method

PR #665, opened 2018-06-03
Lesson 09: Done and ready for review

mayc4t:
File submitted: mailroom5.py
                   test_mailroom5.py
   Use autopep -i for all files.

### DIFF
--- a/students/mayc4t/lesson09/mailroom5.py
+++ b/students/mayc4t/lesson09/mailroom5.py
@@ -1,0 +1,239 @@
+#!/usr/bin/env
+# #############################################################################
+# Written By: Mayc4t
+# June 3, 2018
+# Self-paced Python
+# lesson 9 -
+# fileID: mailroom5.py
+# #############################################################################
+
+
+import collections
+from collections import *
+
+
+class Donor():
+    # You’ll want a Donor class –
+    # this will hold all the information about the donor
+    # , and have attributes,
+    #            properties,
+    #            and methods to provide access to the donor specific information that is needed.
+    def __init__(self, first, last, donations=None):
+        self._last = last
+        self._first = first
+        self._donations = donations
+        self._num_gift = len(donations)
+        self._avg = sum(donations)/len(donations)
+
+    @property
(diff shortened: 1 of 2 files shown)


students/mayc4t/lesson09/mailroom5.py
```diff
@@ -0,0 +1,239 @@
...
+        self._num_gift = len(donations)
+        self._avg = sum(donations)/len(donations)
+
+    @property
+    def donations(self):
+        return(self._donations)
+
+    @property
+    def name(self):
+        self._name = "{} {}".format(self._first, self._last)
```
robalford: You shouldn't be assigning to an instance attribute inside the property method like this. If you `return "{} {}".format(self._first, self._last)` without assigning to `self._name`, then `self.name` will be accessible on the object instance using the same dot syntax as an instance attribute, but the logic used to calculate the value will be nicely encapsulated in the `name` method.

students/mayc4t/lesson09/mailroom5.py
```diff
@@ -0,0 +1,239 @@
...
+        else:
+            return(sum(self.donations)/len(self.donations))
+
+    @property
+    def num_gift(self):
+        return len(self.donations)
+
+    @staticmethod
+    def sort_by_name(self):
+        return (self._lname, self._fname)
```
robalford: Static method indicates a method that does not require access to self

students/mayc4t/lesson09/mailroom5.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def add_donation(self, donor):
+        # check if donor is in self._donors_db
+        if any(donor.name in x for x in self.names):
+            idx = self.names.index(donor.name)
+            #print ("Exist")
+            #print (idx)
+            gift = donor.total_donation
+            self._donors_db[idx].add_donation(gift)
+            #print (self._donors_db[idx].__str__())
```
robalford: It's a good idea to delete commented out debugging statements like this before committing

PR #671, opened 2018-06-04
Series code for Assignment 2

rfelts:
Here is the Fibonacci and Lucas series code problem for assignment 2. Based on the assignment description I wrote the code using recursive functions. Which becomes increasing inefficient as higher series values are requested. But as I'm guessing the assignment was designed to teach how recursion works in coding. I did not switch to another method like iteration to see if it was more efficient.

Russell


natasha-aleksandrova:
@rfelts good observation on recursion efficiency here 😉

students/rfelts/lesson02/series.py
```diff
@@ -0,0 +1,114 @@
...
+    Returns:
+    int: The nth value in the fibonacci series
+
+    """
+
+    if n > 1:
+        return fibonacci(n - 1) + fibonacci(n - 2)
+    else:
+        return n
+
```
charlesrobison: Nice and concise!

students/rfelts/lesson02/series.py
```diff
@@ -0,0 +1,114 @@
...
+    int: The nth value in the lucas series
+
+    """
+
+    if n == 0:
+        return 2
+    elif n == 1:
+        return 1
+    else:
+        return lucas(n - 1) + lucas(n - 2)
```
charlesrobison: Might consider adding a catch for negative number inputs.

students/rfelts/lesson02/series.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def sum_series(n, seriesA=0, seriesB=1):
```
natasha-aleksandrova: note that per python style guide, var names (and functions) should follow snake_case convention -> `series_a`

PR #674, opened 2018-06-05
added file for lesson 2: grid printer

mkdir-washington-edu:
mkdir01/lesson02/grid_printer_final.py

### DIFF
--- a/students/mkdir01/lesson02/grid_printer_final.py
+++ b/students/mkdir01/lesson02/grid_printer_final.py
@@ -1,0 +1,27 @@
+def pS(x):  # printSign
+    print(x + " ", end="")
+    return
+
+
+def print_grid2(gDimension = 3, gSize = 4):
+    gSize = gSize + 1  # adds to gSize based on images on assignment page
+    for row in range(0, (gSize) * (gDimension) + 1):  # controls the row you are on
+        if row % (gSize) == 0:  # 0th, gSize, gSize*2, etc
+            for col in range(0, (gSize) * (gDimension) + 1):  # writes the symbols in columns
+                if col % gSize == 0:
+                    pS("+")
+                else:  # all other columns
+                    pS("-")
+                if col == gSize * gDimension:
+                    print()
+        else:  # i.e. all other rows
+            for col in range(0, gSize * gDimension + 1):
+                if col % gSize == 0:
+                    pS("|")
+                else:
+                    pS(" ")
+                if col == gSize * gDimension:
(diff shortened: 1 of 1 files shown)


students/mkdir01/lesson02/grid_printer_final.py
```diff
@@ -0,0 +1,27 @@
+def pS(x):  # printSign
```
natasha-aleksandrova: pep8 style guideline: function names and var names should use snake_case convention -> `def ps(x):`

also `ps` isn't very intuitive, instead of adding a comment next to function why not go ahead and name the function with that?:
```
def print_sign(sign):
```

students/mkdir01/lesson02/grid_printer_final.py
```diff
@@ -0,0 +1,27 @@
...
+    print(x + " ", end="")
+    return
```
natasha-aleksandrova: no need for return here

students/mkdir01/lesson02/grid_printer_final.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def print_grid2(gDimension = 3, gSize = 4):
```
natasha-aleksandrova: same comment about using snake_case convention

students/mkdir01/lesson02/grid_printer_final.py
```diff
@@ -0,0 +1,27 @@
...
+    gSize = gSize + 1  # adds to gSize based on images on assignment page
+    for row in range(0, (gSize) * (gDimension) + 1):  # controls the row you are on
+        if row % (gSize) == 0:  # 0th, gSize, gSize*2, etc
+            for col in range(0, (gSize) * (gDimension) + 1):  # writes the symbols in columns
```
natasha-aleksandrova: no need for parens around `gSize` and `gDimension` vars

students/mkdir01/lesson02/grid_printer_final.py
```diff
@@ -0,0 +1,27 @@
...
+                if col % gSize == 0:
+                    pS("+")
+                else:  # all other columns
+                    pS("-")
+                if col == gSize * gDimension:
+                    print()
+        else:  # i.e. all other rows
+            for col in range(0, gSize * gDimension + 1):
```
natasha-aleksandrova: not necessary to pass in zero as first param to `range` as that is the default value

PR #677, opened 2018-06-05
Completed Lesson08 assignment. Enjoyed this one a lot."

letoui:


students/ian_letourneau/Lesson08/circle_lab.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @classmethod
+    def from_diameter(cls, value):
+        """Add an alternate constructor to create Circle
+        objects from a given diameter. recalculates radius
+        using half the given value."""
+        self = cls(value)
+        self._diameter = value
+        self._radius = value/2
+        return self
```
natasha-aleksandrova: this method should simply be 
```
@classmethod
def from_diameter(cls, value):
    return cls(value / 2)
```
where you create instance of circle, since circle expects a radius - you will convert diameter to radius.

students/ian_letourneau/Lesson08/test_circle_lab.py
```diff
@@ -0,0 +1,114 @@
...
+
+# Test Step One
+
+
+def test_circle_creation():
+    c = cr.Circle(5)
+    assert c.radius == (5)
+
+    new_c = cr.Circle(23)
+    assert new_c.radius == (23)
```
natasha-aleksandrova: you can really do these inline:
`assert cr.Circle(23).radius == 23`

students/ian_letourneau/Lesson08/test_circle_lab.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env python3
+# Ian Letourneau
+# 6/4/2018
+
+import math
+import circle_lab as cr
```
natasha-aleksandrova: note that aliases should be generally avoided as they reduce readability 

you can instead do:
`from circle_lab import Circle`

students/ian_letourneau/Lesson08/test_circle_lab.py
```diff
@@ -0,0 +1,114 @@
...
+# Test Step Eight
+
+
+def test_comparison_overrides():
+    c1 = cr.Circle(2)
+    c2 = cr.Circle(4)
+
+    assert (c1 > c2) == False
+    assert (c1 < c2) == True
+    assert (c1 == c2) == False
```
natasha-aleksandrova: for true checks, you don't to test for True, that's what `assert` provides:
`assert c1 < c2` is sufficient

for False, you can either:
`assert not c1 > c2` or `assert c1 > c2 is False`

Using `==` with True/False isn't pythonic.

PR #678, opened 2018-06-05
Finished OO Mailroom

UWTestEngr:


students/Matt_Hudgins/Lesson09/mailroom_5.py
```diff
@@ -0,0 +1,144 @@
...
+class Donor:
+
+    def __init__(self, name, donation=None):
+        self.donation = donation
+        self.name = name
+        self.total = sum(self.donation)
+        if donation is None:
+            self.donation = []
+        else:
+            self.donation = donation
```
natasha-aleksandrova: shortcut tip: `self.donations = donations or []`

note I used `donations` var name since we should be transparent about singular vs plural to clarify to the reader what data this var is expected to hold.

students/Matt_Hudgins/Lesson09/mailroom_5.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @property
+    def add_donation(self, amount):
```
natasha-aleksandrova: this should be a function not property

students/Matt_Hudgins/Lesson09/mailroom_5.py
```diff
@@ -0,0 +1,144 @@
...
+        self.donation.append(amount)
+        self.total += amount
+        self.average = self.total / len(self.donation)
```
natasha-aleksandrova: `self.average` is being defined here for the first time; all instance variables should be defined in `__init__` method.

students/Matt_Hudgins/Lesson09/mailroom_5.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def __init__(self, name=None):
+        self.donors = [] if name is None else name
+
+
+    def add_donor(self, name):
+        self.donors.append(name)
+
+
+    def get_donor(self, name):
```
natasha-aleksandrova: some notes about styling: 
* you should have 1 line space between class methods (module functions need 2 lines space)
* you also need 2 lines space between classes.

students/Matt_Hudgins/Lesson09/mailroom_5.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+    def donor_input(self):
+        return input("Enter a donor name or input 'List' for a list of donors\n>")
+
+
+    def send_thankyou(self):
+        don_input = None
+        while not don_input:
+            don_input = self.donor_input()
```
natasha-aleksandrova: keep in mind the separation of concern here; you should have no input functionality in your Data class.

if you were to "upgrade" your mailroom to have UI instead of CLI, there should be no changes to `Donor` or `DonorDatabase` classes.

as it is now you will have to make some changes.

PR #679, opened 2018-06-05
Slicing, list, and string formatting labs for lesson 03

smitco:
The first 3 exercises for lesson 03- slicing, list, string formatting

### DIFF
--- a/students/smitco/lesson03/list_lab.py
+++ b/students/smitco/lesson03/list_lab.py
@@ -1,0 +1,102 @@
+#lesson 03 lists lab
+
+#!/usr/bin/env python3
+
+
+fruit= ["apples", "pears", "oranges", "peaches"]
+
+
+#series 1
+fruit1 = fruit[:]
+print(fruit1)
+
+response = input("Add a new fruit to our list: ")
+fruit1.append(response)
+print(fruit1)
+
+response = input("Pick a number between 1 and 5: ")
+print("You picked " + fruit1[int(response)-1])
+
+response = input("Add another fruit: ")
+fruit1 = [response] + fruit1
+print(fruit1)
+
+response = input("One more: ")
+fruit1.insert(0, response)
+print(fruit1)
+
+def startcheck(basket):
+    withp = []
+    for f in basket:
+        if f.startswith("p") or f.startswith("P"):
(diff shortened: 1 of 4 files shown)


students/smitco/lesson03/list_lab.py
```diff
@@ -0,0 +1,102 @@
...
+fruit2.remove(response)
+print(fruit2)
+
+print("Our fruit basket multipled!")
+doublefruit2= fruit2*2
+print(doublefruit2)
+
+response = input("What else should go? ")
+for f in (doublefruit2):
+    if f == response:
```
robalford: This can be shortened to `while response in doublefruit2:`

students/smitco/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,75 @@
...
+print()
+print()
+
+#task 3- dynamically rewrite "the 3 numbers are: {:d}, {:d}, {:d}".format(1,2,3) 
+testtup = (1, 2, 3, 4, 5, 6)
+testtup2 = (4, 7, 5)
+testtup3 = (6, 19, 22, 34)
+def strform(tup):
+    l = len(tup)
+    lst = ""
```
robalford: It's a good idea to always use meaningful variable names, and in the case of `l`, you probably don't need to assign that to a variable at all (just call `len` where you need it)

students/smitco/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,75 @@
...
+print(fruitinfo20)
+print()
+print()
+
+#task 6- use string formatting to print data (name, age, cost) in columns
+namedata = ["Trixie", "Lassie", "Polly", "Buddy", "Remy"]
+agedata = [4, 2, 7, 4, 1]
+costdata = [400, 1000, 200, 500, 1200]
+assert len(namedata) == len(agedata)
+assert len(agedata) == len(costdata)
```
robalford: `assert` statements are generally used for tests and not in the main flow of your program, since they will cause the program to crash if they fail

robalford:
@smitco remember to submit the assignments on the course website as well so we can grade. Thanks!

PR #684, opened 2018-06-09
Lesson 03 Mailroom

smitco:
New pull request after re-forking. Deleted draft file and added final copy of the mailroom exercise for approval. All looks fine on my end.


students/smitco/lesson03/mailroom.py
```diff
@@ -44,25 +45,43 @@ def intro():
...
-        
-if __name__ == '__main__':        
+def report():
+    print()
+    print("{:<20} {:<15} {:^12} {:<15}".format("Donor", "Total Given", "Number", "Average Given"))
+    for i in range(len(donors)):
+        tot = sum(amounts[i])
+        num = len(amounts[i])
+        print("{:<20} ${:^15.2f} {:^12} ${:^15.2f}".format(donors[i], tot, num, tot/num))
+    print()
```
natasha-aleksandrova: this report is supposed to be sorted. please review instructions and add the functionality for next time.

students/smitco/lesson03/mailroom.py
```diff
@@ -44,25 +45,43 @@ def intro():
...
-    dollars = input("What is the donation amount? ")
-    if int(dollars) < (10 ** 6):
-        position = donors.index(n)
-        a = amounts[position]
-        a = a + (int(dollars),)
-        amounts[position] = a
-        return dollars
+    if n == "quit":
+        print()
+        pass
```
natasha-aleksandrova: the pass statement here and line 56 have no effect; they should be removed.

students/smitco/lesson03/mailroom.py
```diff
@@ -20,7 +20,8 @@ def menu():
             card(name, value)
             flag = True
         elif choice == "2":
-            flag = False
+            report()
+            flag = True
         else:
             flag = False
```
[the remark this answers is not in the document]
natasha-aleksandrova: the `intro` function has some duplication, can you spot it and refactor?

students/smitco/lesson03/mailroom.py
```diff
@@ -2,12 +2,12 @@
 #!/usr/bin/env python3
 
 #starting list of 5+ donors and 1-3 donations each
-donors = ["John", "Jane", "Judith", "Joseph", "Jenny"]
+donors = ["John Travolta", "Jane Fonda", "Judy Blume", "Joey Tribbiani", "Jenny Gump"]
 amounts = [(5000, 7500), (10000, 8000, 6500), (3000, 3000), (9000,), (10300, 13750, 12500)]
```
natasha-aleksandrova: why use immutable data structure (tuple) if you do need to modify it?

students/smitco/lesson03/mailroom.py
```diff
@@ -44,25 +45,43 @@ def intro():
...
+        elif int(dollars) < (10 ** 6):
+            position = donors.index(n)
+            a = amounts[position]
+            a = a + (int(dollars),)
+            amounts[position] = a
+            return dollars
+        else:
+            print("\nNice try, Moneybags.\n")
+            quit()
 def card(n, v):
```
natasha-aleksandrova: avoid using `n` and `v` type of vars since the reader won't know what those mean.

students/smitco/lesson03/mailroom.py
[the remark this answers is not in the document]
smitco: One of the tasks I struggled with for this exercise was to have the program return to the menu when a choice is completed. What I think I'm gathering now is that it would be better to have a simple menu function and then put the code to loop back to the menu in the name == main block instead of the menu function. Is that right?

For the intro function, I struggled to have it loop back to the original prompt after the user asks to see a list, so I did it the undesirable way of just copying the code block. Should I have used a while loop to get back to the input for the donor name?

students/smitco/lesson03/mailroom.py
smitco: That makes sense looking back at it. I had pass in there originally, but I added the extra print line for aesthetics. I didn't think about not needing pass once I had the print statement in there. Thanks!

students/smitco/lesson03/mailroom.py
smitco: Thanks! I'm trying to break this bad habit.

PR #687, opened 2018-06-10
Creating pull request for Trigrams and Exceptions assignments

willhwhite:
Accidentally included the Mailroom assignment. Nothing has changed from what is included here and what was included in the previous pull request for the Mailroom file.

### DIFF
--- a/students/WWhite/lesson04/trigrams.py
+++ b/students/WWhite/lesson04/trigrams.py
@@ -1,0 +1,48 @@
+import string, random
+
+file = open('sherlock_small.txt', 'r')
+file_text = file.read().lower()
+file.close()
+
+translator = str.maketrans(string.punctuation, ' '*len(string.punctuation))
+
+new_file_text = file_text.translate(translator).split()
+
+
+def trigrams(list):
+    trigram_list = []
+    for i in range(len(list)-2):
+        trigram_list.append(('{} {}'.format(list[i], list[i+1]), list[i+2]))
+    return trigram_list
+
+
+def convert_to_dict(list):
+    trigram_dict = dict()
+    for i in list:
+        if i[0] in trigram_dict:
+            trigram_dict[i[0]].append(i[1])
+        else:
+            trigram_dict[i[0]] = [i[1]]
+
+    return trigram_dict
+
+
+new_dict = convert_to_dict(trigrams(new_file_text))
+new_book = list()
+random_key = random.choice(list(new_dict))
+random_key_split = random_key.split()
+new_book.append(random_key_split[0])
+new_book.append(random_key_split[1])
+
+count = 0
+while random_key in new_dict and count < 100:
+    random_key_split = random_key.split()
+    random_value = random.choice(new_dict[random_key])
+    new_book.append(random_value)
+
+    random_key = random_key_split[1] + " " + random_value
+
+    count += 1
+
+print(' '.join(new_book[0:]))
+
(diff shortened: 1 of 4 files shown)


students/WWhite/lesson04/trigrams.py
```diff
@@ -0,0 +1,48 @@
+import string, random
```
robalford: These `import` statements should go on separate lines

students/WWhite/lesson04/trigrams.py
```diff
@@ -0,0 +1,48 @@
...
+
+file = open('sherlock_small.txt', 'r')
```
robalford: It's a best practice to use a `with` statement when working with files which will automatically close the file when you exit the `with` block

students/WWhite/lesson04/trigrams.py
```diff
@@ -0,0 +1,48 @@
...
+file_text = file.read().lower()
+file.close()
+
+translator = str.maketrans(string.punctuation, ' '*len(string.punctuation))
+
+new_file_text = file_text.translate(translator).split()
+
+
+def trigrams(list):
```
robalford: This variable name (`list`) is overwriting the Python built-in, I would suggest naming it something different and specific to your program

students/WWhite/lesson04/trigrams.py
```diff
@@ -0,0 +1,48 @@
...
+    trigram_list = []
+    for i in range(len(list)-2):
+        trigram_list.append(('{} {}'.format(list[i], list[i+1]), list[i+2]))
+    return trigram_list
```
robalford: Seems like you could build the dict directly while looping over the text, rather than use an intermediary list

students/WWhite/lesson04/trigrams.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def convert_to_dict(list):
+    trigram_dict = dict()
+    for i in list:
+        if i[0] in trigram_dict:
+            trigram_dict[i[0]].append(i[1])
+        else:
+            trigram_dict[i[0]] = [i[1]]
```
robalford: Check out `dict.setdefault` or `collections.defaultdict` to simplify this

students/WWhite/lesson05/Mailroom.py
```diff
@@ -0,0 +1,121 @@
...
+    return str_letter
+
+
+def add_donation(donor_name):  # Function for the user to add a donation
+    while True:
+        try:
+            donation_amount = input("Please input the donation amount: ")
+            donation_amount = float(donation_amount)
+            break
+        except:
```
robalford: It's a good idea to identify and catch the expected exception in your try/except statements, so would recommend using `except ValueError:`

students/WWhite/lesson05/Mailroom.py
```diff
@@ -0,0 +1,121 @@
...
+            name = switch_func_dict.get(1)()
+            switch_func_dict.get(2)(name)
+
+
+        elif str_input == '2':  # If the user enters 2, create a report and then return to the original report
+            switch_func_dict.get(3)()
+
+        elif str_input == '3':  # If the user enters 3, create letters for everyone, saved in individual text files
+            switch_func_dict.get(4)()
+
```
robalford: If you map user selections to keys in the dict, you can use a try/except and catch a `KeyError` here for bad user input

PR #688, opened 2018-06-10
Lesson07 scripts ready for review

seaneightysix:


students/Sean_Tasaki/Lesson07/html_render.py
```diff
@@ -0,0 +1,129 @@
...
+Sean Tasaki
+6/10/2018
+Lesson07.html_render
+"""
+
+# Step 1
+class Element:
+    tag_name = ''
+    indent = '    '
+    test = 'test'
```
robalford: You shouldn't need this `test` attribute

students/Sean_Tasaki/Lesson07/html_render.py
```diff
@@ -0,0 +1,129 @@
...
+    def __init__(self, content=None, **style):
+        self.style = style
+        if content:
+            self.content = [content]
+        else:
+            self.content = []
```
robalford: This could be a one-liner `self.content = [content] if content else []`

students/Sean_Tasaki/Lesson07/html_render.py
```diff
@@ -0,0 +1,129 @@
...
+                item.render(fileout)
+            else: 
+                fileout.write(item) 
+        fileout.write("</{}>\n".format(self.tag_name))
+
+class Title(OneLineTag):
+    tag_name = 'title'
+
+class SelfClosingTag(Element):
+    tag_name = ''
```
robalford: Don't need to override `tag_name` here

PR #691, opened 2018-06-11
finished lesson08

yixingxu:


students/yixingxu/lesson08/circle.py
```diff
@@ -0,0 +1,39 @@
+from math import pi
+
+class Circle():
```
natasha-aleksandrova: you don't need the parens here

students/yixingxu/lesson08/circle.py
```diff
@@ -0,0 +1,39 @@
+from math import pi
```
natasha-aleksandrova: the `__init__.py` file would be on the same level as `circle.py` in the directory

students/yixingxu/lesson08/test_circle.py
```diff
@@ -0,0 +1,53 @@
+from circle import Circle
+
+def test_init():
+    c = Circle(4)
+    assert (c.radius == 4) is True
```
natasha-aleksandrova: you don't need `is True`, it is redundant; it's basically saying <something> is True is True

`assert c.radius == 4` is sufficient

students/yixingxu/lesson08/test_circle.py
```diff
@@ -0,0 +1,53 @@
...
+
+def test_diameter():
+    c = Circle(4)
+    assert (c.diameter == 8) is True
+    assert (c.radius == 4) is True
+    c.diameter = 2
+    assert (c.diameter == 2) is True
+    assert (c.radius == 1) is True
+
+def test_area():
```
natasha-aleksandrova: module functions should have two lines space between them

students/yixingxu/lesson08/test_circle.py
```diff
@@ -0,0 +1,53 @@
...
+def test_mul_():
+    c2 = Circle(4)
+    assert (c2*3 == Circle(12)) is True
+
+def test_ls_():
+    c1 = Circle(2)
+    c2 = Circle(4)
+    assert ((c1 > c2) == False) is True
+    assert ((c1 < c2) == True) is True
+    assert ((c1 == c2) == False) is True
```
natasha-aleksandrova: you should use `is` with False/True and `is True` is again redundant:

`assert not c1 == c2` or `assert (c1 == c2) is False`

PR #692, opened 2018-06-12
Added exception handling to Mailroom (V3)

willhwhite:


students/WWhite/lesson05/Mailroom.py
```diff
@@ -1,29 +1,46 @@
...
+        try:
+            return_number = input(prompt_text)
+            return_number = float(return_number)
+
+            if min_limit <= return_number <= max_limit:
+                break
+            else:
+                print("Number out of range.")
+        except:
+            print("Invalid input.")
```
charlesrobison: Works well - nice!

students/WWhite/lesson05/Mailroom.py
```diff
@@ -1,29 +1,46 @@
...
+    while True:
+        try:
+            return_number = input(prompt_text)
+            return_number = float(return_number)
+
+            if min_limit <= return_number <= max_limit:
+                break
+            else:
+                print("Number out of range.")
+        except:
```
natasha-aleksandrova: you should catch specific exception, in this case are you catching `float` conversion? if so your try/except should be around float and catch ValueError

students/WWhite/lesson05/Mailroom.py
```diff
@@ -102,16 +113,15 @@ def send_letters():
...
             name = switch_func_dict.get(1)()
             switch_func_dict.get(2)(name)
 
-
-        elif str_input == '2':  # If the user enters 2, create a report and then return to the original report
+        elif option_selected == 2:  # If the user enters 2, create a report and then return to the original report
             switch_func_dict.get(3)()
 
-        elif str_input == '3':  # If the user enters 3, create letters for everyone, saved in individual text files
+        elif option_selected == 3:  # If the user enters 3, create letters for everyone, saved in individual text files
```
natasha-aleksandrova: this block should be simplified a bit, the idea of using a "switch dict" is to actually seemlessly have your program switch to various flows, for example:
```
switch_dict = {"1": func1, "2": func2}
user_in = input("")
try:
    switch_dict[user_in]()
except KeyError:
    print("invalid option")
```

PR #694, opened 2018-06-12
Lesson 4, Markov/ Trigram

mcbraunie:
Trigram program written, text file included for example

### DIFF
--- a/students/MicahBraun/Lesson 04/Markov_Analysis.py
+++ b/students/MicahBraun/Lesson 04/Markov_Analysis.py
@@ -1,0 +1,100 @@
+# ----------------------------------------------------------------------------------------------------------------------
+# NAME: MICAH BRAUN
+# PROJECT: Markov_Analysis.py
+# PURPOSE: Read in string data from text file and perform data manipulations on data.
+# DATE: 06/11/2018
+# DESCRIPTION: Uses text from large body of text (I used The Brothers Karamazov) and selects a designated number of
+# text to accept as for analysis (300 in this case). Creates a dictionary of each word in the text and its associated
+# words that could follow it (based on the text available) and stores the values in the trigram format. The text is
+# then strung together "randomly" based on this predictive analysis and then printed out for the user to view.
+# ----------------------------------------------------------------------------------------------------------------------
+import re
+import random
+# --------------------------------------------------    DATA    --------------------------------------------------------
+potential_words_map = {}                           # map from root_words to a list of potential_words
+root_word = ()                                     # current tuple of words
+big_list = []                                      # used for final data display
+
+#   --------------------------------------------     FUNCTIONS     -----------------------------------------------------
+
+
+def process_f(filename, order=3):
+    """Reads-in file contents (from filename) with default value of 3 (trigram)"""
+
+    fp = open(filename)                             # open file
+
+    for line in fp:                                 # for each line of text in file: do something
+        for word in line.rstrip().split():          # for each word in each line: strip \r and split into tuple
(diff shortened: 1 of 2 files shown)


students/MicahBraun/Lesson 04/Markov_Analysis.py
```diff
@@ -0,0 +1,100 @@
...
+def process_word(word, order=3):
+    """Processes each tuple/word (str, int)
+    "During the first few iterations, all we do is store up the words;   <--- This quote is from Think Python's Example.
+    after that we start adding entries to the dictionary."               <--- I liked their description better than
+                                                                         <--- anything I could come up with.
+    """
+    global root_word                                       # access global tuple word collection
+    if len(root_word) < order:                             # if the length of the tuple is less than order(3), add-
+        root_word += (word,)                               # -word to root_word tuple. it needs to be at least 3 to work
+        return
```
robalford: The keys in the dict should be two word pairs, rather than three words per the assignment instructions

students/MicahBraun/Lesson 04/Markov_Analysis.py
```diff
@@ -0,0 +1,100 @@
...
+
+    try:
+        potential_words_map[root_word].append(word)        # add value of root_word to dict potential_words_map{}
+    except KeyError:
+        potential_words_map[root_word] = [word]            # if key of root_word does not already exist in dict, create
```
robalford: You can avoid this try/except with `dict.setdefault` or `collections.defaultdict`

students/MicahBraun/Lesson 04/Markov_Analysis.py
```diff
@@ -0,0 +1,100 @@
...
+
+    for i in range(n):                                              # for items in range 0-300
+        potential_words = potential_words_map.get(start, None)      # potential_words = random
+        if potential_words == None:                                 # if == None, randomize(300 - 300)
+            randomize(n-i)                                          # randomize all
+            return                                              # Not None
+        word = random.choice(potential_words)                       # word = random selection from potential_words
+        big_list.append(word + " ")                                 # append word to big_list along with whitespace
+        start = arrange_trigram(start, word)                       # start = call arrange_trigram() on start, word
+
```
robalford: I don't think this is accomplishing quite what the assingment specifies. The algorithm should begin with a random word pair, append that pair to the new text (it's good to use a list at this point as you're doing), then choose a random following word from the dict to append to the new text. The next iteration should take the last two word pair from the new text and use them to look up another random following word to append and so on

students/MicahBraun/Lesson 04/Markov_Analysis.py
```diff
@@ -0,0 +1,100 @@
...
+    regex = re.compile('(.*?(\s))*')            # regex = re.compile((any character != \n\t\r) (repeat <-- 0:n times
+                                                # (repeat 0-1 times for resulting -->)(any whitespace)(repeat 0:n)
+    while whole != "":                                              # while whole isn't empty string:
+        break_pos = regex.match(whole[:100]).end()                  # break_pos breaks after 100 words on whitespace
+        print(whole[:break_pos])                                    # print whole up to break_pos
+        whole = whole[break_pos:]                                   # whole = whole starting from previous break_pos
+
+
+def main(filename, n=300, order=3):                                 # main function calls all other functions
+    n = int(n)                                                      # n = default length of text to read-in from file
```
robalford: The typical pattern in Python would be to specify type of the input parameter in a docstring

PR #696, opened 2018-06-13
finished lesson09

yixingxu:


students/yixingxu/lesson09/mailroom5.py
```diff
@@ -0,0 +1,162 @@
+import sys
+import os
+
+class Donor():
```
robalford: Don't need the parens for a class definition

students/yixingxu/lesson09/mailroom5.py
```diff
@@ -0,0 +1,162 @@
...
+        self._name = name
+        self._donations = []
+
+    @property
+    def name(self):
+        return self._name
+
+    @name.setter
+    def name(self, donor_name):
+        self._name = donor_name
```
robalford: You don't need to use a getter and setter if you aren't adding any logic or processing beyond setting the attribute

students/yixingxu/lesson09/mailroom5.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @property
+    def donations(self):
+        return self._donations
+
+    def add_donation(self, donor_donation):
+        self.donations.append(donor_donation)
+
+
+class Donors_Collection():
```
robalford: Use CamelCase (no underscore) for class definition per Pep8

students/yixingxu/lesson09/mailroom5.py
```diff
@@ -0,0 +1,162 @@
...
+        for donor in self._donors:
+            name_list.append(donor.name)
+        return name_list
+
+
+    def add_donation_amount(self, name, donation_amount):
+        for donor in self._donors:
+            if donor.name == name:
+                donor.donations.append(donation_amount)
+                break
```
robalford: Seems like this should be a method on the Donor class

students/yixingxu/lesson09/mailroom5.py
```diff
@@ -0,0 +1,162 @@
...
+        seperator_width = 2
+        title = ['Donor Name', '|', 'Total Given', '|', 'Num Gifts', '|', 'Average Gift']
+        divider = '-' * (
+                    donor_name_width + total_given_width + num_gifts_width + average_gift_width + seperator_width * 5)
+        print('{:<20}{:<2}{:<15}{:<2}{:<5}{:<2}{:<15}'.format(*title))
+        print(divider)
+        for item in a_list:
+            print('{:<20}{:2}{:15.2f}  {:8} {:>2}{:12.2f}\n'.format(*item))
+
+    def send_a_thank_you(self):
```
robalford: It's better to separate the code that deals with user input from the code that handles the data so the input processing would work better inside a module level function

PR #698, opened 2018-06-13
added test_mailroom.py

rgpag:
added test_mailroom.py for L06 review.

### DIFF
--- a/students/rgpag/lesson06/test_mailroom.py
+++ b/students/rgpag/lesson06/test_mailroom.py
@@ -1,0 +1,63 @@
+import datetime
+import os
+from mailroom import (don_in, donors, new_donor, update_don, menu_sel_1,
+                      menu_sel_2, menu_sel_3, menu_switch_dict)
+
+
+def test_don_in():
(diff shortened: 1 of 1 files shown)


students/rgpag/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_menu_sel_1():
+    # check new donor
+    assert menu_sel_1('Robbie', 100) == ('Robbie', 1, 100)
+    # check returning donor
+    assert menu_sel_1('Robbie', 100) == ('Robbie', 2, 200)
+    assert menu_sel_1('menu') is False
+    x = 'Jeff Bezos\nMark Zuckerberg\nBill Gates\nPaul Allen\nRobbie'
+# not sure if this approach is what I want to do here...
+    assert menu_sel_1('list') == print(x)
```
natasha-aleksandrova: same here, this also has dependency on `input` function which will wait for user input (not something that can be part of unit tests)

PR #699, opened 2018-06-14
Lesson 2 exercises.

Parzival-X:


students/rob_sanchez/Lesson 2/Fizz_Buzz.py
```diff
@@ -0,0 +1,16 @@
+def FizzBuzz():
```
robalford: The convention in Python is to use `snake_case` for functions and variables and `CamelCase` for class names

students/rob_sanchez/Lesson 2/Fizz_Buzz.py
```diff
@@ -0,0 +1,16 @@
...
+    #variables
+    num1=3
+    num2=5
+    range_start=1
+    range_end=101
```
robalford: Please run your code through a [Pep8 linter](http://pep8online.com), or configure your text editor to catch pep8 violations (there are instructions for how to do this for SublimeText in the course materials). Pep8 is the Python style guide and includes many conventions for making your code more readable.

students/rob_sanchez/Lesson 2/Fizz_Buzz.py
```diff
@@ -0,0 +1,16 @@
...
+
+    for i in xrange(range_start,range_end):
```
robalford: This should use the built-in Python3 `range` function rather than the Python2`xrange` function. Please make sure that you are using Python 3 for your class work, rather than Python 2 which comes pre-installed on many personal computers. There are instructions for installing the latest version of Python in the course materials [here](https://uwpce-pythoncert.github.io/PythonCertDevel/supplemental/installing/index.html#installing-python).

students/rob_sanchez/Lesson 2/Fizz_Buzz.py
```diff
@@ -0,0 +1,16 @@
...
+        if i%num1==0 and i%num2==0:
+            print('FizzBuzz')
+        elif i%num1==0:
+            print('Fizz')
+        elif i%num2==0:
```
robalford: Include white space around operators for Pep8

students/rob_sanchez/Lesson 2/Grid_Printer.py
```diff
@@ -0,0 +1,41 @@
+#Grid Printer exercise
+def print_grid(n):
+    #variables
+    grid_size=2
+
+    if n%2 == 0:
+        row2_size=n+1
+    else: row2_size=n
+
+    size=n/2
```
robalford: You need to use floor division, convert to int or figure out some other way to handle remainders here because Python won't let you multiply a string times a float the `row` operations below

students/rob_sanchez/Lesson 2/Grid_Printer.py
```diff
@@ -0,0 +1,41 @@
...
+
+    plus='+ '
+    minus='- '
+    bar='|'
+    z=' '
```
robalford: It's a best practice to use meaningful variable names

students/rob_sanchez/Lesson 2/Grid_Printer.py
```diff
@@ -0,0 +1,41 @@
...
+
+    row=plus + minus*(size) + plus + minus*(size) + plus
+    row2=bar + z*(row2_size) + bar + z*(row2_size) + bar
```
robalford: The parens around `size` and `row2_size` are not necessary here

students/rob_sanchez/Lesson 2/Series.py
```diff
@@ -0,0 +1,99 @@
...
+        return var2
+    else:
+        #Surrounding with a try blcok to catch and return
+        #any exceptions raised during execution
+        try:
+            #use recursion to return the sum of the previews two numbers
+                return Fibonacci(n-2) + Fibonacci(n-1)
+        except Exception as e:
+            print(e.message)
+            raise
```
robalford: We won't be introducing exceptions for several lessons so they're probably not necessary to use here, but in general you want to hone in on a particular exception that you think might be raised (i.e. ValueError).

students/rob_sanchez/Lesson 3/Slicing.py
```diff
@@ -0,0 +1,34 @@
...
+    new_seq = remove_every_other(last_four)
+    return new_seq
+
+#Reverses all elements in the sequence
+def reverse_sequence(seq):
+    new_seq = seq[::-1]
+    return new_seq
+#Divides the sequence into thirds and returns the middle, last, and first third.
+def mid_last_first(seq):
+    l=len(seq)/3
```
robalford: In Python3 you would use floor division operator here: `//`

PR #700, opened 2018-06-14
added lesson 3 - slicing lab

mkdir-washington-edu:
mkdir01/lesson03/slicing.py

### DIFF
--- a/students/mkdir01/lesson03/slicing.py
+++ b/students/mkdir01/lesson03/slicing.py
@@ -1,0 +1,34 @@
+
+
+def exchange_first_last(x):
+    return x[-1:] + x[1:len(x)-1] + x[:1]
+
+def every_other(x):
+    return x[::2]
+
+def mid_every_other(x):
+    return x[4:-4:2]
+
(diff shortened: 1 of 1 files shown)


students/mkdir01/lesson03/slicing.py
```diff
@@ -0,0 +1,34 @@
+
+
+def exchange_first_last(x):
+    return x[-1:] + x[1:len(x)-1] + x[:1]
```
natasha-aleksandrova: note that you can just do `x[1:-1]`, no need for `len(x) - 1`

students/mkdir01/lesson03/slicing.py
```diff
@@ -0,0 +1,34 @@
...
+
+def every_other(x):
+    return x[::2]
+
+def mid_every_other(x):
+    return x[4:-4:2]
+
+def reversed(x):
```
natasha-aleksandrova: be careful about overwriting built-ins, `reversed` is built-in function

PR #702, opened 2018-06-17
comment

vvinodh:


students/vvinodh/Lesson5/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+    print("""\n------------------ THANK YOU ------------------\n
+type 'list' to to see a complete list of donors -
+type 'menu' at any time to return to the menu""")
+
+    # asks for a name from the user, or for the 'list' prompt
+    name_check = input("Enter first and last name of a donor\n\n-> ").lower()
+
+    try:
+        options.get(name_check)()
+    except TypeError:
```
robalford: This outer try/except does not seem necessary as you would not generally place the expected flow of the program inside of an `except` block

students/vvinodh/Lesson5/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+            thankyou()
+        else:
+            name_check = name_check.title()
+            try:
+                donors[name_check].append(donation)
+            except KeyError:
+                donors.setdefault((name_check), []).append(donation)
+            finally:
+                letter_dict = {"name": name_check, "don": donation}
+                print("""\n------------------ Letter ------------------\n
```
robalford: Remember that the `finally` code block will run regardless of whether an exception is raised

PR #712, opened 2018-06-22
2nd mailroom assignment, incorporated feedback from 1st mailroom

mkdir-washington-edu:
mailroom2.py

### DIFF
--- a/students/mkdir01/lesson04/mailroom2.py
+++ b/students/mkdir01/lesson04/mailroom2.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+import sys
+
+# changed from list of lists to a dict with values as list
+data = {'John Doe': [120.00, 353.33, 400.00], 'Jane Doe': [3500.04, 2624.44], 'John Smith': [33.33, 400.00, 29.20], 'Jane Smith': [2.00], 'Billy Jo Jones': [100.00, 100.00, 100.00]}
+
+
+# mode 1 - "send a thank you"
+def thank_you():
+    while True:
+        name = input("Please enter a full name (or type \'list\', or hit \'return\' to go back)> ")
+        if name == "list":
+            for k in data:
+                print(k)
+        elif name == "":
+            return
+        else:
+            dollars = input("Please enter a donation amount (ex: 500.05)> ")
+            for k in data:
+                if k == name:
+                    data[k].append(round(float(dollars), 2))
+                    break
+            else:  # if no name match, make new entry
+                data[name] = [(round(float(dollars), 2))]
+            print(f"Dear {name},\n\nThank you for your donation of {data[name][-1]:.2f}.\n\nLove,\nThe Charity Org\n")
+            return
+
+
+# mode 2 - "create a report"
+def sort_by_dollars(donor_lists):  # bubble sort
+    counter = 0  # counter will ensure we've taken one full pass of the list where all rows are in order before quitting
+    temp_row = []
+    while counter < (len(donor_lists) - 1):  # we will get an out of bounds error if we don't stop at len-1
+        counter = 0
+        for i in range(len(donor_lists) - 1):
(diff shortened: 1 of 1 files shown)


students/mkdir01/lesson04/mailroom2.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python3
+import sys
+
+# changed from list of lists to a dict with values as list
+data = {'John Doe': [120.00, 353.33, 400.00], 'Jane Doe': [3500.04, 2624.44], 'John Smith': [33.33, 400.00, 29.20], 'Jane Smith': [2.00], 'Billy Jo Jones': [100.00, 100.00, 100.00]}
```
natasha-aleksandrova: 👍

students/mkdir01/lesson04/mailroom2.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+# mode 1 - "send a thank you"
+def thank_you():
+    while True:
+        name = input("Please enter a full name (or type \'list\', or hit \'return\' to go back)> ")
+        if name == "list":
+            for k in data:
+                print(k)
```
natasha-aleksandrova: what you have is totally correct, but here is a shortcut version:
```
>>> print('\n'.join(data))
John Doe
Jane Doe
John Smith
Jane Smith
Billy Jo Jones```

File: students/mkdir01/lesson04/mailroom2.py [hunk 3]
```diff
@@ -0,0 +1,110 @@
...
+        elif name == "":
+            return
+        else:
+            dollars = input("Please enter a donation amount (ex: 500.05)> ")
+            for k in data:
+                if k == name:
+                    data[k].append(round(float(dollars), 2))
+                    break
```
natasha-aleksandrova: looks great! to make this even easier I would recommend taking a look at defaultdict from collections module:
https://www.accelebrate.com/blog/using-defaultdict-python/

students/mkdir01/lesson04/mailroom2.py
```diff
@@ -0,0 +1,110 @@
...
+    while counter < (len(donor_lists) - 1):  # we will get an out of bounds error if we don't stop at len-1
+        counter = 0
+        for i in range(len(donor_lists) - 1):
+            if donor_lists[i][1] < donor_lists[i + 1][1]:
+                temp_row = donor_lists.pop(i)
+                donor_lists.insert(i + 1, temp_row)
+            else:
+                counter += 1  # when we get through a whole for loop withought "bubbling" we leave the while loop
+    else:
+        return donor_lists
```
natasha-aleksandrova: I would still recommend using `sorted` instead of custom sort functions. 

it is usually preferred to use something already in place instead of reimplementing for these reasons:
* readability: reader is most likely familiar with the `sorted` function and easy to follow
* avoids bugs: a built-in function or a common library most likely not to have bugs due to its wide usage.

natasha-aleksandrova: let me know if you need a practical example how to fit `sorted` into your code.

students/mkdir01/lesson04/mailroom2.py
```diff
@@ -0,0 +1,110 @@
...
+def main():
+    menu_dict = {1: thank_you, 2: create_report, 3: send_letters, 4: end_program}
+    while True:
+        print("Choose a number:")
+        print("(1) Send a Thank You")
+        print("(2) Create a Report")
+        print("(3) Send letter to everyone")
+        print("(4) quit")
+        mode = input("> ")
+        if mode == "1" or mode == "2" or mode == "3" or mode == "4":
```
natasha-aleksandrova: shortcut tip:
change your `menu_dict` keys to strings or convert `mode` to int, then
```
if mode in menu_dict:
    menu_dict[mode]()
else:
    continue
```

PR #714, opened 2018-06-22
Submitting lesson 1 assignments - Quite the experience.......

ptfbanks:
I seem to have succeeded in finding as many wrong approached to completing this assignment as possible.  I suppose backing into the solution counts for some level of achievement.

### DIFF
--- a/students/ptfbanks/Lesson01/breakme.py
+++ b/students/ptfbanks/Lesson01/breakme.py
@@ -1,0 +1,20 @@
+#syntaxerror
+print string
+
+#Name error
+y=a
+x=45
+print (y+X)
+
+# indentation Errror
+ int (a)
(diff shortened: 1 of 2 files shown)


students/ptfbanks/Lesson01/breakme.py
```diff
@@ -0,0 +1,20 @@
...
+b= "try this"
+print (a+b)
+
+#type error
+b=3
+txt= b[:2]
+
+#Attribute error
+x=3
+x.append (2)
```
natasha-aleksandrova: recommend installing a linter for your editor or running it separate; this will catch pep8 style guideline violations for python.

some examples:
`x.append (2)` -> `x.append(2)`
`b=3` -> `b = 3`

PR #717, opened 2018-06-24
fixes to mailroom2 and added lesson 4 - Trigram Kata

mkdir-washington-edu:
Trigram_Kata.py

### DIFF
--- a/students/mkdir01/lesson04/Trigram_Kata.py
+++ b/students/mkdir01/lesson04/Trigram_Kata.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+
+import random
+storage = {}  # dict storing keys of tuples and values of lists
+global_loop_key0 = ""
+global_loop_key1 = ""
+
+
+def seperate_words(temp_file):
+    temp_chars = list(temp_file)
+    # break apart at " ", do not break at "-", turn "--" into a comma, remove "\n", "at" != "at."
+    for i, char in enumerate(temp_chars):
+        if char == "-":  # turn "--" into ","
+            if temp_chars[i + 1] == "-":
+                temp_chars[i] = ", "
+                temp_chars.pop(i + 1)  # and pop() second dash
+        elif char == "\n":
+            temp_chars[i] = " "
+        elif char == '':
+            temp_chars[i] = " "
+        elif char == '\"':
+            if temp_chars[i + 1] == "\\":
+                temp_chars[i] = " "
+                temp_chars.pop(i + 1)
+
+    return "".join(temp_chars).split(" ")  # turn into a list of words
+
+
+def add_to_storage(word_list):
+    # create dict from list
+    for i in range(len(word_list) - 2):
+        tup = (word_list[i], word_list[i + 1])
+        elem = word_list[i + 2]
+        if elem:
+            if tup in storage and elem not in storage[tup]:
+                storage[tup].append(elem)
+            else:
+                storage[tup] = [elem]
+
+
+def read_file(file_name):
+    f = open(file_name)
+    temp_file = f.read()
+    f.close()
+    # take words and make into a list
+    add_to_storage(seperate_words(temp_file))
+
+
+def write_file(file_name, new_story):
+    f = open(file_name, 'w')
+    f.write(new_story)
+    f.close()
+    return
+
+
+def look_for_pattern(temp_list):
+    for i in range(1, int(len(temp_list) / 2)):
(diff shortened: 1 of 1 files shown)


students/mkdir01/lesson04/Trigram_Kata.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python3
```
natasha-aleksandrova: note module names should also be snake_case -> `trigram_kata`

students/mkdir01/lesson04/Trigram_Kata.py
```diff
@@ -0,0 +1,130 @@
...
+            main = 0  # exit loop
+
+        counter += 1
+
+    write_file(file_name, new_story)
+
+
+def main():
+    read_file(input("input file name: "))
+    create_text(input("output file name: "), input("Minimum character length: "))
```
natasha-aleksandrova: in mailroom we dealt with a global data structure (something that is expected); however normally you wouldn't be dealing with global data structures.

here would be a good case for functions that return a value and pass into next function, like:
``` 
words = read_file_into_words(file_name)
create_new_text(words)
```

students/mkdir01/lesson04/Trigram_Kata.py
```diff
@@ -0,0 +1,130 @@
...
+            if tup in storage and elem not in storage[tup]:
+                storage[tup].append(elem)
+            else:
+                storage[tup] = [elem]
+
+
+def read_file(file_name):
+    f = open(file_name)
+    temp_file = f.read()
+    f.close()
```
natasha-aleksandrova: it is preferred to use `with` statement when working with files.

students/mkdir01/lesson04/Trigram_Kata.py
```diff
@@ -0,0 +1,130 @@
...
+    # take words and make into a list
+    add_to_storage(seperate_words(temp_file))
+
+
+def write_file(file_name, new_story):
+    f = open(file_name, 'w')
+    f.write(new_story)
+    f.close()
+    return
```
natasha-aleksandrova: `with` statement for working with files & no need for `return` statement.

students/mkdir01/lesson04/Trigram_Kata.py
```diff
@@ -0,0 +1,130 @@
...
+            return 0  # if the current tuple matches the initial tuple, main = 0
+        elif (tup[1], storage[tup][i]) in storage:
+            return detect_infinite_loop(tup[1], storage[tup][i])  # else keep searching the dict
+        else:
+            return 1  # we come to a point where we learn the loop is not infinite
+
+
+def create_key():
+    # choose a random key where the first word doesn't end in punctuation
+    tup = random.choice(list(storage.keys()))
```
natasha-aleksandrova: no need to call `.keys`

students/mkdir01/lesson04/Trigram_Kata.py
```diff
@@ -0,0 +1,130 @@
...
+    while tup[0] == '' or tup[0][-1] == ('.' or ',' or '"' or '!' or '?' or ';' or ':' or "'" or ')'):
```
natasha-aleksandrova: better -> `tup[0][-1] in ('.', ',', '!'):`

students/mkdir01/lesson04/Trigram_Kata.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def look_for_pattern(temp_list):
+    for i in range(1, int(len(temp_list) / 2)):
+        # compares -1 to -2, then -1:-2 to -3:-4, etc...
+        if temp_list[-1:((-1 * i) - 1): -1] == temp_list[((-1 * i) - 1)::-1]:
+            global_loop_key0 = temp_list[(-1 * i)]  # first elem of looping list
+            global_loop_key1 = temp_list[(-1 * i) + 1]  # second elem of looping list
+            # copy the loop, from start to end (we've been traversing end to start)
+            return detect_infinite_loop((global_loop_key0, global_loop_key1))  # returns 0 if loop is found
```
natasha-aleksandrova: I'm not I sure I understand the infinite loop

mkdir-washington-edu:
Thanks!

In a reductive example, the infinite loop happens if your dict has:
('Where', 'was') => 'I'
('I', 'was') => 'when'
('was', 'when') => 'I'
('when', 'I') => 'was'

Possible with shorter texts, so I wanted to build something that would
account for that.

PR #721, opened 2018-06-25
Completed String Formatting Lab and Mailroom Pt 1

Parzival-X:
Completed the remaining exercises for Lesson 3.


students/rob_sanchez/Lesson 3/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+                "donations": [
+                    {"donation-number": '1', "amount": '4500'},
+                    {"donation-number": '2', "amount": '1500'}
+                ]},
+              "Adam Sandler": {
+                "donations": [
+                    {"donation-number": '1', "amount": '500'},
+                    {"donation-number": '2', "amount": '2400'}
+                ]}
+              }
```
natasha-aleksandrova: lesson 3 mailroom was technically supposed to use a list for data structure so I guess you jumped ahead 😉 

I would recommend to consider alternate data structure:
`{"Katy Perry": [500, 2400]}`
this should really simplify your code; since you're only storing donations, no need to create a dictionary with _only_ ever having one key. also, `donation-number` is kind of redundant as you will be able to easily get size via `len(donor_dict[name])`

students/rob_sanchez/Lesson 3/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def main():
+    print("\nWelcome!\n")
+    welcome()
```
natasha-aleksandrova: no need for indirection; call `welcome` directly.

students/rob_sanchez/Lesson 3/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+    print("\nSelect an option below:\n")
+
+    selection = welcome_options()
+
+    if selection == '1':
+        send_thank_you()
+    elif selection == '2':
+        create_report()
+    elif selection == '3':
+        sys.exit()
```
natasha-aleksandrova: 👍 for using `sys.exit`

students/rob_sanchez/Lesson 3/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+    headers = ["Donor Name", "Total Given", "Num Gifts", "Average Gift"]
+    print(("{:<30} | {:^15} | {:^15} | {:^15}").format(*headers))
+    print("-"*82)
+
+    donor_report = get_summary()
+
+    for item in donor_report:
+        print(("{:<30} ${:>17} {:>16} ${:>14}").format(*item))
+
+    welcome()
```
natasha-aleksandrova: it is best to use `while True` loop instead of calling `welcome()` in each function.

Imagine you expand your program and having to remember to call `welcome` function to control the main program flow.

this shouldn't be the concern of "feature" functions.

students/rob_sanchez/Lesson 3/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+        for item in history:
+            total_given.append(float(item.get('amount')))
+            num_gifts.append(item.get('donation-number'))
+        donor_names[i].append("{:.2f}".format(sum_values(total_given)))
+        donor_names[i].append(len(num_gifts))
+        donor_names[i].append("{:.2f}".format(sum_values(total_given)/len(num_gifts)))
+        total_given = []
+        num_gifts = []
+        i += 1
+    return donor_names
```
natasha-aleksandrova: the report summary should be sorted (check instructions); I'm not seeing the sorting.

students/rob_sanchez/Lesson 3/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+            num_of_donations.append(item.get('amount'))
+
+        new_index = len(num_of_donations)
+
+        donor_dict[donor_name]["donations"].append(
+                            {"donation-number": '', "amount": ''})
+        donor_dict[donor_name]["donations"][new_index]["donation-number"] = new_index + 1
+        donor_dict[donor_name]["donations"][new_index]["amount"] = float(amt_input)
+
+        send_email(donor_name, amt_input)
```
natasha-aleksandrova: see my comment about data structure above; this can be really simplified with the changes I have suggested.

PR #723, opened 2018-06-25
commnet

vvinodh:


students/vvinodh/Lesson6/test_mailroom.py
```diff
@@ -0,0 +1,36 @@
...
+        "Mark Zuckerberg      $     21864.47           3  $   7288.16\n"
+        "Jeff Bezos           $      1755.66           3  $    585.22\n"
+        "Paul Allen           $       947.56           3  $    315.85\n")
+
+
+    assert expected == actual
+
+
+def test_letters():
+    assert os.path.isfile('William Gates, III_2018-06-25.txt')
```
robalford: You need to call `letters()` function here and use `datetime` module rather than hard coding the dates to get this to pass

PR #724, opened 2018-06-25
Adding the list_lab file for lesson 3.

rfelts:
I'm adding the completed list_lab file for lesson 3. 

My linting is still no working in Sublime Text. I've uninstalled everything and reinstalled/configured it. With no luck. So, I believe my script is pep8 compliant but I'm not positive.


students/rfelts/lesson03/list_lab.py
```diff
@@ -1,2 +1,139 @@
...
+    fruit_list = ["Pineapples"] + fruit_list[:]
+    print(fruit_list, "\n")
+
+    fruit_list.insert(0,"Cherries")
+    print(fruit_list, "\n")
+
+    #Display all the fruits that begin with “P”
+    print("Printing all fruits in the list that start with the letter P:\n")
+    for x in fruit_list:
+        if x[:1] == "P":
```
natasha-aleksandrova: a more common convention is to check on zero element -> `if x[0] == "P"`
or
`if x.startswith("P"):`

PR #726, opened 2018-06-26
Prepared Grit Print script (lesson 2) and re-aligned my readme file.

ptfbanks:
Submission of lesson 2 assignment.
My approach was to first prepare more of a mechanical way to prepare a generic grid matrix.
I then relayed this into a function print_grid of 4 variables, dynamically entered.


ptfbanks:
FYI... My readme file was placed in the detail lesson directory, So I moved it to my student directory.  But I failed to update it with note of my second lesson input.  I'll update that when I submit Lesson03 assignment.

PR #727, opened 2018-06-26
Lesson05 assignment - except_exercise.py, mailroom3.py

wpucilow:
Lesson05 assignment - except_exercise.py, mailroom3.py

### DIFF
--- a/students/Wieslaw_Pucilowski/lesson05/mailroom3.py
+++ b/students/Wieslaw_Pucilowski/lesson05/mailroom3.py
@@ -1,0 +1,180 @@
+#!/usr/bin/env python3
+# lesson04 mailroom part2
+from operator import itemgetter
+from collections import defaultdict
+
+__author__ = "Wieslaw Pucilowski"
+
+
+dict_all = {
+            ('Richard', 'Lionheart'): [100.5, 36.30, 230],
+            ('Andreas', 'Bolen'): [220, 1000],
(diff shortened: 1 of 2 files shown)


students/Wieslaw_Pucilowski/lesson05/mailroom3.py
```diff
@@ -0,0 +1,180 @@
...
+    return greetings
+
+
+def write_letter(first, last):
+    message = donor_greeting(first, last)
+    if message:
+        try:
+            f = open(first+'_'+last+'.txt', 'w')
+            f.write(message)
+            f.close()
```
natasha-aleksandrova: it is preferred to use `with` statement when working with files.

students/Wieslaw_Pucilowski/lesson05/mailroom3.py
```diff
@@ -0,0 +1,180 @@
...
+    try:
+        return {'first_name': first, 'last_name': last,
+                'donations': len(dict_all[(first, last)]),
+                'total': sum(dict_all[(first, last)])}
+    except KeyError:
+        return None
+
+
+def list_donors():
+    for donor in sorted(dict_all.keys()):
```
natasha-aleksandrova: note that you don't need to call `.keys()` here

wpucilow:
thankyou for the comment. mailroom3.py has been updated and new pull request done.

PR #733, opened 2018-06-29
added lesson6

aurel1212:
I used pep8 to check the files. I think there are no formatting issues.

### DIFF
--- a/students/AurelPerianu/Lesson6/test_mailroom_4.py
+++ b/students/AurelPerianu/Lesson6/test_mailroom_4.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+
+import unittest
+import mailroom_4 as mailroom
+import os
+
+
+class MailroomTest(unittest.TestCase):
+
+    def test_donor_amount(self):
+        self.assertEqual(mailroom.donors['Donor A'], [3580, 34124.31, 7654])
+
+    def test_list_donors(self):
+        del mailroom.donors['x']
+        self.assertEqual(list(mailroom.donor_list()), ['Donor A', 'Donor B',
+                         'Donor C', 'Donor D', 'Donor E', 'Donor F'])
+
+    def test_report(self):
+        self.assertTrue(mailroom.report())
+
+    def test_letter(self):
+        self.assertTrue(mailroom.letter('Donor A', 34124.31))
+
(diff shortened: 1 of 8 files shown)


students/AurelPerianu/Lesson6/test_mailroom_4.py
```diff
@@ -0,0 +1,40 @@
...
+
+import unittest
+import mailroom_4 as mailroom
+import os
+
+
+class MailroomTest(unittest.TestCase):
+
+    def test_donor_amount(self):
+        self.assertEqual(mailroom.donors['Donor A'], [3580, 34124.31, 7654])
```
robalford: This test is not necessary since it's really just testing the python dict works as expected

students/AurelPerianu/Lesson6/test_mailroom_4.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def test_list_donors(self):
+        del mailroom.donors['x']
```
robalford: Remember that your tests should be able to run in total isolation from each other -- you can use unittest's setUp and tearDown methods to deal with sharing test data across your individual test cases

students/AurelPerianu/Lesson6/test_mailroom_4.py
```diff
@@ -0,0 +1,40 @@
...
+        self.assertEqual(list(mailroom.donor_list()), ['Donor A', 'Donor B',
+                         'Donor C', 'Donor D', 'Donor E', 'Donor F'])
+
+    def test_report(self):
+        self.assertTrue(mailroom.report())
```
robalford: I would recommend refactoring the `report` function to return the report as a string. That way you could test the dynamically generated output of your program

students/AurelPerianu/Lesson6/test_mailroom_4.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def test_letter(self):
+        self.assertTrue(mailroom.letter('Donor A', 34124.31))
```
robalford: Would refactor `letter` as well to return a string so that you can test that it works properly. Having these functions return strings to be printed elsewhere also makes your program more flexible overall

PR #741, opened 2018-06-30
lesson 5, mailroom 3

mkdir-washington-edu:
mailroom3.py

### DIFF
--- a/students/mkdir01/lesson05/mailroom3.py
+++ b/students/mkdir01/lesson05/mailroom3.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+import sys
+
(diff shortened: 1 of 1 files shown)


students/mkdir01/lesson05/mailroom3.py
```diff
@@ -0,0 +1,105 @@
...
+def main():
+    menu_dict = {"1": thank_you, "2": create_report, "3": send_letters, "4": end_program}
+    while True:
+        print("Choose a number:")
+        print("(1) Send a Thank You")
+        print("(2) Create a Report")
+        print("(3) Send letter to everyone")
+        print("(4) quit")
+        mode = input("> ")
+        if mode in menu_dict:
```
robalford: If you use bracket syntax (`menu_dict[mode]`) to access the dispatch dict, you could use a try/except for `KeyError` here to catch bad user input

PR #743, opened 2018-07-01
Session03 assignment #1.

toddlikesbeer:
I forgot to make frequent commits as I was working on the assignment, I will try and get better at that on future assignments. 

Feedback from last time was that my code was over indented, I believe I have fixed that by making my tab 4 spaces. I need to find out how to do that by default in Sublime 2, as each new file reverts back to actual tabs. 

I believe that is all.   --toddschultz

### DIFF
--- a/students/toddschultz/session03/slicing_lab_exercise.py
+++ b/students/toddschultz/session03/slicing_lab_exercise.py
@@ -1,0 +1,48 @@
+#with the first and last items exchanged.
+def exchanged_string(seq):
+    return seq[-1:] + seq[1:-1] + seq[:1]
+exchanged_string("abcdefghijklmnopqrstuvwxyz")
+
+def exchanged_tuple(seq):
+    return seq[-1:] + seq[1:-1] + seq[:1]
+exchanged_tuple(("airplane", "boat", "car", "donkey", "elevator", "frigate", "glider", "hovercraft", "inflatable"))
+
+#with every other item removed.
+def every_other_tuple(seq):
+    return seq[::2] 
+every_other_tuple(("apple", "banana", "Corn", "Date", "Elderberry", "Fig", "Grape", "Honeydew", "Imbe"))
+
+def every_other_string(seq):
+    return seq[::2] 
+every_other_string("123456789")
+
+#with the first 4 and the last 4 items removed, and then every other item in between.
+def remove_fours_every_other_string(seq):
+    return seq[4:-4:2] 
+remove_fours_every_other_string("----123456789----")
+
+def remove_fours_every_other_list(seq):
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
@toddlikesbeer here is the doc from sublime on how to make this work:
https://www.sublimetext.com/docs/2/indentation.html

students/toddschultz/session03/slicing_lab_exercise.py
```diff
@@ -0,0 +1,48 @@
...
+    return seq[3:6] + seq[6:9] + seq[0:3]
+thirds_list(["apple", "banana", "Corn", "Date", "Elderberry", "Fig", "Grape", "Honeydew", "Imbe"])
+
+def thirds_string(seq):
+    return seq[3:6] + seq[6:9] + seq[0:3]
+thirds_string("123456789")
+
+def thirds_string_len(seq):
+    return seq[int((len(seq) * .34)):int((len(seq) * .67))] + seq[int((len(seq) * .67)):len(seq)] + seq[:int((len(seq) * .34))]
+thirds_string_len("123456789")
```
charlesrobison: This is a creative solution for a viable length sequence!  Another would be to divide the length by 3 - e.g., len(seq)//3

students/toddschultz/session03/slicing_lab_exercise.py
```diff
@@ -0,0 +1,48 @@
+#with the first and last items exchanged.
+def exchanged_string(seq):
+    return seq[-1:] + seq[1:-1] + seq[:1]
+exchanged_string("abcdefghijklmnopqrstuvwxyz")
```
charlesrobison: Don't necessarily need to put examples of how to run the functions after each function - although it's nice to see line of thinking.  You should definitely add some tests though, per the assignment.

students/toddschultz/session03/slicing_lab_exercise.py
```diff
@@ -0,0 +1,48 @@
...
+exchanged_tuple(("airplane", "boat", "car", "donkey", "elevator", "frigate", "glider", "hovercraft", "inflatable"))
+
+#with every other item removed.
+def every_other_tuple(seq):
+    return seq[::2] 
+every_other_tuple(("apple", "banana", "Corn", "Date", "Elderberry", "Fig", "Grape", "Honeydew", "Imbe"))
+
+def every_other_string(seq):
+    return seq[::2] 
+every_other_string("123456789")
```
natasha-aleksandrova: for each task there should be a single function that handles any iterable (string, list, tuple); especially given that your code is identical, there is no need to duplicate the code with a different name.

PR #744, opened 2018-07-02
Completed lesson 4 exercises.

Parzival-X:


students/rob_sanchez/Lesson 4/kata_fourteen.py
```diff
@@ -0,0 +1,80 @@
...
+            for item in raw_text.split():
+                word_list.append(item.lower())
+    return(word_list)
+
+
+# Builds a dictionary from the list of words extracted from a text file
+def build_trigram(in_list):
+    new_dict = {}
+
+    for item in range(len(in_list)-1):
```
robalford: Should this be `len(in_list)-2` to avoid IndexError?

students/rob_sanchez/Lesson 4/kata_fourteen.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+            if (in_list[item], in_list[item+1]) not in new_dict:
+                new_dict[in_list[item], in_list[item+1]] = [in_list[item+2]]
+            else:
+                new_dict[(in_list[item], in_list[item+1])].append(in_list[item+2])
```
robalford: You can use `dict.setdefault` here to simplify

students/rob_sanchez/Lesson 4/mailroom.py
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env python3
+import sys
+import collections
```
robalford: Doesn't look like you're using collections

students/rob_sanchez/Lesson 4/mailroom.py
```diff
@@ -0,0 +1,151 @@
...
+
+# List of donors and donation amounts
+donor_dict = {"Tom Cruise": [100, 200, 300],
+              "Michael Jordan": [1300],
+              "Katy Perry": [4500, 1500],
+              "Adam Sandler": [500, 2400]}
+
+
+def main():
+    welcome()
```
robalford: Don't need to define a function that only calls another function

students/rob_sanchez/Lesson 4/mailroom.py
```diff
@@ -0,0 +1,151 @@
...
+    prompt = "\nChoose an action:\n"
+    menu_sel = ("\n1 - Send a Thank You\n2 - Create a Report\n"
+                "3 - Send letters to everyone\n4 - Quit\n")
+
+    # User selection
+    while True:
+        try:
+            user_selection = input(prompt + menu_sel)
+            options.get(int(user_selection))()
+            # user_selection = input()
```
robalford: It's a good idea to delete old comments before committing your code to git

students/rob_sanchez/Lesson 4/mailroom.py
```diff
@@ -0,0 +1,151 @@
...
+        except ValueError:
+            pass
+        except NameError:
+            pass
+        except TypeError:
+            pass
```
robalford: No need for the except blocks if you're not going to handle the exceptions

students/rob_sanchez/Lesson 4/mailroom.py
```diff
@@ -0,0 +1,151 @@
...
+    name_set = set()
+    for donor, value in donor_dict.items():
+        name_set.add(donor)
+
+    while donor_name.lower() == "list":
+        print ("\nList of Donors: " + print_list(list(name_set)))
+        donor_name = input("\nPlease enter the Donor's full name:\n")
+
+    # Add donor if it doesn't exist in current dictionary
+    if donor_name not in list(name_set):
```
robalford: You can use `collections.defaultdict` or `dict.setdefault` here to simplify

students/rob_sanchez/Lesson 4/mailroom.py
```diff
@@ -0,0 +1,151 @@
...
+def create_report():
+    headers = ["Donor Name", "Total Given", "Num Gifts", "Average Gift"]
+    str_format = "{:<30} ${:>17} {:>16} ${:>14}"
+    print(("\n{:<30} | {:^15} | {:^15} | {:^15}").format(*headers))
+    print("-"*82)
+
+    donor_total = {}
+    num_gifts = {}
+    avg_gift = {}
+    summary = {}
```
robalford: Don't need to define these as empty dicts here

students/rob_sanchez/Lesson 4/mailroom.py
```diff
@@ -0,0 +1,151 @@
...
+
+    donor_total = {key: sum(donor_dict[key]) for key in donor_dict}
+    donor_total = dict(sorted(donor_total.items(), key=lambda t: t[1], reverse=True))
+    num_gifts = {key: len(donor_dict[key]) for key in donor_total}
+    avg_gift = {key: sum(donor_dict[key])/len(donor_dict[key]) for key in donor_total}
```
robalford: I'd figure out a way to perform these operations using the original data structure rather than having to create and keep these additional dicts in sync

students/rob_sanchez/Lesson 4/mailroom.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def send_letters():
+    new_list = {}
+    donor_total = {}
+
+    donor_total = {key: sum(donor_dict[key]) for key in donor_dict}
+
+    for key, value in donor_dict.items():
+        f = open(key+'.txt', 'w')
```
robalford: It's best to use a `with` statement when working with files

PR #746, opened 2018-07-02
First Submission

andykhkwok:
Hi, 
I am submitting the first part of my lesson 2 exercise. I just want to make sure all the processes are correct before turning in the remaining parts. Thanks.

Andy Kwok


students/AndyKwok/lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,16 @@
...
+for n in range(1, 101):
+	#prints out Fizz/Buzz/FizzBuzz accordingly by if the value can be divided equally by assigned values 
+	if n%3 is 0 and n%5 is 0:
+		print('FizzBuzz')
+	elif n%3 is 0:
+		print('Fizz')
+	elif n%5 is 0:
+		print('Buzz')
+	else:
+		print(n)
```
natasha-aleksandrova: couple notes:
* the code is over-indented, per pep8 you should indent by 4 spaces (no tabs); pycharm does this for you, or you can configure an editor like sublime to auto-indent correctly.
* `is` should be used with `is None`, `is True` comparison (it's an object comparison); in other cases you should use `==` equality operator
* don't forget proper spacing (having a linter would help you with that 😉 ) -> `elif n % 5 == 0`

otherwise code logic looks good!

students/AndyKwok/lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,16 @@
+#Description: Program that prints out Fizz upon division of 3 and Buzz upon division of 5
```
natasha-aleksandrova: it isn't a big deal now since these are scripts.. but python modules (aka python files) should follow snake_case convention -> `fizz_buzz.py`

students/AndyKwok/lesson02/GridPrinter.py
```diff
@@ -0,0 +1,39 @@
...
+	for j in range(x):	
+		for k in range(x):
+			print('+' + ' - '*y , end=' ')
+		print('+')
+		for i in range(x):	
+				print('|' + '   '*y , end=' ')
+		print('|')
+	for k in range(x):
+		print('+' + ' - '*y , end=' ')
+	print('+')
```
natasha-aleksandrova: generally good idea to avoid one letter vars; it is ok in certain cases like comprehensions (we'll learn later) but it is best to use meaningful vars so that the reader can easily understand the code.

also, if you would like to describe the function you can use docstrings, here is the format:
```
def print_grid(x, y):
    """Prints a grid with x by y size (etc)
    x: int
    y: int
    """
    pass 
```
there are various formats on how you define params and return .. but as long as you stick to triple double quotes, you should be good style wise.

natasha-aleksandrova:
@andykhkwok your submission on the course website is correct! make sure to remember to make submission for your fizzbuzz assignment as well.

PR #748, opened 2018-07-02
Lesson 3 assignments; lesson 1 updated, lesson 2 initial request. 

toddlikesbeer:
I've updated the Slicing Lab Exercise with feedback from the instructors, including: 
1. Removing the function calls
2. Removing the duplicate functions
3. Adding Asserts. 

This is also the initial turn in for the second homework, List Lab.

### DIFF
--- a/students/toddschultz/session03/list_lab.py
+++ b/students/toddschultz/session03/list_lab.py
@@ -29,11 +29,11 @@ b = input("Which fruit would you like to delete? ")
 fruit.remove(b)
 print(fruit)
 fruit = fruit * 2
-print(fruit)
-c = input ("Which fruit would you like to delete? ")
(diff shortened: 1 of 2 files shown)


students/toddschultz/session03/list_lab.py
```diff
@@ -0,0 +1,61 @@
...
+print(fruit)
+del fruit [-1]
+print (fruit)
+b = input("Which fruit would you like to delete? ")
+fruit.remove(b)
+print(fruit)
+fruit = fruit * 2
+while fruit != []:
+	c = input ("Which other fruit would you like to delete " + str(fruit) + "..? ")
+	for i in fruit:
```
robalford: The outer for loop is not necessary here

PR #749, opened 2018-07-03
Unit Testing, Html Renderer, Circle Class

willhwhite:
Pull request with unit test, html renderer, and circle class assignments

### DIFF
--- a/students/WWhite/lesson06/test_mailroom4.py
+++ b/students/WWhite/lesson06/test_mailroom4.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+
+import unittest
+import mailroom4
+
+
+class TestMailroom(unittest.TestCase):
+
+    def setUp(self):
+        mailroom4.dict_donations = {
+            "bill gates": [25000.00, 3], "john snow": [10.56, 1], "monet holt": [50000.00, 2], "jeff bezos": [123500.09, 3],
+            "john wick": [120000.00, 2]
+        }
+
+    def tearDown(self):
+        pass
+
+    def test_sort_list(self):
+        newList = mailroom4.sort_list()
+        self.assertEqual("jeff bezos", newList[0][0])
+        self.assertEqual("john snow", newList[-1][0])
+
+    def test_add_donation_amount(self):
+        mailroom4.add_donation_amount("test person", 1000)
+        self.assertIn("test person", mailroom4.dict_donations)
+
+    def test_print_letter(self):
+        expected = '''
+Dear Bill Gates,
+
+Thank you for your donation of 1000.00, it is very much appreciated.
+
+Kind Regards,
+Your Favorite Local Charity
+'''
+        self.assertEqual(expected, mailroom4.print_letter("bill gates", 1000))
+
+
+if __name__ == '__main__':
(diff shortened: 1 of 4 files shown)


students/WWhite/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,40 @@
...
+class TestMailroom(unittest.TestCase):
+
+    def setUp(self):
+        mailroom4.dict_donations = {
+            "bill gates": [25000.00, 3], "john snow": [10.56, 1], "monet holt": [50000.00, 2], "jeff bezos": [123500.09, 3],
+            "john wick": [120000.00, 2]
+        }
+
+    def tearDown(self):
+        pass
```
robalford: You don't need to include tearDown in your test case if you're not adding any functionality

students/WWhite/lesson07/html_render.py
```diff
@@ -0,0 +1,105 @@
+/>:class Element:
+    indentation = "  "
+    tag_class = ""
+    attributes = ""
```
robalford: `attributes` should be an instance attribute since the value can be different for different subclass instances

students/WWhite/lesson07/html_render.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __init__(self, content=None, **kwargs):
+        if content:
+            self.content = [content]
+            for key, value in kwargs.items():
+                self.attributes += ' {}="{}"'.format(key, value)
```
robalford: `self.attributes` should not depend on `self.content` since some elements (like `ul`) are instantiated without passing in any content

students/WWhite/lesson07/html_render.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class Title(OneLineTag):
+    tag_class = "title"
+
+
+class H(OneLineTag):
+    tag_class = "h"
+
+    def __init__(self, headerLevel, content=None):
```
robalford: remember to use `snake_case` for args, variable and function names

students/WWhite/lesson08/circle.py
```diff
@@ -0,0 +1,108 @@
...
+    @classmethod
+    def from_diameter(class_object, diameter):
+        if diameter >= 0:
+            return class_object(diameter / 2)
+        else:
+            raise TypeError("Invalid diameter passed")
+
+    @property
+    def radius(self):
+        return self._radius
```
robalford: you don't need to use a `@property` or setter if you aren't adding any special logic or calculations at the time of attribute assignment or access

students/WWhite/lesson08/circle.py
```diff
@@ -0,0 +1,108 @@
...
+print(c1 > c2)
+print(c1 < c2)
+print(c1 == c2)
+print(c2 == c3)
+
+
+circle_list = [Circle(6), Circle(7), Circle(8), Circle(4), Circle(0), Circle(2), Circle(3), Circle(5), Circle(9), Circle(1)]
+print(circle_list)
+circle_list.sort()
+print(circle_list)
```
robalford: This should have unit tests

PR #752, opened 2018-07-03
Added the lesson 03 slicing lab exercise

SLammertink:


students/SLammertink/Lesson03/slicing_lab.py
```diff
@@ -0,0 +1,45 @@
...
+    return y[::2]
+
+# Function with the first 4 and the last 4 items removed, and then every other item in between.
+
+def rem_first_last_every_other(z):
+    return z[4:-4:2]
+
+# Function with the elements reversed (just with slicing).
+
+def reversed(a):
```
natasha-aleksandrova: be careful about over-writing built-in names; you should choose a name that does not conflict.

students/SLammertink/Lesson03/slicing_lab.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Function with every other item removed.
+
+def rem_every_other(y):
+    return y[::2]
+
+# Function with the first 4 and the last 4 items removed, and then every other item in between.
+
+def rem_first_last_every_other(z):
+    return z[4:-4:2]
```
natasha-aleksandrova: instead of using anonymous var names like x, y, z etc.. use a var name that makes it easier for reader to understand what the code does. Something like `seq` or `sequence` is more clear and should be used in all of your functions for consistency.

PR #753, opened 2018-07-03
lesson 08 for review

rgpag:
please review "circle_class.py" for lesson 08

### DIFF
--- a/students/rgpag/lesson08/circle_class.py
+++ b/students/rgpag/lesson08/circle_class.py
@@ -1,0 +1,60 @@
+import math
+
+
+class Circle(object):
+    def __init__(self, the_radius):
+        self._radius = the_radius
+        self._diameter = self._radius*2
+        self._area = math.pi*self._radius**2
+
+    @classmethod
+    def from_diameter(cls_object, the_diam):
+        return cls_object(the_diam/2)
+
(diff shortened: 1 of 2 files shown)


students/rgpag/lesson08/circle_class.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class Circle(object):
+    def __init__(self, the_radius):
+        self._radius = the_radius
+        self._diameter = self._radius*2
+        self._area = math.pi*self._radius**2
+
+    @classmethod
+    def from_diameter(cls_object, the_diam):
```
natasha-aleksandrova: the convention is to use `cls` for the var name (similar to how `self` is used)

students/rgpag/lesson08/test_circle.py
```diff
@@ -0,0 +1,60 @@
...
+    c = Circle.from_diameter(8)
+    assert c.radius == 4
+
+
+def test_s6():
+    c = Circle(2)
+    assert repr(c) == 'Circle(2)'
+
+
+def test_s7():
```
natasha-aleksandrova: it is better to use meaningful names for test functions like -> `def test_equality`

PR #760, opened 2018-07-06
Chapter 7 HTML render

aurel1212:


students/AurelPerianu/Lesson7/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+class Element:
+    """Base class to run functions"""
+    tag = "html"
+    indent = "    "
+
+    def __init__(self, content=None, **kwargs):
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
```
robalford: This could be a one-liner:
```
```

students/AurelPerianu/Lesson7/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+            for key, value in self.atrbs.items():
+                atb += f' {key}="{value}";'
+        return str(atb)
+
+
+class OneLineTag(Element):
+
+    def render(self, resp, cur_ind=0):
+        ax = f'{(cur_ind) * self.indent}<{self.tag}>'
+        ay = f'{self.content[0]}{self.extra_attrib()}</{self.tag}>\n'
```
robalford: It's a best practice to use meaningful variable names for these kinds of assignments and makes your code more readable

students/AurelPerianu/Lesson7/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+    def render(self, resp, cur_ind=0):
+        resp.write(
+            f'{(cur_ind) * self.indent}<{self.tag}{self.extra_attrib()} />\n')
+
+
+class Html(Element):
+    tag = 'html'
+
+    def render(self, resp, cur_ind=0):
+        file_out.write(cur_ind + "<!DOCTYPE html>\n")
```
robalford: `file_out` should be passed into this method as an argument

PR #762, opened 2018-07-06
OO Mailroom

willhwhite:
Pushing final version of OO Mailroom

### DIFF
--- a/students/WWhite/lesson09/oo_mailroom.py
+++ b/students/WWhite/lesson09/oo_mailroom.py
@@ -165,4 +165,4 @@ if __name__ == "__main__":  # If this is the main file, run the below
             break
 
     print("Thank you, the program will now exit")
-    raise SystemExit  # Exit the program
+    raise SystemExit  # Exit the progra
--- a/students/WWhite/lesson09/oo_mailroom.py
+++ b/students/WWhite/lesson09/oo_mailroom.py
@@ -1,0 +1,168 @@
+#!/usr/bin/env python3
+
+# -------------------------------------#
+# Desc: Object Oriented Mailroom
+# Dev: Will White
+# Date: 7/2/2018
+# ChangeLog: (When,Who,What)
+# -------------------------------------#
+import math
+
+
+class Donor(object):
+    def __init__(self, name="", donation=0):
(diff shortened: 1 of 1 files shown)


students/WWhite/lesson09/oo_mailroom.py
```diff
@@ -0,0 +1,168 @@
...
+# -------------------------------------#
+# Desc: Object Oriented Mailroom
+# Dev: Will White
+# Date: 7/2/2018
+# ChangeLog: (When,Who,What)
+# -------------------------------------#
+import math
+
+
+class Donor(object):
```
natasha-aleksandrova: note that in python3 you don't need to extend from object -> `class Donor:` is sufficient

students/WWhite/lesson09/oo_mailroom.py
```diff
@@ -0,0 +1,168 @@
...
+    def __init__(self, name="", donation=0):
```
natasha-aleksandrova: is name attribute really optional?

students/WWhite/lesson09/oo_mailroom.py
```diff
@@ -0,0 +1,168 @@
...
+        self.name = name
+        self.total_donation = 0
+        self.donationCount = 0
+
+        self.donation = donation
+
+    @property
+    def name(self):
+        return self._name
```
natasha-aleksandrova: couple things here:
* your class attributes should be defined in `__init__` (neither `self._name` or `self._donation` are)
* property is useful when attribute needs additional functionality or to ensure that the value cannot be changed. this is not the case here.

students/WWhite/lesson09/oo_mailroom.py
```diff
@@ -0,0 +1,168 @@
...
+        return str_letter
+
+
+class DonorCollection(object):
+    def __init__(self):
+        self.donors = []
+
+    def add_donor(self, name, amount):
+        newDonor = Donor(name, amount)
+        self.donors.append(newDonor)
```
natasha-aleksandrova: pep8: newDonor -> should be `new_donor`

PR #764, opened 2018-07-07
Lesson 4 - Kata

vvinodh:


students/vvinodh/Lesson4/kata.py
```diff
@@ -1,26 +1,52 @@
...
-if __name__ == '__main__':
-    process_file()
+import math
+
+
+
+def randomkey(i):
+
+    return random.choice(list(i.keys()))
+
```
robalford: Include two blank lines between functions for Pep8

students/vvinodh/Lesson4/kata.py
```diff
@@ -1,26 +1,52 @@
...
+def randvalue(i, k):
+
+    return random.choice(i[k])
+
+
+if __name__ == "__main__":
+    book = 'sherlock_small.txt'
+    with open(book, 'r') as fp:
+        input = fp.read()
```
robalford: It's best to avoid using Python built-ins like `input` for variable names

students/vvinodh/Lesson4/kata.py
```diff
@@ -1,26 +1,52 @@
...
+        fp.close()
+
+    input = input.replace('\n', ' ').replace('\r', ' ').lower()
+    punctuation = set(['--', '.', ',', '!', '(', '('])
```
robalford: This can be defined as a set literal, rather than calling `set` on a list

students/vvinodh/Lesson4/kata.py
```diff
@@ -1,26 +1,52 @@
...
+    for p in punctuation:
+        input = input.replace(p, f'{p}')
+    words = input.split()
+
+    trigrams = {}
+    for j in range(len(words) - 1):
+        key = tuple(words[j:j + 1])
```
robalford: This should be `tuple(words[j:j + 2])` to grab a two word pair. You also need to adjust the termination point of the loop to be `len(words) - 2` and `value` to be `words[j + 2]`

PR #765, opened 2018-07-07
Lesson05

smitco:
Except exercise complete.

Exceptions added to mailroom3, plus comments from mailroom2 and an attempt at complying with PEP-8 formatting. 

I looked at lines 82, 90, 102 for possibly using comprehensions, but I struggled to write them in a way that was cleaner than what is already there. Should I keep working on re-writing those loops as comprehensions? Are there other spots I should be using comprehensions?

### DIFF
--- a/students/smitco/lesson05/mailroom3.py
+++ b/students/smitco/lesson05/mailroom3.py
@@ -1,9 +1,10 @@
 # lesson 04 mailroom pt 2
 # !/usr/bin/env python3
 
+
 import os
 import datetime
-from statistics import mean
+
 
 # starting list of 5+ donors and 1-3 donations each
 donors = {"John Travolta": [5000, 7500],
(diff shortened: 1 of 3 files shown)


students/smitco/lesson05/mailroom3.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def report():
+    """Print report of donors and donation details"""
+    donor_sums = {}
+    for d in donors:
+        total = sum(donors[d])
+        count = len(donors[d])
+        average = total/count
+        donor_sums[d] = [total, count, average]
```
natasha-aleksandrova: out of the lines you mentioned in your comment this is the only one that could be a good candidate for comprehensions; you want to avoid losing a lot of readability just for the sake of one-liners 🙂 

```
donor_sums = {k: [sum(v), len(v), sum(v) / len(v)] for d in donors.items()}
```

PR #769, opened 2018-07-08
Completed lesson 5 exercises. 

Parzival-X:


students/rob_sanchez/Lesson 5/mailroom_pt3.py
```diff
@@ -0,0 +1,159 @@
...
+            new_list["total"] = donor_total[key]
+            f.write(create_letter(new_list))
+
+    return False
+
+
+# Helper methods:
+
+# Returns a list of formatted values
+def print_list(in_val):
```
natasha-aleksandrova: `def get_formatted_values` might be better fit for the name since nothing is being printed here?

students/rob_sanchez/Lesson 5/mailroom_pt3.py
```diff
@@ -0,0 +1,159 @@
...
+
+    for key, value in donor_dict.items():
+        # f = open(key+'.txt', 'w')
+        with open(key+'.txt', 'w') as f:
+            new_list["donor_name"] = key
+            new_list["last_donation"] = donor_dict[key][-1]
+            new_list["total"] = donor_total[key]
+            f.write(create_letter(new_list))
+
+    return False
```
natasha-aleksandrova: what's the significance of returning False?

students/rob_sanchez/Lesson 5/mailroom_pt3.py
```diff
@@ -0,0 +1,159 @@
...
+    donor_total = {key: sum(donor_dict[key]) for key in donor_dict}
+    donor_total = dict(sorted(donor_total.items(), key=lambda t: t[1], reverse=True))
+    num_gifts = {key: len(donor_dict[key]) for key in donor_total}
+    avg_gift = {key: sum(donor_dict[key])/len(donor_dict[key]) for key in donor_total}
+
+    for key, value in donor_total.items():
+        print(str_format.format(key, round(float(donor_total[key]), 2),
+              num_gifts[key], round(avg_gift[key], 2)))
+
+    return False
```
natasha-aleksandrova: why return False? function is technically "void", shouldn't need a return.

students/rob_sanchez/Lesson 5/mailroom_pt3.py
```diff
@@ -0,0 +1,159 @@
...
+
+# Sends a thank you email to the selected donor
+def send_thank_you():
+    # Holds a history of the donor's name and donation amount
+    history = {}
+
+    # Create list of current donors
+    name_set = set()
+    for donor, value in donor_dict.items():
+        name_set.add(donor)
```
natasha-aleksandrova: couple things here:
* since you're not using `value` var u should only loop over keys: `for name in donor_dict` should do
* using a set here doesn't really add value since names are unique (cannot have duplicate keys in dict)
* to actually get a list of names -> `names = list(donor_dict)` very simple 🙂

students/rob_sanchez/Lesson 5/mailroom_pt3.py
```diff
@@ -0,0 +1,159 @@
...
+    print("-"*82)
+
+    donor_total = {key: sum(donor_dict[key]) for key in donor_dict}
+    donor_total = dict(sorted(donor_total.items(), key=lambda t: t[1], reverse=True))
+    num_gifts = {key: len(donor_dict[key]) for key in donor_total}
+    avg_gift = {key: sum(donor_dict[key])/len(donor_dict[key]) for key in donor_total}
+
+    for key, value in donor_total.items():
+        print(str_format.format(key, round(float(donor_total[key]), 2),
+              num_gifts[key], round(avg_gift[key], 2)))
```
natasha-aleksandrova: although its great that you got to practice comprehensions it might be best to use regular loops as you can go from making 5 loops to 2 loops.

PR #770, opened 2018-07-09
Lesson10

githubtater:
I tried my best with using map. It was incredibly difficult to keep everything straight while using the class version of mailroom. However, I tried using map/filter/reduce with the non-OO version and found it even more difficult. Nevertheless, here's my final lesson submission.


natasha-aleksandrova:
@githubtater you may review a doc on branching here:
https://uwpce-pythoncert.github.io/PythonCertDevel/supplemental/dev_environment/feature_branching.html

(related to your question in submission)

natasha-aleksandrova:
@githubtater what did you find challenging when working with map/filter/reduce?

same comment as before.. I cannot merge it due to conflict.

githubtater:
My main challenge was extracting the information I needed from the class objects so it could be used by map/filter/reduce.  It's like anything else, the more I use them the more it'll become clearer (The class materials have been great! The problem here is that the student just needs more practice.

By the way, I made a new pull request for lesson10.  I'm not sure what I did to my local git directory, but here's the steps I took to remedy:

  1.  renamed my local directory to somename.OLD/
  2.  Cloned my master fork of UWPCE-PYTHONCert (I created a branch at the beginning of lesson10, but this is not what I ultimately used)
  3.  After cloning, I copied the contents for lesson10 into the same directory as my previous lessons
  4.  I successfully added, committed, and pushed lesson10/ to my git master branch
  5.  Here's the new pull request link:  https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/773<https://github.com/githubtater/Self_Paced-Online.git>

Hopefully that works out better.  Let me know if there's any problems.

Thank you for all your help,

Roy

PR #772, opened 2018-07-09
added a readme file

ltrisolini:


students/ltrisolini/Lesson01/break_me.py
```diff
@@ -0,0 +1,23 @@
...
+def SintaxError():
+    print( 2 / 2 ))
+
+def typeerror():
+    a= 'Hello'
+    a=a + ' World'
+    a=a +5
+    print(a)
+    
+def attributeerror():
```
natasha-aleksandrova: some python style tips:
* keep an eye out for overwriting python built-ins like `NameError` you will notice different syntax highlighting
* function names (and var names) should follow snake_case convention -> `syntax_error`
* multiple words should be separated by underscores -> `type_error`

PR #776, opened 2018-07-11
lesson 8 circle class and tests

mkdir-washington-edu:
circle.py, test_circle.py

### DIFF
--- a/students/mkdir01/lesson08/circle.py
+++ b/students/mkdir01/lesson08/circle.py
@@ -1,0 +1,143 @@
+#!/usr/bin/python
+
+from math import pi
+
+
+class Circle():
+
+    def __init__(self, radius):
+        self._radius = radius
+
+    @property
+    def radius(self):
+        return self._radius
+
+    @property
+    def diameter(self):
+        return self._radius * 2
+
+    @diameter.setter
+    def diameter(self, diam):
+        self._radius = diam / 2
+
+    @property
+    def area(self):
+        return pi * self._radius ** 2
+
+    @classmethod
+    def from_diameter(cls, diam):
+        return cls(diam / 2)
+
+    def __str__(self):
+        return(f'Circle with radius: {self.radius:f}')
+
+    def __repr__(self):
+        return(f'Circle({self.radius})')
+
+    def __add__(self, outer):
+        if type(outer) == Circle:
+            return Circle(self.radius + outer.radius)
+        else:
+            return Circle(self.radius + outer)
+
+    def __radd__(self, outer):
+        if type(outer) == Circle:
+            return Circle(self.radius + outer.radius)
+        else:
+            return Circle(self.radius + outer)
+
+    def __iadd__(self, outer):
+        if type(outer) == Circle:
+            return Circle(self.radius + outer.radius)
+        else:
+            return Circle(self.radius + outer)
+
+    def __mul__(self, outer):
+        if type(outer) == Circle:
+            return Circle(self.radius * outer.radius)
+        else:
+            return Circle(self.radius * outer)
+
+    def __rmul__(self, outer):
+        if type(outer) == Circle:
+            return Circle(self.radius * outer.radius)
+        else:
+            return Circle(self.radius * outer)
+
+    def __imul__(self, outer):
+        if type(outer) == Circle:
+            return Circle(self.radius * outer.radius)
+        else:
+            return Circle(self.radius * outer)
+
+    def __sub__(self, outer):
+        if type(outer) == Circle:
+            return Circle(self.radius - outer.radius)
+        else:
+            return Circle(self.radius - outer)
+
+    def __rsub__(self, outer):
+        if type(outer) == Circle:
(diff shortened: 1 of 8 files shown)


students/mkdir01/lesson08/circle.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class Circle():
+
+    def __init__(self, radius):
+        self._radius = radius
+
+    @property
+    def radius(self):
+        return self._radius
```
robalford: No need to use a `@property` to define an instance attribute if you aren't adding any extra logic or calculations. It's fine to pass set `self.radius` directly via the `__init__` function

mkdir-washington-edu:
Thanks for the feedback!

I'm likely mistaken, but I thought the class videos talked about using this
technique because it keeps someone from being able to use *radius* to set
*_radius*. I.e., helping enforce the fact that _radius, in other languages,
would be a private class.

Am I wrong in my understanding of using @property this way?

robalford:
@mkdir-washington-edu no worries. Once you've opened a pull request from master, all your commits to master will show up in the PR until it is merged. This is fine, I can go ahead and review your Lesson 9 work here if it is ready.

@natasha-aleksandrova let's hold off on merging this then until I have a chance to review Lesson 9.

students/mkdir01/lesson08/circle.py
natasha-aleksandrova: regarding your comment below on this..

step 1 didn't specifically call out making radius to be a property so not required but yes having "protection" on it is best in this case.

however I would recommend using double leading underscores which makes the var "private"; consider below example:
```
>>> class Example:
...     def __init__(self, radius):
...             self._radius = radius
...     @property
...     def radius(self):
...             return self._radius
...
...
>>>
>>> ex = Example(2)
>>> ex._radius = 44
>>> ex.radius
44
>>> class Example:
...     def __init__(self, radius):
...             self.__radius = radius
...     @property
...     def radius(self):
...             return self.__radius
...
>>> ex = Example(2)
>>> ex.__radius = 44
>>> ex.radius
2
```

robalford: @mkdir-washington-edu if you haven't already watched it this optional video from the course materials this is a great [talk](https://youtu.be/HTLu2DFOdTg) on private versus public attribute access, and how to use getters and setters in python

students/mkdir01/lesson09/mailroomOO.py
```diff
@@ -0,0 +1,269 @@
...
+            return(self.donations[key])
+
+    def __setitem__(self, key):
+        pass  # use add_dollars instead
+
+    def add_dollars(self, donation_amt, donation_date=str(date.today())):
+        if donation_date in self.donations:
+            self.donations[donation_date].append(float(donation_amt))
+        else:
+            self.donations[donation_date] = [float(donation_amt)]
```
robalford: This would be a good use case for `dict.setdefault`

students/mkdir01/lesson09/mailroomOO.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def report_total_donations(self):
+        total_amt = 0.00
+        for day in self.donations:
+            for amt in self.donations[day]:
+                total_amt += amt
+        return round(total_amt, 2)
```
robalford: This could be done a bit more concisely with sum and a list comp. Something like:
```
sum([v for value in self.donations.values() for v in value])
```
But if your donations data structure requires you to use a nested for loop every time you need to make one of these calculations, you might consider reworking it for more efficient access

students/mkdir01/lesson09/mailroomOO.py
```diff
@@ -0,0 +1,269 @@
...
+        dates = []
+        for k in self.donations:
+            dates.append(k)
+        dates.sort(reverse=True)
+        return dates[0]
+###
+
+
+###
+class Group_Donors:
```
robalford: Class names should be `CamelCase` with no underscores for Pep8

students/mkdir01/lesson09/mailroomOO.py
```diff
@@ -0,0 +1,269 @@
...
+
+def create_donor_list(obj):
+    # create [[donor name, total donations, # gifts, avg gift], [donor 2 info]]
+    temp_list = []
+    for i, name in enumerate(obj.donor_list):
+        temp_list.append([name, obj[name].report_total_donations(), obj[name].report_total_num_gifts(), (obj[name].report_total_donations() / obj[name].report_total_num_gifts())])
+    sort_by_dollars(temp_list)
+    return temp_list
+
+
```
robalford: You might consider moving some of these module level functions into the `Group_Donors` class for better encapsulation

mkdir-washington-edu:
Thanks! I have lesson 10 ready to go. Should I wait until this is merged,
or should I submit it now? I believe my last day of class is today.

Thanks.

natasha-aleksandrova:
@mkdir-washington-edu you're perfectly fine! thanks

PR #781, opened 2018-07-13
Here's the mailroom file for assignment 3.

rfelts:
I'm now using PyCharm. So hopefully I'm more PEP8 compliant now. 

Also I added a try - except block in the prompt_donation_amount function. But I didn't use a basic Python exception or create a special exception like I would in Java. I'm assuming exceptions and exception handling will be covered in a later lesson/class.

### DIFF
--- a/students/rfelts/lesson03/mailroom.py
+++ b/students/rfelts/lesson03/mailroom.py
@@ -1,0 +1,142 @@
+#!/usr/bin/env python3
+
+# Russell Felts
+# Assignment 3 - Mailroom
+
+# Data structure that holds a list of your donors and a history of the amounts they have donated.
+donor_data = [['Lionel Messi', 100], ['Cristiano Ronaldo', 5000, 25, 9450], ["Gianluigi Buffon", 1000000, 2500.50],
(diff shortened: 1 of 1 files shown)


students/rfelts/lesson03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    input_string = "\nPlease enter one of the following options: " + ', '.join(menu_options) + ": "
+    response = input(input_string)
+
+    while not "quit" == response.lower():
+        if response.lower() == "send thank you":
+            send_thank_you()
+        elif response.lower() == "create report":
+            create_report()
+        elif not any(s.lower() == response.lower() for s in menu_options):
```
robalford: This could be handled with an `else`

students/rfelts/lesson03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    prompt_string = "\nPlease enter " + donor + "'s donation amount: "
+    response = input(prompt_string)
+
+    # Verify the value entered is numeric if not prompt the user for a new value
+    while True:
+        try:
+            response = float(response)
+            return response
+        except:
```
robalford: It's best to catch the specific exception that you expect to be raised so that you can handle it, `ValueError` in this case

PR #782, opened 2018-07-14
Added the mailroom 2 exercise from lesson 04

SLammertink:


SLammertink:
Took me a while but thanks to Natasha's tips I managed to get it right!

students/SLammertink/Lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,101 @@
...
+
+Use dicts where appropriate
+See if you can use a dict to switch between the users selections. See Using a Dictionary to switch for what this means.
+Try to use a dict and the .format() method to do the letter as one big template rather than building up a big string in parts.
+'''
+
+import collections
+import pathlib
+
+donors = collections.defaultdict(list)
```
natasha-aleksandrova: nice job on using defaultdict!

students/SLammertink/Lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def quit():
+    ''' a function to exit the program '''
+    print("Exiting the program now, bye bye! ")
+    pretty()
+    exit()
+
+
+switches = { "1" : sending_a_thank_you, "2" : creating_a_report, "3" : sending_letters_to_everyone, "4" : quit }
```
natasha-aleksandrova: 👍 

`quit` is a python built-in, you should avoid using vars or your function names that overwrite the built-ins; something like `quit_program` would be better
also, pep8, no need for those extra spaces 🙂

students/SLammertink/Lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,101 @@
...
+
+def sending_a_thank_you():
+    '''Checks whether is name is in the name dictionary if not
+    it adds it, then will ask the user for a donation and adds it to that users dictionary '''
+    input_name = input("Please enter the donor's name: ")
+    if input_name not in donors:
+        print('')
+        print("That name is not in the donor list, will add it now.") # notify the user that the name will be added to the list
+        print('')
+        donors[input_name] = [] # Creates a new list for that user name where the donation will be stored
```
natasha-aleksandrova: since you're using a defaultdict for your dictionary, you don't need to initialize new name with an empty list

`donors[input_name].append(input_don)` is sufficient.

students/SLammertink/Lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,101 @@
...
+    corresponds to an action that will be executed '''
+    pretty()
+    while True:
+        print("Please choose from the following menu: \n")
+        for k, v in switch_dict.items():
+            print(k, v)
+        pretty()
+        choice = input("Please enter the choice you want to make: ")
+        pretty()
+        if choice == "list":
```
natasha-aleksandrova: you could add this option to your `switches` dictionary to avoid having a special case in your main loop.

students/SLammertink/Lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,101 @@
...
+            print("List of current donors:\n-----------------------\n")
+            for k in donors.keys():
+                print(k)
```
natasha-aleksandrova: note that you don't need to call `keys` since it will iterate over keys by default.

also here is a one liner 🙂-> `'\n'.join(list(donors))`

students/SLammertink/Lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,101 @@
...
+        print("Thank you {} for your generous donation!".format(input_name)) # format the email message
+        print('')
+
+
+def creating_a_report():
+    ''' a function that will create a simple report '''
+    print("{:^30}{:5}{:^25}{:5}{:^30}{:5}{:^20}{:5}".format('Name', '|', 'Donation','|', 'Number of gifts','|', 'Average gift', '|'))
+    print("{:-<116}".format(''))
+    print(' ')
+    for k, v in donors.items():
```
natasha-aleksandrova: don't forget that this report should be sorted.

PR #789, opened 2018-07-16
Exercise 3 Assignments

ABartles:
Pull request for exercise 3 assignments. 
Please let me know if there are any issues or items I should work on.  Thank you for your time.

### DIFF
--- a/students/ABartles/Exercise3/Mailroom.py
+++ b/students/ABartles/Exercise3/Mailroom.py
@@ -1,0 +1,82 @@
+def Send_Thanks(dnrslst):
+    response = "Not a donor or the word list"
+
+    while response not in dnrslst:
+        print("----------------------------------------")
+        response = input("Who would you like to thank ('first last', 'list', or 'exit'? ")
+        response = response.title()
+
+        if response == "Exit":
+            break
+        elif response == "List":
+            print("----------------------------------------")
+            print("Below is the list of Donors:")
+            print(List_Donors(dnrslst))
+        elif response != "List":
+            print("----------------------------------------")
+            amount = input("How much was dontated by {}? ".format(response))
+            dnrslst = Rec_Donation(dnrslst, response, amount)
+            print("----------------------------------------")
+            Thanks_Email(response, dnrslst)
+
+def List_Donors(dnrslst):
+    name_lst = []
+    for i in dnrslst:
+        if i[0] not in name_lst:
+            name_lst.append(i[0])
+    return name_lst
+
+def Rec_Donation(dnrslst, donor, money):
+    donor_tpl = (donor, money)
+    dnrslst.append(donor_tpl)
+    return dnrslst
+
+def Thanks_Email(donor, dnrslst):
+    money = Sum_Gift(donor, dnrslst)
+    print("Dear {},".format(donor))
+    print("Thank you for your generous gift of {}.".format(money))
+
+def Sum_Gift(donor, dnrslst):
+    total = 0
+    for i in dnrslst:
+        if i[0] == donor:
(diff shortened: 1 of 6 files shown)


students/ABartles/Exercise3/Mailroom.py
```diff
@@ -0,0 +1,82 @@
+def Send_Thanks(dnrslst):
```
robalford: The convention in Python is to use `snake_case` for function and variable names

students/ABartles/Exercise3/Mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+    print("Dear {},".format(donor))
+    print("Thank you for your generous gift of {}.".format(money))
+
+def Sum_Gift(donor, dnrslst):
+    total = 0
+    for i in dnrslst:
+        if i[0] == donor:
+            total = total + int(i[1])
+    return total
+
```
robalford: You might consider using a different data structure, something like 
```
donors = [['Donor', [120, 250]], ['Donor2', [120, 250]],]
```
this would simplify some of your code for dealing with donations and let you use Python built-ins like `sum()` and `len()` on the donation lists to get the total and count of donations.

students/ABartles/Exercise3/Mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+    for i in dnrslst:
+        if i[0] == donor:
+            count = count + 1
+    return count
+
+def Create_Report(dnrslst):
+    dnrNames = List_Donors(dnrslst)
+    print(f'{"Donor Name":<20}{"|":<1}{"Total Given":^15}{"|":<1}{"Num Gifts":^15}{"|":<1}{"Average Gift":>15}')
+    print("-" * 68)
+    iCount = 1
```
robalford: No need to initialize this here

students/ABartles/Exercise3/SlicingLab.py
```diff
@@ -0,0 +1,58 @@
...
+def my_reversed(seq):
+    i = 1
+    while i <= len(seq):
+        if i == 1:
+            newseq = seq[len(seq)-i:]
+        else:
+            newseq = newseq + seq[len(seq)-i:len(seq)-i+1]
+        i += 1
+    return newseq
+
```
robalford: you can use this `seq[::-1]` to reverse a sequence with slicing

students/ABartles/Exercise3/list_lab.py
```diff
@@ -0,0 +1,105 @@
...
+print(lst_s1)
+
+response = input("\nEnter a fruit: ")
+response = response.title()
+lst_s1.insert(0, response)
+print(lst_s1)
+
+print("\nFruits starting with 'p':")
+for i in lst_s1:
+    if i[0] == "P":
```
robalford: You can use `str.lower()` to make this case-insensitive

students/ABartles/Exercise3/list_lab.py
```diff
@@ -0,0 +1,105 @@
...
+        response= response.upper()
+        if response == "N":
+            while fruit in lst_s3: # handles removing duplicates in list
+                lst_s3.remove(fruit)
+            response = "Not N or Y"
+        elif response == "Y":
+            i += 1
+            response = "Not N or Y"
+    if i == len(lst_s3):
+        break
```
robalford: You can accomplish this a bit more concisely by using slicing syntax to loop through a copy and remove the items from the original list with something like this where `fruits` is the original list:
```
for fruit in fruits[:]:  
    while fruit in fruits:  
            fruits.remove(fruit)
```

PR #791, opened 2018-07-16
Chapter 9

aurel1212:
Chapter 9 assignment.

### DIFF
--- a/students/AurelPerianu/Lesson9/mailroom_5.py
+++ b/students/AurelPerianu/Lesson9/mailroom_5.py
@@ -1,0 +1,220 @@
+#!/usr/bin/env python3
+
+
+class Donor:
+
+    def __init__(self, name, donations=None):
+        if ' ' in name:
+            self._first, self._last = name.split(' ')
+        else:
+            self._first = ''
+            self._last = name
+        self._name = name
+        if donations:
+            self._donations = donations
+        else:
+            self._donations = []
+
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def first_name(self):
+        return self._first
+
+    @property
+    def last_name(self):
+        return self._last
+
+    @property
+    def donations(self):
+        return self._donations
+
+    @property
+    def full_name(self):
+        return f"{self._first} {self._last}"
+
+    def __str__(self):
+        return f"{self._first} {self._last} {self._donations}"
+
+    def __repr__(self):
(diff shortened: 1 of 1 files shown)


students/AurelPerianu/Lesson9/mailroom_5.py
```diff
@@ -0,0 +1,220 @@
...
+    def name(self):
+        return self._name
+
+    @property
+    def first_name(self):
+        return self._first
+
+    @property
+    def last_name(self):
+        return self._last
```
natasha-aleksandrova: no need to use property for something like first name or last name since there is no additional special logic.

you can access the attribute directly

students/AurelPerianu/Lesson9/mailroom_5.py
```diff
@@ -0,0 +1,220 @@
...
+            To:       {0:s}
+            Subject:  Your donation of ${1:,.2f}
+            Dear {0:s},\n
+
+            Thank you for your donation of ${1:,.2f}.\n
+            """
+        print(txt.format(self._name, amount))
+
+
+class Donors_All:
```
natasha-aleksandrova: class names should follow `UpperCamelCase` convention; no underscores.

students/AurelPerianu/Lesson9/mailroom_5.py
```diff
@@ -0,0 +1,220 @@
...
+    def __init__(self, donors=None):
+        if donors:
+            self._donors = donors
+        else:
+            self._donors = []
```
natasha-aleksandrova: you can also use shortcuts:
`self.donors = donors or []`
or
`self.donors = donors if donors else []`

students/AurelPerianu/Lesson9/mailroom_5.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _get_donor_names(self):
+        return [donor.name for donor in self._donors]
+
+    def _donation_summary(self):
+        sort_rep = []
+        for donor in self._donors:
+            sort_rep.append([donor.name, donor.total_donations,
+                             donor.len_donations, donor.avg_donations])
+            sort_rep.sort(key=lambda elem: elem[1], reverse=True)
```
natasha-aleksandrova: should this be sorted outside of your for loop?

PR #792, opened 2018-07-17
cindy waldron's assignment from lesson 3

cindywaldron:


students/cindywaldron/lesson3/list_lab.py
```diff
@@ -0,0 +1,71 @@
...
+# enter another fruit
+response = input("Enter another fruit name > ")
+# insert an item to front of the list
+a_list.insert(0, response)
+print("Inserted " + response + " to front of list.")
+#display the list
+print(a_list)
+# print fruit name starts with 'P'
+for item in a_list:
+    if 'P' in item:
```
natasha-aleksandrova: it is best to ensure you're only checking for first letter like `item.startswith('P')` or `item[0] == 'P'`

students/cindywaldron/lesson3/slicing.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python3
+
+def exchange_first_last(seq):
+    return seq[-1:] + seq[1:-1] + seq[:1]
+
+def remove_every_other(seq):
+    return seq[0:len(seq):2]
```
natasha-aleksandrova: can be simplified to `seq[::2]` since you're setting default values there

students/cindywaldron/lesson3/mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+
+if __name__ == '__main__':
+    # Display main menu
+    display_main_menu()
+    selection = 'Select an option (1, 2 or 3) ===>'
+    choice = input(selection)
+    while choice != '3':
+        if ( choice == '1'):
+            send_thank_you()
+        elif ( choice == '2'):
```
natasha-aleksandrova: no need for the parens

students/cindywaldron/lesson3/mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    options = [ '1. Send a Thank You', '2. Create a Report', '3. Quit']
+    option_str = '\n'.join(['\t'+item for item in options])
+    print(option_str)
+
+if __name__ == '__main__':
+    # Display main menu
+    display_main_menu()
+    selection = 'Select an option (1, 2 or 3) ===>'
+    choice = input(selection)
```
natasha-aleksandrova: `display_main_menu` and `input` function calls should go inside of your `while` loop

students/cindywaldron/lesson3/mailroom.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    for donor in donors:
+        print(donor[0])
+
+def display_main_menu():
+    """
+    Display main menu
+    """
+    options = [ '1. Send a Thank You', '2. Create a Report', '3. Quit']
+    option_str = '\n'.join(['\t'+item for item in options])
```
natasha-aleksandrova: 👍 on using `join`

PR #795, opened 2018-07-17
Lesson 9 HW

vvinodh:


students/vvinodh/Lesson9/mailroom_hw9.py
```diff
@@ -0,0 +1,134 @@
...
+
+donor_data = {
+        'William Gates, III': [65000.00, 2, 326892.24],
+        'Mark Zuckerberg': [16000.00, 3, 5465.37],
+        'Jeff Bezos': [877.33, 1, 877.33],
+        'Paul Allen': [708.42, 3, 236.14],
+              }
+
+
+class Donor_ops:
```
natasha-aleksandrova: pep8: class names should be UpperCamelCase -> DonorOps

students/vvinodh/Lesson9/mailroom_hw9.py
```diff
@@ -0,0 +1,134 @@
...
+        for donor in donor_data:
+            list_text += (donor + '\n')
+        return list_text
+
+    def report():
+        columns = ["DONOR", "TOTAL", "AVERAGE", "NUMBER"]
+        report_text = str("\n--------------- Report -------------\n\n")
+        report_text += str('{:<30}{:<15}{:<15}{:<15}\n'.format(*columns))
+        for name in donor_data:
+            donation_info = Donor(donor_data[name])
```
natasha-aleksandrova: instead of creating a Donor object every time you need to use, this class should hold reference to collection of these objects, for example
```
donor_collection = DonorCollection(Donor("jeff", [100, 200]), Donor("bob", [139]))
```
then functions like report can reference collection of donors that was initialized at the time of creating of donor collection object.

natasha-aleksandrova:
@vvinodh I'm going to ahead and merge this since you've put in good effort and are nearing your due date. I would recommend reviewing my comments and making sure you understand the OO concepts.

PR #801, opened 2018-07-19
list_lab.py ready for review

andykhkwok:
List_lab.py is ready for review. 
Mailroom.py requires additional testing to debug prior to submission.

### DIFF
--- a/students/AndyKwok/lesson03/list_lab.py
+++ b/students/AndyKwok/lesson03/list_lab.py
@@ -1,0 +1,102 @@
+# Description: 
+# Author: Andy Kwok
+# Last Updated: 07/14/2018
+# ChangeLog: 
+# v1.0 - Initialization
+
+import os 
+
+#!/usr/bin/env python
+
+list = ['Apples', 'Pears', 'Oranges', 'Peaches']
+
+#Series 1
+def series_one(list):
+    """ Function to display and add item to list
+    list: list of fruit
+    new_list = list after mod in fruit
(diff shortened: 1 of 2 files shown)


students/AndyKwok/lesson03/list_lab.py
```diff
@@ -0,0 +1,102 @@
...
+    list: list of fruit
+    new_list = list after mod in fruit
+    new_fruit: new element to add to list
+    id: index value
+    new_fruit_2: another new element to add to list
+    selection: list item scan
+    """ 
+    print(list)
+    new_fruit = input('Please enter another fruit > ')
+    list = list + [new_fruit]
```
robalford: It's best not to use Python built-ins like `list` for variable names as this will overwrite them

students/AndyKwok/lesson03/list_lab.py
```diff
@@ -0,0 +1,102 @@
...
+    id = input('Number on the list to display? ')
+    print('Number {} on the list is {}'.format(id, list[int(id)-1]))
+    new_fruit_2 = input('Please enter another fruit > ')    
+    list = [new_fruit_2] + list
+    print('Add: ', list)
+    list.insert(0,new_fruit_2)
+    print('Add & insert', list)
+    new_list = []
+    for selection in list:
+        if selection[0] == 'P':
```
robalford: You can use `str.lower()` for case insensitive version

PR #803, opened 2018-07-20
Session03 complete

toddlikesbeer:
I had some git issues part way through session03. WIth Natasha slack assistance I re-cloned, so I think I'm good. Let me know if something was messed up along the way. 

This is getting fun! That mailroom was much more involved than I initially thought it would be. 

Oh, hopefully I got my tab issues settled as well. 

ToddS

### DIFF
--- a/students/toddschultz/session03/mailroom.py
+++ b/students/toddschultz/session03/mailroom.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python3
+
+global donors
+donors = ["Imp", 100.00, "Fred Flintstone", 50.00, 50.00, "Mighty Mouse", 420, 1.00, 1.00, "Road Runner", 999.00, "Papa Smurf", 1000.00, 1000.00, 1000.00]
+
+def prompt():
+	action = input("Would you like to (S)end a Thank You, (C)reate a Report or (Q)uit? ")
+	if action.lower() == "s":
+		thank_you()
+	elif action.lower() == "c":
+		create_report
+	elif action.lower() == "q":
+		print("Have a nice day!")
+	else:
+		print("Please enter S, C, or Q.")
+		prompt()
+
+def thank_you():
+	person = (input("Who would you like to send a Thank You to? "))
+	if person.lower() == "list":
+		try_again()
+	elif person in donors: 
+		donation = float(input("How much was the donation? "))
+		location = donors.index(person)
+		donors.insert((location + 1),donation)
+		print("\n\n" + person + ":\n")
+		print("\tThank you very much for your generouse donation.")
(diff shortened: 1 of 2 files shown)


students/toddschultz/session03/mailroom.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+
+global donors
+donors = ["Baby Huey", 1123.00, 456.00, 1789.00, "Mighty Mouse", 99.99, "Fred Flintstone", 5550.00, 5555.00, "Road Runner", 199999.00, "Papa Smurf", 1001.00, 1002.00, 1003.00]
+
+def prompt():
+    action = input("Would you like to (S)end a Thank You, (C)reate a Report or (Q)uit? ")
```
natasha-aleksandrova: you could add `lower()` at the end to avoid calling it each time:
`action = input("").lower()`

students/toddschultz/session03/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+    if action.lower() == "s":
+        thank_you()
+    elif action.lower() == "c":
+        create_report()
+    elif action.lower() == "q" or action.lower() == "quit":
+        print("Have a nice day!")
+    else:
+        print("Please enter S, C, or Q.")
+        prompt()
```
natasha-aleksandrova: a better alternative to calling `prompt()` all over the place is to have a while True loop so that it forever loops until program is terminated.

students/toddschultz/session03/mailroom.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+
+global donors
+donors = ["Baby Huey", 1123.00, 456.00, 1789.00, "Mighty Mouse", 99.99, "Fred Flintstone", 5550.00, 5555.00, "Road Runner", 199999.00, "Papa Smurf", 1001.00, 1002.00, 1003.00]
```
natasha-aleksandrova: it would be easier (code-wise) to have a list of lists here like:
```
donors  = [["baby huey", 1123, 456], ["mighty mouse", 99, 100]]
```

students/toddschultz/session03/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+        average = 0.00
+        a = donors.index(x)
+        b = a + 1
+        while type(donors[b]) is float:
+            total = total + donors[b]
+            b += 1
+            num += 1
+            if b == c:
+                break
+#Now we have donor, donations, number of donations, total, and average; need to format it.
```
natasha-aleksandrova: don't forget this report is supposed to be sorted (check requirements)

PR #804, opened 2018-07-21
Lesson 09 - OOP Mailroom

mcbraunie:


students/MicahBraun/Lesson 09/mailroom_09.py
```diff
@@ -0,0 +1,420 @@
...
+
+    def __init__(self, name):
+        """
+        Initializing characteristics to carry through module (e.g. name, donations, avg_donation, and getter)
+        :param name:
+        """
+        self.name = name
+        self.donations = []
+        self.avg_donation = 0
+        self.get_avg_donation()
```
natasha-aleksandrova: it makes more sense to call this function `set_avg_donation` or have it as a `property`

PR #805, opened 2018-07-22
Completed lesson02.

astrawmyer:
I'm not sure of the best format for docstrings. Let me know if there is a better way than what I put.


students/astrawmyer/lesson02/series.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def sum_series(n, a=0, b=1):
+    """Calculate the nth number of a fibonacci style series.
+    
+    Parameters:
+        first: This is the position of the series to return.
+        second (optional. default=0): This is the first seed vaalue to add.
+        third (optional, default=1); This is the second seed value to add.
+    """
```
natasha-aleksandrova: there are various formats for docstrings, but what you have works!

you may however adjust to this:
```
Parameters:
    n: this is the position ..
    a: first seed value
    b: second seed value
```

PR #806, opened 2018-07-22
added circle assignment lesson08

stefanjp1:


students/stefanjp1/lesson08/circle.py
```diff
@@ -0,0 +1,49 @@
+from math import pi
+
+class Circle:
+    
+    def __init__(self, the_radius):
+        self.radius = the_radius 
+        self._diameter = the_radius * 2
+        self._area = pi * the_radius ** 2
```
robalford: The calculations for diameter and area should be moved into their `@properties`

students/stefanjp1/lesson08/circle.py
```diff
@@ -0,0 +1,49 @@
...
+        self.radius = value / 2
+    
+    @property
+    def area(self):
+        return self._area
+    
+    @classmethod
+    def from_diameter(cls, diameter):
+        the_radius = diameter / 2
+        return Circle(the_radius)
```
robalford: This can be `return cls(diameter / 2.0)`

PR #809, opened 2018-07-23
mailroom.py submission, list_lab.py updated per comment

andykhkwok:
Mailroom.py is ready for review. Necessary comments will be updated on part 2 in Lesson 4 .
list lab updated per previous comment.
Both are submitted on the website.

Can you please review the trigrams.py exercise and provide guidance if I am understanding the prompt from the exercise correctly? My current understanding is to build a dictionary using a book then use the dictionary to build another story in iteration till no more match can be found.

### DIFF
--- a/students/AndyKwok/lesson04/trigrams.py
+++ b/students/AndyKwok/lesson04/trigrams.py
@@ -1,0 +1,66 @@
+# Description: Trigram Exercise
+# Author: Andy Kwok
+# Last Updated: 07/22/2018
+# ChangeLog: 
+# v1.0 - Initialization
+
+import random
+import re
+
+# Read contents from file
+f = open('sherlock_mid.txt')
+read_data = f.read()
+f.close()
+
+# Split words into a string by , . - ; ( ) " ' ! and space
+split_data = re.split('[,|\.|\-|\;|(\|)|\"|\'|!|\s]+',read_data)
+# Remove '' caused by re.split
+if split_data[-1] == '':
+    del split_data[-1]    
+
+# Generating dictionary
+word_index = {}
+for counter, word in enumerate(split_data):
+    if counter != (len(split_data)-2):
+        next_word = word + ' ' + split_data[counter+1]
+        # Check if key already exist
+        existing_value = word_index.get(next_word, ["Key_Do_not_Exists"])
+        if existing_value == ["Key_Do_not_Exists"]:
+            # Adding new key and value
+            word_index.update({next_word: [split_data[counter+2]]})
+        else:
+            # Adding new value to existing key
+            word_index[next_word].append(split_data[counter+2])
+    else:
+        break
+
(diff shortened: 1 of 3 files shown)


students/AndyKwok/lesson04/trigrams.py
```diff
@@ -0,0 +1,66 @@
...
+            # Adding new value to existing key
+            word_index[next_word].append(split_data[counter+2])
+    else:
+        break
+
+# Seek user to intial word input
+starter_word = input("Please provide two words to develop a trigram sequence> ")
+trigrams_list = starter_word.split(' ')        
+
+while trigrams_list != None:
```
natasha-aleksandrova: it is more pythonic to use `is not None` 

`is` used with comparing objects like `is None` or `is True`
`==` is used with everything else

even more pythonic is to use `while trigrams_list`... however I don't think your trigrams_list would ever be None?

students/AndyKwok/lesson04/trigrams.py
```diff
@@ -0,0 +1,66 @@
...
+# ChangeLog: 
+# v1.0 - Initialization
+
+import random
+import re
+
+# Read contents from file
+f = open('sherlock_mid.txt')
+read_data = f.read()
+f.close()
```
natasha-aleksandrova: it is preferred to use `with` statement when working with files.

students/AndyKwok/lesson04/trigrams.py
```diff
@@ -0,0 +1,66 @@
...
+    del split_data[-1]    
+
+# Generating dictionary
+word_index = {}
+for counter, word in enumerate(split_data):
+    if counter != (len(split_data)-2):
+        next_word = word + ' ' + split_data[counter+1]
+        # Check if key already exist
+        existing_value = word_index.get(next_word, ["Key_Do_not_Exists"])
+        if existing_value == ["Key_Do_not_Exists"]:
```
natasha-aleksandrova: more pythonic:
```
if word_index.get(next_word):
    word_index[next_word].append(split_data[counter + 2])
else:
    word_index[next_word] = [split_data[counter + 2]]
```
alternatively using defaultdict makes things even easier 😉

students/AndyKwok/lesson04/trigrams.py
```diff
@@ -0,0 +1,66 @@
...
+
+# Seek user to intial word input
+starter_word = input("Please provide two words to develop a trigram sequence> ")
+trigrams_list = starter_word.split(' ')        
+
+while trigrams_list != None:
+    # Set key and find key value
+    search_word = trigrams_list[-2] + ' ' + trigrams_list[-1]
+    word_found = word_index.get(search_word)
+    if word_found != None:
```
natasha-aleksandrova: more pythonic:
```
if word_index.get(search_word):
    ....
```

PR #826, opened 2018-07-29
add simple file

au-gold:
This is submittal for print grid problem.

### DIFF
--- a/students/Mintae_K/lesson02/lesson02_fizz.py
+++ b/students/Mintae_K/lesson02/lesson02_fizz.py
@@ -1,0 +1,14 @@
+def fizz_buzz():
+    for i in range(100):
+        check3=(i+1) % 3
+        check5=(i+1) % 5
+        if check3 == 0 and not (check5 == 0):
+            #print(check3)
+            print("fizz")
+        elif check5 == 0 and not (check3 == 0):
+            print("buzz")
+        elif check5 == 0 and (check3 == 0):
+            print("fizzbuzz")
+        else:
+            print(i+1)
+            i=i+1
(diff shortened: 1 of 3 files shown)


students/Mintae_K/lesson02/lesson02_fizz.py
```diff
@@ -0,0 +1,14 @@
+def fizz_buzz():
+    for i in range(100):
```
natasha-aleksandrova: your range should be `range(1, 101)` to start at 1 and include a 100 number

students/Mintae_K/lesson02/lesson02_fizz.py
```diff
@@ -0,0 +1,14 @@
...
+        check3=(i+1) % 3
+        check5=(i+1) % 5
+        if check3 == 0 and not (check5 == 0):
+            #print(check3)
+            print("fizz")
+        elif check5 == 0 and not (check3 == 0):
+            print("buzz")
+        elif check5 == 0 and (check3 == 0):
+            print("fizzbuzz")
```
natasha-aleksandrova: this would flow better if you started from here:
```
if check5 == 0 and check3 == 0:
    print('fizzbuzz')
elif check3 == 0:
    print('fizz')
elif check5 == 0:
    print('buzz')
else:
    print(n)
```
this way your follow if checks are simpler

students/Mintae_K/lesson02/series.py
```diff
@@ -0,0 +1,35 @@
...
+            x = x + [x[i-2] + x[i-1]]
+    print (x)
+    print(x[i])
+    return x[i]
+
+#test to see 10th number in lucas series is equal to 77
+assert lucas(10) == 76
+
+#test to see 10th number in fib series is equal to 77
+assert fibonacci(10) == 34
```
natasha-aleksandrova: it's actually `fibonacci(9) == 34` as per instructions `n` starts with zero index.

PR #827, opened 2018-07-29
Lesson02 Grid Printer Submission

mxwllndrsn:
Submitting a pull request per the assignment.

Thanks!

Max

### DIFF
--- a/students/mxwllndrsn/lesson02/fizz_buzz.py
+++ b/students/mxwllndrsn/lesson02/fizz_buzz.py
@@ -1,0 +1,28 @@
+#uw python 210
+#lesson 02
+#max anderson
+
+# fizzin and a buzzin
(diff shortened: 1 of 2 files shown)


mxwllndrsn:
Added fizzbuzz

students/mxwllndrsn/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,28 @@
...
+#lesson 02
+#max anderson
+
+# fizzin and a buzzin
+# for fun you can select which multiples get fizzed/buzzed
+# defaults to 3 and 5 per assignment
+
+def fizz_buzz(a=3, b=5):
+
+    for i in range(100):
```
natasha-aleksandrova: you can have your range be `range(1, 101)` to start at 1 (avoid checking for zero) and it will include number 100

PR #829, opened 2018-07-30
Lesson02 homework and exercises submission files

tlugosan:
fibonacci_and_friends,py file is a little bit an overdone of the series.py implementation. But I wanted to get as many tries at this implementation as I could.

### DIFF
--- a/students/tlugosan/lesson02/fibonacci_and_friends.py
+++ b/students/tlugosan/lesson02/fibonacci_and_friends.py
@@ -1,0 +1,47 @@
+""" Compute and print a generalization of Fibonacci series when the first 2 numbers are custom."""
+
+
+def iter_gen_fibonacci(input_number, first_number, second_number):
+    """Computes Fibonacci numbers when the numbers are parameterized."""
+    fib_list = [first_number, second_number]
+    if input_number <= 1:
+        return fib_list[:input_number + 1]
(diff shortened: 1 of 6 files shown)


tlugosan:
I am having problems figuring out how to split a long return statement. I tried to split the line with an enter before the "+" of the statement but it failed to pass. What can I do?

students/tlugosan/lesson02/fibonacci_and_friends.py
```diff
@@ -0,0 +1,47 @@
...
+    print(result_list)
+
+
+def recursive_gen_fibonacci(input_number, first_number, second_number):
+    """Computes Fibonacci numbers when the numbers are parameterized."""
+    if input_number == 0:
+        return first_number
+    if input_number == 1:
+        return second_number
+    return recursive_gen_fibonacci(input_number - 1, first_number, second_number) + recursive_gen_fibonacci(input_number - 2, first_number, second_number)
```
natasha-aleksandrova: I believe your question on multi-line is referring to this line:
```
    return recursive_gen_fibonacci(input_number - 1, first_number, second_number) + recursive_gen_fibonacci(
        input_number - 2, first_number, second_number)
```

You can break lines up after opening bracket. If you use something like Pycharm, it can auto-format the code for you.

PR #832, opened 2018-07-31
Lesson 04 Assignments

Swt16:
Lesson 04 Assignments.
Plus a minor modification for the lesson 03 lab

### DIFF
--- a/students/ShinTran/lesson04/kata.py
+++ b/students/ShinTran/lesson04/kata.py
@@ -1,0 +1,78 @@
+'''
+Shin Tran
+Python 210
+Lesson 4 Assignment
+'''
+
+
+import string # For punctuation
+import random # For a random number generator
+
+# To store all the two word combo as keys and the following word as a value
+word_dict = {}
+
+
+# Opens a text file, parses though it line by line, storing two consecutive
+# words as keys, and the following word as a value
+def open_file():
+	global word_dict
+	sentence = []
+	contents = open('sherlock.txt')
+	#contents = open('sherlock_small.txt')
+	for line in contents:
+		unmod_sentence = line
+		for c in string.punctuation:
+			unmod_sentence = unmod_sentence.replace(c,' ')
+		sentence = unmod_sentence.split()
+		for i in range(0, len(sentence)-2):
+			the_key = sentence[i] + " " + sentence[i+1]
+			the_value = sentence[i+2]
+			if the_key not in word_dict:
+				word_dict.setdefault(the_key,[]).append(the_value)
+		sentence = []
+	contents.close()
+#	for key, value in word_dict.items():
+#		print(key + ": " + str(value))
+
+
+# Picks a key by random initially, appends those two words to a list,
+# then pulls the value and appends that to a list as well, take the 2nd
+# word of the key and the value, makes that the new key and repeats
+# based on the word count that's randomly generated, max of 15 words,
+# the loop breaks if no key/value is found
+def make_sentence():
+	global word_dict
+	the_sentence = []
+	first_set = random.choice(list(word_dict.keys()))
+	third_word = random.choice(word_dict[first_set])
+	the_sentence.append(first_set.split()[0])
+	the_sentence.append(first_set.split()[1])
+	the_sentence.append(third_word)
+	word_count = random.randint(5,14)
+	temp_key = ""
+	for i in range(1, word_count):
+		temp_key = str(the_sentence[i]) + " " + str(the_sentence[i+1])
+		next_word = word_dict.get(temp_key, 'NoKeyFound')
+		if next_word != 'NoKeyFound':
+			the_sentence.append(next_word[0])
+		else:
+			break
+	make_proper(the_sentence)
+
+
+# Takes a list as a parameter, turns that into a string using a for loop,
+# capitalizes the first character,strips off the space at the end,
+# and appends a period to the end so it's a sentence
+def make_proper(sentence):
+	s = ""
+	for word in sentence:
+		s += word + " "
+	s = s[:-1].capitalize() + '.'
+	print(s)
+
+# Python program to use main for function call
+# runs the program to make 15 sentences
+if __name__ == "__main__":
+    open_file()
+    for j in range(0, 15):
+	    make_sentence()
--- a/students/ShinTran/lesson04/mailroom2.py
+++ b/students/ShinTran/lesson04/mailroom2.py
@@ -1,0 +1,137 @@
+'''
+Shin Tran
+Python 210
+Lesson 4 Assignment
+'''
+
+master_list = [] # For all donations made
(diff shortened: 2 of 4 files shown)


students/ShinTran/lesson04/kata.py
```diff
@@ -0,0 +1,78 @@
...
+# To store all the two word combo as keys and the following word as a value
+word_dict = {}
+
+
+# Opens a text file, parses though it line by line, storing two consecutive
+# words as keys, and the following word as a value
+def open_file():
+	global word_dict
+	sentence = []
+	contents = open('sherlock.txt')
```
natasha-aleksandrova: it is preferred to use `with` statement when working with files. (refer to course materials)

students/ShinTran/lesson04/kata.py
```diff
@@ -0,0 +1,78 @@
...
+import random # For a random number generator
+
+# To store all the two word combo as keys and the following word as a value
+word_dict = {}
+
+
+# Opens a text file, parses though it line by line, storing two consecutive
+# words as keys, and the following word as a value
+def open_file():
+	global word_dict
```
natasha-aleksandrova: it is best to avoid using `global` unless really needed.

in this case your function could return `word_dict` value instead of using it as a global.

students/ShinTran/lesson04/kata.py
```diff
@@ -0,0 +1,78 @@
...
+
+# Picks a key by random initially, appends those two words to a list,
+# then pulls the value and appends that to a list as well, take the 2nd
+# word of the key and the value, makes that the new key and repeats
+# based on the word count that's randomly generated, max of 15 words,
+# the loop breaks if no key/value is found
+def make_sentence():
+	global word_dict
+	the_sentence = []
+	first_set = random.choice(list(word_dict.keys()))
```
natasha-aleksandrova: note that `.keys()` is not necessary here

students/ShinTran/lesson04/kata.py
```diff
@@ -0,0 +1,78 @@
...
+	third_word = random.choice(word_dict[first_set])
+	the_sentence.append(first_set.split()[0])
+	the_sentence.append(first_set.split()[1])
```
natasha-aleksandrova: shortcut: `the_sentence.extend(first_set.split() + third_word)`

students/ShinTran/lesson04/kata.py
```diff
@@ -0,0 +1,78 @@
...
+	the_sentence.append(third_word)
+	word_count = random.randint(5,14)
+	temp_key = ""
+	for i in range(1, word_count):
+		temp_key = str(the_sentence[i]) + " " + str(the_sentence[i+1])
+		next_word = word_dict.get(temp_key, 'NoKeyFound')
+		if next_word != 'NoKeyFound':
+			the_sentence.append(next_word[0])
+		else:
+			break
```
natasha-aleksandrova: please ensure your editor is set to 4 spaces (no tabs); all of the code is over-indented.

students/ShinTran/lesson04/kata.py
```diff
@@ -0,0 +1,78 @@
...
+	third_word = random.choice(word_dict[first_set])
+	the_sentence.append(first_set.split()[0])
+	the_sentence.append(first_set.split()[1])
+	the_sentence.append(third_word)
+	word_count = random.randint(5,14)
+	temp_key = ""
+	for i in range(1, word_count):
+		temp_key = str(the_sentence[i]) + " " + str(the_sentence[i+1])
+		next_word = word_dict.get(temp_key, 'NoKeyFound')
+		if next_word != 'NoKeyFound':
```
natasha-aleksandrova: it is more pythonic to check if return is "falsy", meaning:
```
if word_dict.get(temp_key):
    the_sentence.append(next_word[0])
else: 
    break
```

students/ShinTran/lesson04/mailroom2.py
```diff
@@ -0,0 +1,137 @@
...
+def action(user_input):
+    create_name_list()
+    switch_dict = {
+        'list': list_func,
+        '1': send_thanks,
+        '2': print_report,
+        '3': send_letters,
+        '4': quit
+    }
+    if user_input in (switch_dict):
```
natasha-aleksandrova: no need for extra parens around switch_dict

students/ShinTran/lesson04/mailroom2.py
```diff
@@ -0,0 +1,137 @@
...
+        func = switch_dict.get(user_input)()
+    else:
+        action(main_prompt())
+    action(main_prompt())
```
natasha-aleksandrova: instead of manually handling flow this should be using while True loop.

students/ShinTran/lesson04/mailroom2.py
```diff
@@ -0,0 +1,137 @@
...
+        ["Patricia Brown", 19184.81],
+        ["William Rodriguez", 51227.53],
+        ["Jennifer Miller", 53514.94],
+        ["Michael David", 31051.46],
+        ["Mary Johnson", 71646.16],
+        ["William Rodriguez", 69244.21],
+        ["Mary Johnson", 12689.07],
+        ["Michael David", 57145.50],
+        ["Patricia Brown", 81679.15],
+    ]
```
natasha-aleksandrova: in lesson4 this should be using a dictionary data structure; example: `donors = {"Jenn": [999], "Will": [30, 20]}`

students/ShinTran/lesson04/mailroom2.py
```diff
@@ -0,0 +1,137 @@
...
+def send_letters():
+    global master_list
+    global name_list
+    donation_sum = 0
+    message = "Dear {:s},\n\
+    Thank you for the your generous total donation of ${:,.2f}.\n\
+    Sincerely,\n\
+    Your Local Charity"
+    for name in name_list:
+        output = open(name + ".txt",'w')
```
natasha-aleksandrova: it is preferred to use `with` statement when working with files.

students/ShinTran/lesson04/kata.py
Swt16: the_sentence.extend(first_set.split() + third_word)
TypeError: can only concatenate list (not "str") to list

Ended up using this:
the_sentence.extend(first_set.split())
the_sentence.append(third_word)

students/ShinTran/lesson04/kata.py
Swt16: I have no idea what happened there, it's always set to 4 spaces by default from the first or second week.

students/ShinTran/lesson04/kata.py
natasha-aleksandrova: oh yes! third_word is a string; in this case you can do:
```
the_sentence.extend(first_set.split() + [third_word])
```

PR #834, opened 2018-07-31
Lesson03 slicing exercise

tlugosan:
Full implementation of all sub tasks. I still have to work on committing interim coding and not full solution.

### DIFF
--- a/students/tlugosan/lesson03/slicing.py
+++ b/students/tlugosan/lesson03/slicing.py
@@ -1,0 +1,64 @@
+"""Slicing excercises"""
+
+
+def exchange_first_last(a_seq):
+    """Return a copy of a sequence that has first and last item swapped"""
+    return a_seq[-1:] + a_seq[1:len(a_seq) - 1] + a_seq[:1]
+
+
+def remove_odd(a_seq):
+    """Return a copy of a sequence that has first and last item swapped"""
+    return a_seq[::2]
+
+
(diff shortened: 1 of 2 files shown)


students/tlugosan/lesson03/slicing.py
```diff
@@ -0,0 +1,64 @@
+"""Slicing excercises"""
+
+
+def exchange_first_last(a_seq):
+    """Return a copy of a sequence that has first and last item swapped"""
+    return a_seq[-1:] + a_seq[1:len(a_seq) - 1] + a_seq[:1]
```
natasha-aleksandrova: middel portion can be simplified by using -1 slice -> `a_seq[1:-1]`

students/tlugosan/lesson03/slicing.py
```diff
@@ -0,0 +1,64 @@
...
+def remove_odd(a_seq):
+    """Return a copy of a sequence that has first and last item swapped"""
+    return a_seq[::2]
+
+
+def remove_outer_and_skip(a_seq):
+    """Return a copy of a string without first and last 4 and skip"""
+    if len(a_seq) < 9:
+        print("Sequence requires at least 10 paramaters to work.")
+    return a_seq[4:len(a_seq) - 4:2]
```
natasha-aleksandrova: this can be simplified to -> `a_seq[4:-4:2]`

PR #837, opened 2018-08-02
Complete list_lab file

tlugosan:
Could you please give me some suggestion on how to address the docstring issues. Should i used doc string to document each subtask that you ask us to implement or should I use just # and only for the big task to use docstring. Or it doesn't matter here.

### DIFF
--- a/students/tlugosan/lesson03/list_lab.py
+++ b/students/tlugosan/lesson03/list_lab.py
@@ -40,6 +40,7 @@ print(fruit_list)
 print()
 
 """Display all the fruits that start with P using a for loop"""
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
@tlugosan great question; you usually use docstrings with functions/methods, like:
```
def a_function():
    """This function does function things"""
    pass
```

where scripted style code just uses # for comments:
```
# Implement a task to add fruit
fruits.append("apple")
```

but either way not a big deal here 🙂

natasha-aleksandrova:
sounds good 👍

PR #841, opened 2018-08-03
Lesson 04 Submission

andykhkwok:
Non-mailroom exercise for Lesson 5 is already ready for review. 
Thank you. .

### DIFF
--- a/students/AndyKwok/lesson04/mailroom_part2.py
+++ b/students/AndyKwok/lesson04/mailroom_part2.py
@@ -1,0 +1,81 @@
+# Description: Mailroom Program (Database Organizer)
+# Author: Andy Kwok
+# Last Updated: 07/23/2018
+# ChangeLog: 
+# v1.0 - Initialization
+# v2.0 - Adopt file creation and dictionary types
+
+#!/usr/bin/env python3
+
+if __name__ == '__main__':    
+    database = {'Donor A': [2300.30, 1000.01, 2000.00],
+                'Donor B': [1394.12],
+                'Donor C': [14349.00, 175.95],
+                'Donor D': [42800.11, 1500.04, 3534.20],
+                'Donor E': [12312.23, 250.50]}
+    
+def thank_you_select(database):
+    """ Function
+    """
+    entry = input('Enter donor\'s name to search or \'list\' to show all donors> ')
+    if entry == 'list':
+        database_print(database)
+    else:
+        database = add_donation(entry, database)
+
+def report_select(database):
+    """ Function to sort the database and printout the sorted database
+    """
+    database = dict(sorted(database.items(), key=lambda x: x[1], reverse=True))
+    database_print(database)
+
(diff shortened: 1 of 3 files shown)


students/AndyKwok/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,81 @@
...
+# v2.0 - Adopt file creation and dictionary types
+
+#!/usr/bin/env python3
+
+if __name__ == '__main__':    
+    database = {'Donor A': [2300.30, 1000.01, 2000.00],
+                'Donor B': [1394.12],
+                'Donor C': [14349.00, 175.95],
+                'Donor D': [42800.11, 1500.04, 3534.20],
+                'Donor E': [12312.23, 250.50]}
```
natasha-aleksandrova: the `if __name__` check usually goes at the bottom of the script.

natasha-aleksandrova: 👍 on the new data structure

students/AndyKwok/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,81 @@
...
+def report_select(database):
+    """ Function to sort the database and printout the sorted database
+    """
+    database = dict(sorted(database.items(), key=lambda x: x[1], reverse=True))
+    database_print(database)
+
+def letters_everyone(database):
+    for k, v in database.items():
+        file = open(k.replace(" ", "_") + ".txt", "w")
+        file.write(send_thank_you(k, v))
```
natasha-aleksandrova: it is preferred to use `with` statement when working with files and `file` is a reserved keyword in python (notice special highlighting here)

students/AndyKwok/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,81 @@
...
+    return printout
+
+def quit_select(database):
+    """ Function to trigger the program to exit
+    """
+    global option
+    option = 'quit'
+
+option = 'start'
+menu = {'1': thank_you_select, '2': report_select, '3': letters_everyone, '4': quit_select}
```
natasha-aleksandrova: 👍

students/AndyKwok/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,81 @@
...
+    print(
+        '''    
+        1 - Sent a Thank you
+        2 - Create a Report
+        3 - Send Letters to Everyone
+        4 - Quit
+        '''
+        )
+    option = input('Please select an option> ')
+    menu[option](database)
```
natasha-aleksandrova: the entry point should go into the `if __name__ ` block; it will matter more when you get to lesson 6 unit testing where you don't want this block of code to run on `import`

you also don't _have_ to pass in `database` reference to each function as it is in the global scope.

PR #844, opened 2018-08-04
Lesson 5 assignment from Cindy Waldron

cindywaldron:


students/cindywaldron/lesson5/mailroom.py
```diff
@@ -0,0 +1,109 @@
...
+    for k, v in sorted(donors.items(), key=get_key, reverse=True):
+        a_row = '{:20}  $ {:>10.2f}  {:>10d}  $ {:>11.2f}'.format(k, v[0],v[1],v[2])
+        print(a_row)
+
+def display_donors():
+    """
+    Display donor names
+    """
+    new_list = [ donor for donor in donors.keys()]
+    print(*new_list, sep = "\n")
```
natasha-aleksandrova: this can be simplified to:
`print("\n".join(donors))`

students/cindywaldron/lesson5/mailroom.py
```diff
@@ -0,0 +1,109 @@
...
+    selection = 'Select an option (1, 2, 3, or 4) ===>'
+    switch_function_dict = {'1': send_thank_you, '2': create_report, '3': send_letters, '4': quit}
+    while True:
+        # Display main menu
+        display_main_menu()
+        try:
+            choice = input(selection)
+            if switch_function_dict[choice]() == 'exit menu':
+                break
+        except KeyError:
```
natasha-aleksandrova: it i best to wrap try/except around code you expect to throw that exception (in this case lines 106, 107)

PR #845, opened 2018-08-04
Lesson09 assignment - Object Oriented Mailroom

wpucilow:
class Donor and class DonorsCollection, menu is a part of DonorsCollection.

### DIFF
--- a/students/Wieslaw_Pucilowski/lesson09/mailroom5_OO.py
+++ b/students/Wieslaw_Pucilowski/lesson09/mailroom5_OO.py
@@ -40,8 +40,6 @@ class Donor():
     def number(self):
         return len(self.donations)
 
-    # TODO replace positional args by **kwargs or
-    # TODO make name display flexible to print out any James Sr Junior
     @property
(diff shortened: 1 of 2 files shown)


students/Wieslaw_Pucilowski/lesson09/mailroom5_OO.py
```diff
@@ -0,0 +1,207 @@
...
+            'q': self.quit_menu, }
+
+    def menu_selection(self, prompt, dispatch_dict):
+        while True:
+            response = input(prompt)
+            try:
+                if dispatch_dict[response]() == "exit menu":
+                    break
+            except KeyError:
+                print(response, self.wrong_option)
```
natasha-aleksandrova: per our conversation earlier; neither input function nor "main program flow" should reside in the data class that handles donor collections.

The simplest uses case is you are now asked to put mailroom features behind a GUI  - you would now have to refactor this class. The "separation of concern" idea is separate "user interaction code" from "data code".

Hope this helps.

PR #848, opened 2018-08-04
lesson03 works

au-gold:
Added first exercise: Slicing Lab Exercise

other exercises will be worked later date

### DIFF
--- a/students/Mintae_K/lesson03/Slicing_lab_exercise.py
+++ b/students/Mintae_K/lesson03/Slicing_lab_exercise.py
@@ -1,0 +1,3 @@
+def slicing_exercise(x):
+    s= x[-1:]+x[1:-1]+x[0]
+    print ("with the first and last items exchanged",s)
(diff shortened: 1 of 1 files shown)


students/Mintae_K/lesson03/Slicing_lab_exercise.py
```diff
@@ -0,0 +1,31 @@
...
+#next
+def reverse_sort(x):
+    #print ("with the elements reversed (just with slicing):",s)
+    s = x[::-1]
+    return s
+    #next
+def mid3last3first3(x):
+    third = int(len(x)/3)
+    s = x[third:-third] + x[-third:] + x[:third]
+    return s
```
natasha-aleksandrova: couple small comments:
no need for separate return, instead of:
```
s = x[::-1]
return s
```
return
`return x[::-1]`

and make sure to have two separate blank lines between functions.

PR #849, opened 2018-08-04
Lesson03 String Lab

mxwllndrsn:
Submitting per assignment requirements. 

Thanks,

Max


mxwllndrsn:
Last night noticed it throwing an error when pulling upstream, opens sublime text and asks for a message describing reason. Unsure how to proceed. 

Following directions on 'syncing a fork' produces same issue.

Did I commit to wrong repository? I verified before pushing, didn't do anything different.

PR #851, opened 2018-08-04
Submitting lesson 3, list lab exercise.

ptfbanks:
'Eager to get on with the next exercises.

### DIFF
--- a/students/ptfbanks/Lesson03/3-2_list_lab.py
+++ b/students/ptfbanks/Lesson03/3-2_list_lab.py
@@ -1,0 +1,64 @@
+print ("Today we are going to be playing with fruits.  It's a healthy exercise.  Don't take it personally")
+print ("We start with a list of:")
+fruits = ["Apples", "Pears", "Oranges", "Peaches"]
+print (fruits, "\n")
+add_entry = input("Enter your additional favorite fruit: ")
+fruits.append(add_entry)
+print ("Thank you.  The fruit sequence is now:")
+print(fruits,"\n")
+
+i = int(input("Now, in that list, which is your favoite?, Select by number, 1, 2,... :  "))
+print("Ahh, Good Choice, number", (i), " ", fruits[i-1], "\n")
+
+add_entry = input("Another fruit to lead the list? ")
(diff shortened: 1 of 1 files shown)


students/ptfbanks/Lesson03/3-2_list_lab.py
```diff
@@ -0,0 +1,64 @@
+print ("Today we are going to be playing with fruits.  It's a healthy exercise.  Don't take it personally")
```
charlesrobison: print() is a function, so you can remove the space between print and the parentheses.

students/ptfbanks/Lesson03/3-2_list_lab.py
```diff
@@ -0,0 +1,64 @@
...
+
+add_entry = input("Please add another fruit to lead the list?")
+
+fruits.insert(0,add_entry)
+print (fruits, "\n")
+
+print ("interesting.....  All fruits beginning with 'P' include:")
+p_fruits = []
+for element in fruits:
+    if element[0]== "p":
```
charlesrobison: Good idea to use lower() here to make it case-insensitive.  i.e., if element[0].lower() == "p":

students/ptfbanks/Lesson03/3-2_list_lab.py
```diff
@@ -0,0 +1,64 @@
...
+for fruit in (like_fruits):
+    a = input(f"Do you like the fruit {fruit} (y/n):")
+    while a != "n" and a != 'y':
+        a = input(f"'Sorry, {a} will not work. Please enter 'y' or 'n' :")
+    if a == 'n':
+        x_fruit.append(fruit)
+print ("fruits you do not care for: ", x_fruit)
+for fruit in (x_fruit):
+    like_fruits.remove(fruit)
+print("That leaves accepable fruits as: ", like_fruits)
```
charlesrobison: Typo - acceptable

ptfbanks:
1) Thanks for the case-insensitivity tip
3) Typo...I chose math - I never could get English right.

PR #856, opened 2018-08-06
mailroom5_OO.py refractored to fill separation of concerns requirement

wpucilow:
Code changed to match "separation of concern" requirement.
I works however I'm not sure method definition inside Main class. some (DonorsCollection) are @staticmethod, others are just methods defined inside class Main, others are Main @classmethods (based on methods defined inside the class)

### DIFF
--- a/students/Wieslaw_Pucilowski/lesson09/mailroom5_OO.py
+++ b/students/Wieslaw_Pucilowski/lesson09/mailroom5_OO.py
@@ -83,51 +83,6 @@ class DonorsCollection():
         self._donors = {}
         self.donors = val
         self.rep = ""
-        self.wrong_option = "{:<20}".format(" - Wrong option !!!")
-        self.main_prompt = """
-        {:-^30}
-
-        1 - Send a Thank You
-        2 - Create a Report
-        3 - Send letters to everyone
-        q - Quit
-    """.format(' Main Menu ')
-        self.sub_prompt = """
-        {:-^30}
-
-        1 - Add new donor, donation
-        2 - List donors
-        q - Go to Main Menu
-
-    """.format(' Add/List donors ')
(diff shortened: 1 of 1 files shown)


wpucilow:
2 methods DonorsCollection class have been updated: add_donor() and create_report()

students/Wieslaw_Pucilowski/lesson09/mailroom5_OO.py
```diff
@@ -191,17 +157,94 @@ def write_letters(self):
...
+
+    @staticmethod
+    def donor():
+        dd.add_donor()
+
+    @staticmethod
+    def show():
+        dd.display()
+
+    def quit(msg):
```
natasha-aleksandrova: this is not a valid method. if you want to have a regular function inside a class then it needs to be a staticmethod.

students/Wieslaw_Pucilowski/lesson09/mailroom5_OO.py
```diff
@@ -191,17 +157,94 @@ def write_letters(self):
...
+                response = input(prompt)
+                try:
+                    if dispatcher[response]() == "exit menu":
+                        break
+                except KeyError:
+                    print(response, "Wrong response !")
+
+    @classmethod
+    def sub_menu(cls):
+        cls.menu_selection(submenu, subfeatures)
```
natasha-aleksandrova: this and sub_quit should not be classmethods, all of these functions should be staticmethods (which was another reason I suggested regular methods otherwise you're just wrapping all of the staticmethods into a single class, like grouping which is Ok but not really necessary).

I recommend reviewing the classmethod vs staticmethod:
https://www.geeksforgeeks.org/class-method-vs-static-method-python/

PR #858, opened 2018-08-06
lesson003: mailroom exercise

tlugosan:
How do I validate the input to make sure that it fits the requirement. In both cases i need the input() as a string but also as an int or float. 
line 20: if not choice_actions.isnumeric() or int(choice_actions) not in range(1, len(actions) + 1)
line 38: if type(donation_amount) is not float:

### DIFF
--- a/students/tlugosan/lesson03/mailroom.py
+++ b/students/tlugosan/lesson03/mailroom.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+""" Automate a script to send a thank you mail, print a report of donor or quit the program."""
+
+table_dictionary = {
+    'Toni Orlando': [150.00, 200.00, 100.00],
+    'Amanda Clark': [1800.00],
+    'Robin Hood': [1234.56, 4500.34, 765.28],
+    'Gina Travis': [523.10, 75.00],
+    'Mark Johnson': [850.00, 20.14]
+}
+
+
+def select_action(actions):
+    """User selects an action by its corresponding order number."""
+    print('What would you like to do?')
+    for i, item in enumerate(actions):
+        print("{}) {}".format(i + 1, item))
+    choice_actions = input('Pick your selection by their corresponding number: ')
+    while True:
+        if not choice_actions.isnumeric() or int(choice_actions) not in range(1, len(actions) + 1):
+            choice_actions = input('Pick your selection by their corresponding number: ')
+        else:
+            break
+    choice_actions_int = int(choice_actions)
+    return choice_actions_int
+
+
+def sending_thank_you():
+    """Lists all the donors or prompts for a name and donation amount to compile the thank you email."""
+    send_to_name = input("Who do you want to send the email to? Type 'list' for a list of all the donors. ")
+    if send_to_name == 'list':
+        display_donor_dictionary(table_dictionary)
+        send_to_name = input("Who do you want to send the email to? Type 'list' for a list of all the donors. ")
+    if send_to_name not in table_dictionary.keys():
+        table_dictionary[send_to_name] = []
+    donation_amount = float(input("What donation amount do you want to thank them for? "))
+    while True:
+        if type(donation_amount) is not float:
+            donation_amount = float(input('What donation amount do you want to thank them for? '))
+        else:
+            break
+    table_dictionary[send_to_name].append(donation_amount)
+    print("Dear {}, Thank you for your generous contribution of ${:.2f} to our program.".format(send_to_name,
+                                                                                                donation_amount))
+
+
+def display_donor_dictionary(donor_dictionary):
+    """Prints the full list of donors and corresponding donation history."""
+    for key, item in donor_dictionary.items():
+        print("{:<20}: {}".format(key, item))
+
+
+def print_report():
+    """Prints the donor table in equally amount of column width with each donor, total amount, number of gifts and average donation."""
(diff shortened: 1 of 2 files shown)


students/tlugosan/lesson03/mailroom.py
```diff
@@ -0,0 +1,81 @@
...
+    'Robin Hood': [1234.56, 4500.34, 765.28],
+    'Gina Travis': [523.10, 75.00],
+    'Mark Johnson': [850.00, 20.14]
+}
+
+
+def select_action(actions):
+    """User selects an action by its corresponding order number."""
+    print('What would you like to do?')
+    for i, item in enumerate(actions):
```
natasha-aleksandrova: you can tell `enumerate` to start at 1: `enumerate(actions, 1)`

students/tlugosan/lesson03/mailroom.py
```diff
@@ -0,0 +1,81 @@
...
+        print("{}) {}".format(i + 1, item))
+    choice_actions = input('Pick your selection by their corresponding number: ')
+    while True:
+        if not choice_actions.isnumeric() or int(choice_actions) not in range(1, len(actions) + 1):
```
natasha-aleksandrova: there are couple different ways to do.. but it will be cleaner when we get to the mailroom version that uses exception handling.

what you have is perfectly fine, another alternative:
`if not '1' < ans < str(len(options))`

students/tlugosan/lesson03/mailroom.py
```diff
@@ -0,0 +1,81 @@
...
+    """Lists all the donors or prompts for a name and donation amount to compile the thank you email."""
+    send_to_name = input("Who do you want to send the email to? Type 'list' for a list of all the donors. ")
+    if send_to_name == 'list':
+        display_donor_dictionary(table_dictionary)
+        send_to_name = input("Who do you want to send the email to? Type 'list' for a list of all the donors. ")
+    if send_to_name not in table_dictionary.keys():
+        table_dictionary[send_to_name] = []
+    donation_amount = float(input("What donation amount do you want to thank them for? "))
+    while True:
+        if type(donation_amount) is not float:
```
natasha-aleksandrova: this validation doesn't make sense because it is always going to be a float via line 36

if you want to ensure it is numeric then this is best done with exception handling which will be later in lesson 5.

students/tlugosan/lesson03/mailroom.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python3
+""" Automate a script to send a thank you mail, print a report of donor or quit the program."""
+
+table_dictionary = {
+    'Toni Orlando': [150.00, 200.00, 100.00],
+    'Amanda Clark': [1800.00],
+    'Robin Hood': [1234.56, 4500.34, 765.28],
+    'Gina Travis': [523.10, 75.00],
+    'Mark Johnson': [850.00, 20.14]
+}
```
natasha-aleksandrova: for the first revision of mailroom this should have been a list (to practice list functionality)... but yes dictionary makes more sense here and will be used in lesson 4.

students/tlugosan/lesson03/mailroom.py
```diff
@@ -0,0 +1,81 @@
...
+def print_report():
+    """Prints the donor table in equally amount of column width with each donor, total amount, number of gifts and average donation."""
+    table_header = ['Donor Name', 'Total Given', 'Num Gifts', 'Average Gifts']
+    len_header = len(table_header)
+    print("|".join(["{:<20}"] * len_header).format(*table_header))
+    print("-" * (20 * len_header + (len_header - 1)))
+    for k, v in table_dictionary.items():
+        sum_donations = 0
+        for x in v:
+            sum_donations = sum_donations + x
```
natasha-aleksandrova: no need to manually sum the values -> `sum(v)` will do that for you 👍

students/tlugosan/lesson03/mailroom.py
```diff
@@ -0,0 +1,81 @@
...
+        print("{:<20}: {}".format(key, item))
+
+
+def print_report():
+    """Prints the donor table in equally amount of column width with each donor, total amount, number of gifts and average donation."""
+    table_header = ['Donor Name', 'Total Given', 'Num Gifts', 'Average Gifts']
+    len_header = len(table_header)
+    print("|".join(["{:<20}"] * len_header).format(*table_header))
+    print("-" * (20 * len_header + (len_header - 1)))
+    for k, v in table_dictionary.items():
```
natasha-aleksandrova: please implement the report sorted functionality per assignment requirements.

> If the user (you) selected “Create a Report”, print a list of your donors, sorted by total historical donation amount.

students/tlugosan/lesson03/mailroom.py
```diff
@@ -0,0 +1,81 @@
...
+    return choice_actions_int
+
+
+def sending_thank_you():
+    """Lists all the donors or prompts for a name and donation amount to compile the thank you email."""
+    send_to_name = input("Who do you want to send the email to? Type 'list' for a list of all the donors. ")
+    if send_to_name == 'list':
+        display_donor_dictionary(table_dictionary)
+        send_to_name = input("Who do you want to send the email to? Type 'list' for a list of all the donors. ")
+    if send_to_name not in table_dictionary.keys():
```
natasha-aleksandrova: note that calling `.keys()` isn't required; it will loop over keys by default.

PR #860, opened 2018-08-06
Submitting Mailroom Pt. 2

mxwllndrsn:
Didn't see a reason to use dictionary for donor date structure, but worked very well for function dispatch.

### DIFF
--- a/students/mxwllndrsn/lesson04/mailroom_mk2.py
+++ b/students/mxwllndrsn/lesson04/mailroom_mk2.py
@@ -1,0 +1,142 @@
+#!/usr/bin/env python3
+
+#uw python 210
+#lesson 03
+#max anderson
+
+#mailroom.py
+
+
+donor_list = [
+              ['Andrew Jackson', 10000, 5, 2000],
+              ['Benjamin Franklin', 253000, 2, 117500],
+              ['John Kennedy', 20, 1, 20],
+              ['Abraham Lincoln', 807, 2, 403.5],
+              ['Jim Jones', 918, 1, 918]
+             ]
+
+# Index 0 - Donor Name
+# Index 1 - Donation Totals
+# Index 2 - Number of Donations
+# Index 3 - Average Donation
+
+
+#implementing dictionary function menu
+def menu_select(prompt, menu):
+    while(True):
+        select = input(prompt)
+        while(select not in menu.keys()):
+            select = input('Please try again: ')
+        if menu[select]() == 'exit':
+            break
+    print('Goodbye.')
+
+
+def exit_menu():
+    return 'exit'
+
+
+def thank_you():
+    print()
+    name = input('Enter full name of \'Thank You\' recipient: ').title()
+
+    while(name == 'List'):
+        list_donors()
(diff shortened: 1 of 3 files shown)


students/mxwllndrsn/lesson04/mailroom_mk2.py
```diff
@@ -0,0 +1,186 @@
...
+import io
+from datetime import date
+
+donor_list = [
+              ['Andrew Jackson', 10000, 5, 2000],
+              ['Benjamin Franklin', 253000, 2, 117500],
+              ['John Kennedy', 20, 1, 20],
+              ['Abraham Lincoln', 807, 2, 403.5],
+              ['Jim Jones', 918, 1, 918]
+             ]
```
natasha-aleksandrova: from lesson 4 and forward this should be a dictionary. why use a dictionary? because it is VERY efficient and simplifies the code in _this_ particular case.

your donor name is a natural key in our DB; so `donors = {"joe": [100, 200], "tom": [44]}`

which means looking up a donor:
`donor = donors[donor_name]`
and adding to donor
`donor[donor_name.append(donation)`
becomes simple + efficient

.. and because you get to practice dictionaries because they are honestly one of the most common and useful data structures in python 😉

students/mxwllndrsn/lesson04/mailroom_mk2.py
```diff
@@ -0,0 +1,186 @@
...
+# Index 2 - Number of Donations
+# Index 3 - Average Donation
+
+ymdate = str(date.today())
+
+#implementing dictionary function menu
+def menu_select(prompt, menu):
+    while(True):
+        select = input(prompt)
+        while(select not in menu.keys()):
```
natasha-aleksandrova: simplify to `while select not in menu:`

students/mxwllndrsn/lesson04/mailroom_mk2.py
```diff
@@ -0,0 +1,186 @@
...
+
+def enter_donation():
+    print()
+    name = input('Enter full name of \'Donor\': ').title()
+    if in_list(name): donation_update(name)
+    elif not in_list(name): donation_update(new_donor(name))
+
+    letter = input('Create thank you letter? (y/n) ').title()
+    if letter == 'Y': print(format_thx(name))
+    elif letter == 'N': return
```
natasha-aleksandrova: although it is valid, it is discouraged and will be flagged by a linter when using inline if/else statements.

students/mxwllndrsn/lesson04/mailroom_mk2.py
```diff
@@ -0,0 +1,186 @@
...
+    prntstr += ('Sincerely, \n\n')
+    prntstr += ('Automated Form letter')
+
+    return prntstr
+
+def create_report():
+    print()
+    print('Donor Name             | Donation Total   | Num   | Average')
+    print('-----------------------------------------------------------')
+    for i in range(len(donor_list)):
```
natasha-aleksandrova: this report needs to be sorted per original requirements.

PR #861, opened 2018-08-06
Submitting assignment 4

rfelts:
Couldn't figure out how to use variables to represent dictionary keys in string format statements. For example: print(("{key:<{name_padding}}  ${value:>{donation_padding},.2f}") .format(**my_dict)).


natasha-aleksandrova:
here is an example:
```
>>> my_dict = {"name": "natasha", "value": 10}
>>> print(("{name:<20s} ${value:>11.2f}") .format(**my_dict))
natasha              $      10.00
>>>
```

PR #864, opened 2018-08-07
Lesson 09 object oriented mailroom and test

smitco:
Completed lesson 09 exercise to refactor mailroom file using OO programming methods and test appropriately

### DIFF
--- a/students/smitco/lesson09/mailroomOO.py
+++ b/students/smitco/lesson09/mailroomOO.py
@@ -1,0 +1,147 @@
+# lesson 09 mailroom object oriented
+# !/usr/bin/env python3
+
+
+import os
+import datetime
+
+donors = {"John Travolta": [5000, 7500],
+          "Jane Fonda": [10000, 8000, 6500],
+          "Judy Blume": [3000, 3000],
+          "Joey Tribbiani": [9000],
+          "Jenny Gump": [10300, 13750, 12500]}
+
+class Donor(object):
+    def __init__(self, name, *args, **kwargs):
+       self.name = name
+    
+    def donor_name(self, name):
+        if self.name.lower() == "list":
+            print(list(donors.keys()))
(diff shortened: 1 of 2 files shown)


students/smitco/lesson09/mailroomOO.py
```diff
@@ -0,0 +1,147 @@
...
+                self.amount = "exit"
+            elif int(self.amount) >= 10**6:
+                print("\nThe amount entered is too large.")
+                self.amount = "large"
+            else:
+                self.amount = int(self.amount)
+        except ValueError:
+            print("\nInvalid entry.")
+            self.amount = "invalid"
+        return self.amount
```
natasha-aleksandrova: the `donor_name` and `donor_amount` don't really belong in the Data Class. It is also unclear what function they serve? This code is probably best to be part of main program flow.

students/smitco/lesson09/mailroomOO.py
```diff
@@ -0,0 +1,147 @@
...
+        if self.name in donors:
+            donors[self.name].append(self.amount)
+        else:
+            donors[self.name] = [self.amount]
+
+    def donor_email(self, name, amount):
+        print("\nThank you, {}, for your generous donation of ${} "
+              "to the Brave Heart Foundation.".format(self.name, self.amount))
+
+class All_Donors(Donor):
```
natasha-aleksandrova: class naming convention is `UpperCamelCase`

natasha-aleksandrova: AllDonors should _not_ inherit from Donor; inheritance relationship is more like Student (child class) *is* Person (parent class).

The relationship between `Donor` and `AllDonors` is one to many; meaning `AllDonors` should have a reference to collection of donor objects.

Since you're still using global dictionary to hold data, you're not really using the classes to hold any data asked in the assignment.

PR #866, opened 2018-08-07
Lesson 09 OO mailroom and test file reworked

smitco:
Reworked the OO mailroom assignment and associated test file

The suggestions section mentioned a few functionalities not previously used like searching for a donor or saving the updated list. Should those be added into this file, or do I just need to have the code structured to add those in the future?

### DIFF
--- a/students/smitco/lesson09/mailroomOO.py
+++ b/students/smitco/lesson09/mailroomOO.py
@@ -5,58 +5,39 @@
 import os
 import datetime
 
-donors = {"John Travolta": [5000, 7500],
-          "Jane Fonda": [10000, 8000, 6500],
-          "Judy Blume": [3000, 3000],
-          "Joey Tribbiani": [9000],
-          "Jenny Gump": [10300, 13750, 12500]}
 
 class Donor(object):
     def __init__(self, name, *args, **kwargs):
-       self.name = name
-    
-    def donor_name(self, name):
-        if self.name.lower() == "list":
-            print(list(donors.keys()))
-        elif self.name.lower() == "exit":
-            print("\nExiting.\n")
-        return self.name
-    
-    def donor_amount(self, name, amount):
-        self.amount = amount
-        try:
-            if self.amount.lower() == "exit":
-                print("\nExiting.")
-                self.amount = "exit"
-            elif int(self.amount) >= 10**6:
-                print("\nThe amount entered is too large.")
-                self.amount = "large"
-            else:
-                self.amount = int(self.amount)
-        except ValueError:
-            print("\nInvalid entry.")
-            self.amount = "invalid"
-        return self.amount
-
-    def donor_append(self, name, amount):
-        if self.name in donors:
-            donors[self.name].append(self.amount)
-        else:
-            donors[self.name] = [self.amount]
+        self.name = name
 
     def donor_email(self, name, amount):
+        self.amount = amount
         print("\nThank you, {}, for your generous donation of ${} "
               "to the Brave Heart Foundation.".format(self.name, self.amount))
 
-class All_Donors(Donor):
-    def __init__(self, name, amount, *args, **kwargs):
-        Donor.__init__(self, name, amount, *args, **kwargs)
+class AllDonors(object):
+    donors = {}
     
+    def __init__(self, *args, **kwargs):
+        self.donors["John Travolta"] = [5000,7500]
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
you are referring to this line?

> This will hold all the donor objects, but also methods to add a new donor, search for a given donor, etc. If you want a way to save and re-load your data, this class would have that too.

by "search donor" it means to look up a donor when you need to add donations or add a new one
by "save and re-load your data" it means to store information in your object instance and display that same data

it's just using fancier terms that are more common in the industry 🙂

smitco:
Okay thanks! I was thinking "search donor" meant a user-initiated search to
see if a particular person had made a donation, and "save and re-load data"
meant to be able to save the donors and amounts added in a particular
session to be able to see that info the next time the program is loaded.

natasha-aleksandrova:
a little sneak peak.. in next class you will actually persist data in your mailroom program to various databases 😉

students/smitco/lesson09/mailroomOO.py
```diff
@@ -5,65 +5,47 @@
 import os
 import datetime
 
-donors = {"John Travolta": [5000, 7500],
-          "Jane Fonda": [10000, 8000, 6500],
-          "Judy Blume": [3000, 3000],
-          "Joey Tribbiani": [9000],
-          "Jenny Gump": [10300, 13750, 12500]}
 
 class Donor(object):
```
natasha-aleksandrova: your Donor class doesn't seem to serve a lot of purpose right now.

Donor class should hold everything about the single Donor and be integrated with AllDonors; here is some example code:
``` 
class Donor:
    def __init__(self, name, donations):
        self.name = name  # ONLY donor name, no "list"
        self.donations = donations

    @property
    def total(self):
        return sum(self.donations)  # example

    @property
    def average(self):
        pass

    @property
    def num_donations(self):
        pass

    def add(self, donation):
        self.donations.append(donation)

    def get_letter_text(self):
        return f"Letter stuff for {self.name}"


class Donors:
    def __init__(self, donors=None):
        self.donors = donors or {}  # initialize
        # the data structure here is a dictionary with donor name as a key and donor object reference as a value

    @classmethod
    def initialize(cls, *donor_objs):  # getting really fancy here
        return cls(donors={d.name: d for d in donor_objs})

    def add_donation(self, donor_name, donation):
        if donor_name in self.donors:
            self.donors[donor_name].add(donation)
        
        else:
            self.donors[donor_name] = Donor(donor_name, [donation])
    
    def get_report(self):
        """Return report as a list of rows (could also return a string)"""
        rows = []
        for donor in self.donors.values():
            rows.append(f"{donor.name} {donor.total}")
        return rows
    
    def send_letters(self):
        for donor in self.donors.values():
            print(donor.get_letter_text())

    
```
example usage:
```
donors = Donors.initialize(Donor("Natasha", [100]), Donor("Tom", [999]))
print(donors.get_report())
```

students/smitco/lesson09/mailroomOO.py
```diff
@@ -5,65 +5,47 @@
...
     def __init__(self, name, *args, **kwargs):
```
natasha-aleksandrova: you're not using args or kwargs

students/smitco/lesson09/mailroomOO.py
```diff
@@ -5,65 +5,47 @@
...
-
-    def donor_append(self, name, amount):
-        if self.name in donors:
-            donors[self.name].append(self.amount)
-        else:
-            donors[self.name] = [self.amount]
+        self.name = name
 
     def donor_email(self, name, amount):
+        self.amount = amount
```
natasha-aleksandrova: you should not define any class attributes (self.*) outside of `__init__` method.

it's either defined to belong to that class in the constructor or should be a local var.

students/smitco/lesson09/mailroomOO.py
```diff
@@ -113,28 +96,33 @@ def thank_you():
...
 def report():
-    all_donors = All_Donors("None", 0)
-    all_donors.donor_details()
+    all_donors = AllDonors()
     all_donors.donor_report()
 
 def letters():
-    all_donors = All_Donors("None", 0)
-    all_donors.donor_details()
+    all_donors = AllDonors()
```
natasha-aleksandrova: your class should only need to be instantiated once, at the start of your program.

PR #869, opened 2018-08-09
Mailroom3.py

toddlikesbeer:
I'm really starting to like try\except statements; cool stuff. 

When I tried the comprehension on line 117, I ran into 
https://stackoverflow.com/questions/16641119/why-does-append-return-none-in-this-code. I'm going to do some investigation on why appending was returning None, but it works with a print statement. Decided to add that to my list of things to investigate, but call it done and move on. Any hints or suggestions welcome.

### DIFF
--- a/students/toddschultz/session05/mailroom3.py
+++ b/students/toddschultz/session05/mailroom3.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+
+global donors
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
`donors_list = [print(i[0].title()) for i in donors if type(i[0]) is str]`

when you create a list comprehension it generates a list with values on the left side of statement, in your case its your print() statement. since print returns None as a function, you are getting a bunch of None values. 

Also, as mentioned in previous lesson - main donors structure should be a dictionary not a list.

PR #879, opened 2018-08-12
Add assignment 7

cosmonautz:


students/KevinWaldron/lesson07/html_render.py
```diff
@@ -0,0 +1,134 @@
...
+    An html element consists of an open tag potentially containing attributes followed by content and finally a closing tag
+
+    """
+
+    tag = 'html'
+    indent = 4
+
+    def __init__(self, content=None, **kwargs):
+        self.content = []
+        if content is not None:
```
charlesrobison: This can simply be `if content:`

students/KevinWaldron/lesson07/html_render.py
```diff
@@ -0,0 +1,134 @@
...
+        file_out.write(f"</{self.tag}>")
+
+class Title(OneLineTag):
+    """ A title tag """
+    tag = 'title'
+
+class SelfClosingTag(Element):
+    """ A self closing html tag, requires no closing tag and supports no content """
+
+    def __init__(self, **kwargs):
```
charlesrobison: This __init__ shouldn't be necessary as inheriting from Element.

students/KevinWaldron/lesson07/html_render.py
```diff
@@ -0,0 +1,134 @@
...
+    """ A line break tag """
+    tag = 'br'
+
+class A(OneLineTag):
+    """ An anchor tag """
+    tag = 'a'
+
+    def __init__(self, link, content=None):
+        link_ref = {"href":link}
+        Element.__init__(self, content, **link_ref)
```
charlesrobison: Typically want to use **kwargs here, instead of **link_ref - standard convention.

PR #880, opened 2018-08-12
Lesson 7 assignment from Cindy Waldron

cindywaldron:


students/cindywaldron/lesson7/html_render.py
```diff
@@ -0,0 +1,126 @@
...
+    """Head subclass of Element"""
+    tag = "head"
+
+class OneLineTag(Element):
+    """subclass of Element, tags + content are in one line"""
+    def __init__(self, content, **kwargs):
+        self.a_list = []
+        if content is not None:
+            self.a_list.append(content)
+        self.kwargs = kwargs
```
natasha-aleksandrova: no need to do this; the parent constructor is doing the same thing.

students/cindywaldron/lesson7/html_render.py
```diff
@@ -0,0 +1,126 @@
...
+class Title(OneLineTag):
+    """Title subclass of OneLineTag"""
+    tag = "title"
+
+class SelfClosingTag(Element):
+    """SelfClosingTag subclass of Element"""
+    tag = ''
+
+    def __init__(self, **kwargs):
+        self.kwargs = kwargs
```
natasha-aleksandrova: the constructor should be the same as Element

PR #881, opened 2018-08-12
mailroom added

au-gold:


students/Mintae_K/lesson03/mailroom.py
```diff
@@ -0,0 +1,60 @@
...
+    # menu_prompt
+    while not (response == '3'):
+        print('Menu')
+        print('1. Send a Thank You')
+        print('2. Create a Report')
+        print('3. Quit')
+        response = input("Please type number from menu > ")
+        if response == '1':
+            donors = sendingthank(donors)
+        if response == '2':
```
natasha-aleksandrova: This should be an if/elif not if/if since it can only be one of those options.

students/Mintae_K/lesson03/mailroom.py
```diff
@@ -0,0 +1,60 @@
...
+    s = '\n, {:<20}| {:<10} | {:<10}| {:<20}'
+    print(s.format('Donor Name', 'Total Given', "Num Gift", "Average Gift"))
+    print('-' * 60)
+    for x in range(len(donors)):
+        total = sum(donors[x][1:])
+        num_gift = len(donors[x]) - 1
+        s = '{:<20}  $ {:<10.2f}  {:^12} $ {:<20.2f}'
+        print(s.format(donors[x][0], total, num_gift, total / num_gift))
+    print(' ')
+    return donors
```
natasha-aleksandrova: You don't need to return the donors here

students/Mintae_K/lesson03/mailroom.py
```diff
@@ -0,0 +1,60 @@
...
+    if name_exists is False:
+        s = 'New donor, please type the donation amount from {}> '
+        amount = input(s.format(f_name))
+        donors = donors + [[f_name, float(amount)]]
+    else:
+        s = 'Existing donor, please type the donation amount from {}> '
+        amount = input(s.format(f_name))
+        donors[x] = donors[x] + [float(amount)]
+    print('\n', "Thank you {} for the generous donation!".format(f_name), '\n')
+    return donors
```
natasha-aleksandrova: No need to return donors here

students/Mintae_K/lesson03/mailroom.py
```diff
@@ -0,0 +1,60 @@
+# Sending a Thank you
+def sendingthank(donors):
+    f_name = input('Please type the Full Name or list> ')
+    while f_name == 'list':
+        print(' ')
+        for x in donors:
+            print(x[0])
+        print(' ')
+        f_name = input('Please type the Full Name or list> ')
+    for x in range(len(donors)):
```
natasha-aleksandrova: Loop over the list to get each donor:
```
for donor in donors:
    donor_name = donor[0]
    #etc
```

students/Mintae_K/lesson03/mailroom.py
```diff
@@ -0,0 +1,60 @@
...
+# Creating Report
+def createreport(donors):
+    s = '\n, {:<20}| {:<10} | {:<10}| {:<20}'
+    print(s.format('Donor Name', 'Total Given', "Num Gift", "Average Gift"))
+    print('-' * 60)
+    for x in range(len(donors)):
+        total = sum(donors[x][1:])
+        num_gift = len(donors[x]) - 1
+        s = '{:<20}  $ {:<10.2f}  {:^12} $ {:<20.2f}'
+        print(s.format(donors[x][0], total, num_gift, total / num_gift))
```
natasha-aleksandrova: This report should be sorted (check assignment requirements), please add sorting in the next revision.

natasha-aleksandrova:
Overall looks good - could use few improvements and please add missing sorting feature.

students/Mintae_K/lesson03/mailroom.py
au-gold: I think I need to return donors because this function updates and adds new donors. If not returned, the main() will not have the latest donor list.

natasha-aleksandrova: in this particular assignment the main data structure `donors` is defined at the module level, at the top of the file.

that way it is a global data structure - it simplifies the code a lot.

PR #883, opened 2018-08-13
Lesson10 mailroom_FP.py, functional programming

wpucilow:


natasha-aleksandrova:
Nice work on new features. Remember that there should be no `input` functions in your data classes:  `Donor` and `DonorCollection`. Those classes are for data manipulation not to control user flow and interaction.

PR #884, opened 2018-08-13
Lesson 10 mailroom functional programming and test file

smitco:
Used map, filter and sum/reduce to add matching contribution function to the exercise
Based this assignment off mailroom4 file


natasha-aleksandrova:
Nice work on completing the last assignment!

PR #888, opened 2018-08-14
Lesson 3 first exercise

ltrisolini:
Here is the first exercise for Lesson 3, thanks!

### DIFF
--- a/students/ltrisolini/Lesson03/list_lab.py
+++ b/students/ltrisolini/Lesson03/list_lab.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python3
+#List Lab Exercise
+
+
+#Series1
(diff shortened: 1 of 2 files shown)


students/ltrisolini/Lesson03/list_lab.py
```diff
@@ -0,0 +1,87 @@
...
+my_list = ['Cherry'] + my_list
+
+print(my_list)
+
+my_list.insert(0, 'Kiwi')
+
+print(my_list)
+
+for item in my_list:
+    if item[0] == 'P':
```
charlesrobison: Good idea to add 'lower' to catch lower case 'p' input:  if item[0].lower() == 'p':

PR #895, opened 2018-08-15
Submitting lesson3 mailroom exercise

ptfbanks:
I to admit, I needed to consult a variety of other students work to get started on this one.   It was revealing to see the variety of methods employed to meet the assigned specifications.  Much in the assignment required research well beyond the materials covered to date.

### DIFF
--- a/students/ptfbanks/Lesson03/pf_mailroom.py
+++ b/students/ptfbanks/Lesson03/pf_mailroom.py
@@ -1,0 +1,93 @@
+#  MailRoom exercise lesson03 final task
+#    - Build, sort, print a table
+#--------------------------Global Region-----------------------------------#
+donations = [
+    ("Bob Williams", 5580.00),
+    ("Tom Haskell", 345.05),
+    ("Earl Jackson", 4553.45),
+    ("Les Thomas", 8934.00),
+    ("James Black", 154),
+    ("Bob Williams", 3245.50),
+    ("Tom Haskell", 3245.90),
+    ("Earl Jackson", 45.6),
+    ("Les Thomas", 325.00),
+    ("James Black", 1300),
+    ("Bob Williams", 1000),
+    ("Tom Haskell", 243.05),
+    ("Earl Jackson", 345.00),
+    ("Les Thomas", 3665.00),
+    ("James Black", 21)]
+
+#-----------------------------------Define Functions----------------------#
+#Prepare list of donors
+def get_names(list):
+    names = []
+    for entry in (list):
+        name = entry[0]
+        if name not in names:
+            names.append(name)
+    return(names)
+            
+def select_Key(table):
+    return table[1]
+
+#Create history report
+def hist_rpt(dlist):
+    names=get_names(donations)  #refresh names of donors
+    print("|   DonorName    | TotaL Given | Num Gifts   | Avrage Gift |")
+    print("|----------------|-------------|-------------|-------------|")
+    rpt=[]      #list for report rows
+    tc = 0      #set tuple member ID
+    for name in (names):
+        dsum =0      #set donor sum
+        dc = 0       #set donations count
+        tup=[]      #collector for row content
+        for entry in (dlist):    #search donations ist
+            donor = entry[0]
+            if donor == name:
+                dc=dc+1
+                dval = entry[1]
+                dsum = dsum + dval
+        davg = dsum/dc         #calculate donor average donation
+        rpt.append((name, dsum, dc, davg))  #create table entry
+    rpt.sort(key=select_Key, reverse=True)   #sort descending
+    for entry in rpt:
+        print("|{:^16}|{:>13,.02f}|{:^13}|{:>13,.02f}|".format(*entry))
+        print("|----------------|-------------|-------------|-------------|")
+
+def send_thanks(whom, amt):
+    print(f'\n Dear {whom} \n') 
+    print(f'We are pleased with your generous gift of ${amt}.\n')
+    print(f'Take comfort in knowing this will be put to good use.\n\n')
+    print('Sincerely,   The Good Group')
+
+#------------------------Mailroom Body of Logic--------------------------------#
+#------------------User Interface--------------------------#
+print("You have entered the Donation Management screen.")
+stp= "c"
+while stp != "q":
+    stp = input("Do you want to continue? (Enter'c'-continue or 'q'-quit:")
+    if stp == "q":
+        break
+    don_new = input ("Are you adding a new donation?(y/n)")             
+    if don_new == "y":
+        n_entry=[]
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
Thanks for the write up.

I will try to address your questions as part of this PR as well when I review it.

students/ptfbanks/Lesson03/pf_mailroom.py
```diff
@@ -0,0 +1,93 @@
...
+    need_rpt = input("\n Will you be needing a donation history report? (y/n)")
+    if need_rpt == "y":
+        print(hist_rpt(donations))
+
+    need_note = input("\n Do you need to generate a thank you note?  (y/n)")
+    if need_note == "y":
+        whom = input("\n Name of donor:")
+        donation = input("Donated amount: ")
+        print("The following is placed in queue for printing or texting:")
+        print("\n", send_thanks(whom, donation))
```
natasha-aleksandrova: per assignment guidelines:
`Put your main interaction into an if __name__ == '__main__' block.`

it is critical as it will cause issues in later assignments.

students/ptfbanks/Lesson03/pf_mailroom.py
```diff
@@ -0,0 +1,93 @@
...
+        tup=[]      #collector for row content
+        for entry in (dlist):    #search donations ist
+            donor = entry[0]
+            if donor == name:
+                dc=dc+1
+                dval = entry[1]
+                dsum = dsum + dval
+        davg = dsum/dc         #calculate donor average donation
+        rpt.append((name, dsum, dc, davg))  #create table entry
+    rpt.sort(key=select_Key, reverse=True)   #sort descending
```
natasha-aleksandrova: sorting is mentioned here (which is required reading in lesson3):
https://uwpce-pythoncert.github.io/PythonCertDevel/modules/Sequences.html#custom-sorting

for any python built-ins I highly recommend getting familiar with searching official docs:
https://docs.python.org/3.3/howto/sorting.html
https://docs.python.org/3.3/library/stdtypes.html#list.sort

and yes your implementation here is correct.

students/ptfbanks/Lesson03/pf_mailroom.py
```diff
@@ -0,0 +1,93 @@
...
+#Prepare list of donors
+def get_names(list):
+    names = []
+    for entry in (list):
+        name = entry[0]
+        if name not in names:
+            names.append(name)
+    return(names)
+            
+def select_Key(table):
```
natasha-aleksandrova: remember to use snake_case -> `select_key`

a better input var name would be an `item` since you are getting here a single item from your list.

students/ptfbanks/Lesson03/pf_mailroom.py
```diff
@@ -0,0 +1,93 @@
...
+    ("Bob Williams", 3245.50),
+    ("Tom Haskell", 3245.90),
+    ("Earl Jackson", 45.6),
+    ("Les Thomas", 325.00),
+    ("James Black", 1300),
+    ("Bob Williams", 1000),
+    ("Tom Haskell", 243.05),
+    ("Earl Jackson", 345.00),
+    ("Les Thomas", 3665.00),
+    ("James Black", 21)]
```
natasha-aleksandrova: in real applications choosing the right data structure is critical - and as you are learning it may be a trial and error.
it's important to know all available basic data structures and then use one or combine them.

for example a list can contain pretty much anything and get really nested:
`some_list = [1, (1, 2, 4), {"key1": "value1"}, Donor(), ['some-strings']]`

that one contains: integer, tuple, dictionary, object and a list - there is really no limit.

For the mailroom program you get to decide what makes the most sense, keeping in mind that you do want to keep your code simple and not over-complicate it. 

I would suggest following data structure where donations are flat and you have 1 unique donor in the list:
```
donors = [("james smith", [100, 200, 300]), ("les thomas", [343, 33])]
```
it should really simplify your code.

students/ptfbanks/Lesson03/pf_mailroom.py
```diff
@@ -0,0 +1,93 @@
...
+
+#-----------------------------------Define Functions----------------------#
+#Prepare list of donors
+def get_names(list):
+    names = []
+    for entry in (list):
+        name = entry[0]
+        if name not in names:
+            names.append(name)
+    return(names)
```
natasha-aleksandrova: no need for parens they are redundant -> `return names`

students/ptfbanks/Lesson03/pf_mailroom.py
```diff
@@ -0,0 +1,93 @@
...
+    ("Tom Haskell", 243.05),
+    ("Earl Jackson", 345.00),
+    ("Les Thomas", 3665.00),
+    ("James Black", 21)]
+
+#-----------------------------------Define Functions----------------------#
+#Prepare list of donors
+def get_names(list):
+    names = []
+    for entry in (list):
```
natasha-aleksandrova: avoid using `list` for var names, any built-in that you are overwriting can potentially break your code and have spend lots of time trying to figure what is the issue.

students/ptfbanks/Lesson03/pf_mailroom.py
```diff
@@ -0,0 +1,93 @@
...
+        name = entry[0]
+        if name not in names:
+            names.append(name)
+    return(names)
+            
+def select_Key(table):
+    return table[1]
+
+#Create history report
+def hist_rpt(dlist):
```
natasha-aleksandrova: this looks like a C++ convention 🙂 

it is preferred to spell out words (up to a reason of course) but this would be better -> `history_report`

students/ptfbanks/Lesson03/pf_mailroom.py
natasha-aleksandrova: alright let's clean up the entry point here 🙂 

```
prompt_text = "1 - Send Thank you\n 2 - Create Report\n 3 - Quit" # i'm really over simplifying it here

def main():
    while True:
        response = input(promt_text)
        if response == "1":
            send_thank_you()
        elif response == "2":
            create_report()
        elif response == "3":
            quit_program()
        else:
            print("invalid")

if __name__ == '__main__':
    main()
```
above is very clean and readable - this will be even more important for lesson 4.

also, no need to pass in donors main data structure into your functions - it is in the global space

natasha-aleksandrova:
`How to identify a position in an embedded list is still confusing.`

basically what you did in your code, loop over the list and match on an item. this operation can be very expensive.

its all case by case and if you can, choose a better data structure that is more efficient but sometimes that is exactly what you get - for example from a database query - and you do have to loop over all of the records in the list to get what you're looking for.

PR #900, opened 2018-08-16
Mailroom_pt3 submission

tfbanks:
submitting mailroom_pt3

### DIFF
--- a/students/tfbanks/Lesson05/Mailroom_pt3.py
+++ b/students/tfbanks/Lesson05/Mailroom_pt3.py
@@ -1,0 +1,89 @@
+# Mailroom_pt3.py Exercise by tfbanks
+
+# !/usr/bin/env python3
+
+
+# Donor Database containing names and donations
+donor_names = {
+    'Tim Cooker': [1500.50, 25050.50, 15680.75],
+    'Elon Musket': [5250.25, 26500],
+    'Frank Petersmankempt': [550.60],
+    'Megan Morgan': [650.40, 1600, 20625.40],
+    'Marlene Wheeler': [820, 1222.80]}
+
+
+def thank_you():  # Code for selecting Donor and writing a Thank You Note
+    ty_donor = input('Enter the Full Name of a donor (for list of previous donors type list): ').title()
+    letter = ('''\n\nDear {},\n
+Thank you for your generous donation of ${:,.2f}, your generosity is greatly appreciated.
+These funds will help continue our efforts to teach Python to the next generation.\n
+Sincerely,\n\n
+Joesef Edword Bringingham\n\n''')
+    if ty_donor == 'List':  # If List, Prints Donor List and restarts the function
+        for donor in donor_names.keys():
+            print(donor)
+        print('\n')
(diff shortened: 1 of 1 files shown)


students/tfbanks/Lesson05/Mailroom_pt3.py
```diff
@@ -0,0 +1,89 @@
...
+            print('''Enter the actions's associated number''')
+        if answer == 1:
+            thank_you()
+        elif answer == 2:
+            report()
+        elif answer == 3:
+            mass_mail()
+        elif answer == 4:
+            print('Have a nice day')
+            break
```
natasha-aleksandrova: Why was this changed back to previous versions instead of building on top of it? This used to be a switch dict style in a previous submission. I would recommend changing it back.

students/tfbanks/Lesson05/Mailroom_pt3.py
```diff
@@ -0,0 +1,89 @@
...
+    else:
+        donor_names[ty_donor] = [amount]
+
+    print(letter.format(ty_donor, amount))
+
+
+def report():  # Defines parameters of the report of donors, creates a new list of summarized details per donor, sorts, and prints it.
+    print('\nDonor Name            |  Total Given | Num Gifts |  Average Gift')
+    print('------------------------------------------------------------------')
+    donor_rpt = [[donor, sum(donations), len(donations), sum(donations)/len(donations)] for donor, donations in donor_names.items()]
```
natasha-aleksandrova: Usually you want to avoid one letter vars, or short vars but in the case of comprehensions it's ok.

Very common pattern is to use `for k, v in some_dict.items()` or `for key, val in some_dict.items()`
so that the comprehension line is not overly long.

students/tfbanks/Lesson05/Mailroom_pt3.py
```diff
@@ -0,0 +1,89 @@
...
+        print('\n')
+        ty_donor = input('Enter the Full Name of a donor (for list of previous donors type list): ').title()
+
+    while True:
+        try:
+            amount = float(input('Please enter the donation amount: '))
+        except ValueError:
+            print('Please enter a valid donation amount')
+            continue
+        break
```
natasha-aleksandrova: you can remove the `continue` and add `break` into `else` line like so:
```
try:
    # try something
except ValueError:
    print('error')
else:
    break
```

tfbanks:
In one of the comments made with an earlier submittal made me rethink how that switch dict code was working and playing around with it with the try except and removing all the call backs at the end of each selection I thought this was a good option, i will switch it back with the lesson six exercise.

natasha-aleksandrova:
I reviewed the comments and not sure which one that was, apologies if there was confusion.

Let me know if there are other question on that. Switch dict pattern is still a good option here so I would stick with it (reduces boiler plate code).

tfbanks:
I was looking at the comment about “might as well call the function here too…” 

I saw the select (answer) and then the comment after your suggested while true statement, and I thought you were referring to it being ok to use select(answer) and then thought about  just calling the functions there by their selection is what you were referring to because that also made sense to me.

 

That’s ok though,  learned a lot in the process

PR #903, opened 2018-08-19
Assignment 6 updates

Swt16:
Assignment 6 updates

### DIFF
--- a/students/ShinTran/lesson06/TestMailRoom4.py
+++ b/students/ShinTran/lesson06/TestMailRoom4.py
@@ -1,0 +1,54 @@
+'''
+Shin Tran
+Python 210
+Lesson 6 Assignment
+'''
+
+import unittest
+import mailroom4
+from io import StringIO
+
+class TestMailRoom4(unittest.TestCase):
+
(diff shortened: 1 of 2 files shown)


students/ShinTran/lesson06/TestMailRoom4.py
```diff
@@ -0,0 +1,54 @@
...
+    def test_file_output(self):
+        donor_list = [["Jennifer Miller",145076.19],["William Rodriguez", 161585.16],["Patricia Brown",100863.96]]
+        message = "Dear {:s},\n\
+    Thank you for donating ${:,.2f}.\n\
+    Sincerely,\n\
+    Your Local Charity"
+        comp_dict = {}
+        for item in donor_list:
+            comp_dict[item[0]] = message.format(*item)
+        for k, v in comp_dict.items():
```
natasha-aleksandrova: you can also add a check after this line as:
`assert os.path.isfile(expected_filename)`

this will fail early if file does not exist at all


there are two types of test failures: FAILURE and ERROR - FAILURE is when assert fails and ERROR is when the test throws an exception. Ideally you want FAILURE type since you it fails because of your test condition.

PR #905, opened 2018-08-21
Shane Repking Added Lesson 2 Grid Assignment

srepking:
Hi, 
I added the Lesson 2 Grid Assignment for review. Please look at Grid.py

Thank you!
Shane

### DIFF
--- a/students/srepking/Lesson2/FizzBuzz.py
+++ b/students/srepking/Lesson2/FizzBuzz.py
@@ -1,0 +1,11 @@
+def fizzbuzz():
+    for i in range(1,101):
+        if i%3==False and i%5==False:
+            print('FizzBuzz')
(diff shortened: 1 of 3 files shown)


students/srepking/Lesson2/FizzBuzz.py
```diff
@@ -0,0 +1,11 @@
+def fizzbuzz():
+    for i in range(1,101):
+        if i%3==False and i%5==False:
+            print('FizzBuzz')
```
natasha-aleksandrova: when comparing to False and None you should use `is` to ensure it compares as an object:

```
>>> 3 % 3
0
>>> 3 % 3 is False
False
>>> 3 % 3 == False
True
>>>
```
so it should be:
`if i % 3 == 0`

PR #908, opened 2018-08-21
Lesson06: cigar and mailroom unit_test files

tlugosan:
IS it possible to test only in the intermediary tests like the following? And what would be the SOMETHING. I was thinking to try to read the input messages that are printed but I failed. Would that be a correct approach? 
def test_tanks_existing_donor(capsys):
    test_input = ['1', 'list']
    with mock.patch('builtins.input', side_effect=test_input):
        mailroom.select_action_dictionary('1', mailroom.switch_func_dict)
        captured = capsys.readouterr()
        out = captured.out.strip()
        err = captured.err.strip()
        assert SOMETHING

### DIFF
--- a/students/tlugosan/lesson06/test_mailroom4.py
+++ b/students/tlugosan/lesson06/test_mailroom4.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python3
+
+import sys
+from io import StringIO
+import mailroom4 as mailroom
+import unittest
+from unittest import mock
+from mock import patch
+import pdb
+from operator import itemgetter
+import os
+
+
+def test_select_action_dictionary_exit(capsys):
+    test_input = ['4']
+    with mock.patch('builtins.input', side_effect=test_input):
+        mailroom.select_action_dictionary('4', mailroom.switch_func_dict)
+        captured = capsys.readouterr()
+        out = captured.out.strip()
+        assert out == "Before quitting."
+
+
+def test_select_action_dictionary_lower_bound_invalid(capsys):
+    test_input = ['0', '4']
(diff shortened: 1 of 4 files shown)


natasha-aleksandrova:
@tlugosan please use code formatting, it's hard to read code not formatted 🙂 

https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code

natasha-aleksandrova:
@tlugosan yes using capsys and mocking `input` would be the approach to test code like that.

you aren't expected to do (they are advanced features), the idea is to refactor your code to separate user interaction code (`input`, `print` functions) and only unit test "business logic" code.

students/tlugosan/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,121 @@
...
+    result_list = mailroom.print_report()
+    captured = capsys.readouterr()
+    out = captured.out.strip()
+    test_list = []
+    for k in mailroom.table_dictionary:
+        sum_donations = sum(mailroom.table_dictionary[k])
+        total_gifts = len(mailroom.table_dictionary[k])
+        average_gift = sum_donations / total_gifts
+        test_list.append([k, sum_donations, total_gifts, average_gift])
+        test_sorted_new_list = sorted(test_list, key=itemgetter(1), reverse=True)
```
natasha-aleksandrova: This resembles your actual mailroom code - which tells me that this isn't really testing your feature code?
if there is a bug in this code in your mailroom, your test wouldn't catch it because it's the same code here.

It is best for your tests to be a little more explicit, like having expected be a report string or list of formatted string rows. This would require your print_report to be refactored into logic function and then separate print function.

PR #912, opened 2018-08-22
Lesson 4 Mailroom

astrawmyer:
I wasn't sure what was meant in the instructions to use a dict to write the letter. My code uses a function with a string and a flat input and I have no idea how i would modify that without making it complicated.

### DIFF
--- a/students/astrawmyer/lesson04/mailroom.py
+++ b/students/astrawmyer/lesson04/mailroom.py
@@ -11,8 +11,10 @@ def displaylist():
         print(name)
 
 def writeletter(name,amount):
-    print('Dear {},'.format(name))
-    print("Thank you for donating ${:.2f} to the Human Fund. Your money will be used appropriately.".format(amount))
+    line_one = 'Dear {},'.format(name)
+    line_two = "Thank you for donating ${:.2f} to the Human Fund. Your money will be used appropriately.".format(amount)
+    letter = line_one + "\n" + line_two
+    return letter
 
 def writereport(donors):
     print('Donor Name                | Total Given | Num Gifts | Average Gift')
@@ -30,7 +32,7 @@ def thankyou():
         if input_name in ddonors.keys():
             donation = float(input("Enter donation amount:"))
             ddonors[input_name].append(donation)
-            writeletter(input_name,donation)
+            print(writeletter(input_name,donation))
             loop_trigger = False
             break
         elif input_name == 'list':
@@ -40,7 +42,7 @@ def thankyou():
             print("Adding {} to donor database".format(input_name))
             donation = float(input("Enter donation amount:"))
             ddonors[input_name] = donation
-            writeletter(input_name,donation)
+            print(writeletter(input_name,donation))
             loop_trigger = False
             break
 
@@ -66,8 +68,10 @@ def all_letters():
         f = open('{}.txt'.format(name), 'w')
         donation = ddonors[name][0]
         letter = writeletter(name, donation)
-        f.write("{}".format(letter))
+        f.write(writeletter(name, donation))
+        #f.write("{}".format(letter))
         f.close()
+    print("Letter files created.")
 
 
 if __name__ == "__main__":
--- a/students/astrawmyer/lesson04/mailroom.py
+++ b/students/astrawmyer/lesson04/mailroom.py
@@ -67,9 +67,7 @@ def all_letters():
     for name in ddonors:
         f = open('{}.txt'.format(name), 'w')
         donation = ddonors[name][0]
-        letter = writeletter(name, donation)
         f.write(writeletter(name, donation))
-        #f.write("{}".format(letter))
         f.close()
     print("Letter files created.")
 
--- a/students/astrawmyer/lesson04/mailroom.py
+++ b/students/astrawmyer/lesson04/mailroom.py
@@ -60,7 +60,10 @@ def thankyou():
             displaylist()
             break
         else:
-
+            print("adding {} to list".format(input_name))
+            donation = float(input("Enter donation amount:"))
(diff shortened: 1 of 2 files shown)


students/astrawmyer/lesson04/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+    donors_report.sort(reverse=True)
+    writereport(donors_report)
+
+#Function to write letters for everyone.
+def all_letters():
+    for name in ddonors:
+        f = open('{}.txt'.format(name), 'w')
+        donation = ddonors[name][0]
+        f.write(writeletter(name, donation))
+        f.close()
```
natasha-aleksandrova: preferred to use `with` when working with files (this was in the course materials).

students/astrawmyer/lesson04/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+    print("Letter files created.")
+
+
+if __name__ == "__main__":
+    main_switch_function = {"1": thankyou, "2": createreport, "3": all_letters, "4": exit}
+    while True:
+        print("What do you want to do?")
+        response = input("1. Send a Thank You, 2. Create a Report, 3. Send all letters, 4. Quit: ")
+        main_switch_function.get(response)()
```
natasha-aleksandrova: 👍

students/astrawmyer/lesson04/mailroom.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+
+# new file for lesson 4 work
+
+#donors = [["Manny Machado",[12.2,2.51,3.20]],["Adam Jones",[1024.14,22.21,323.45]],["Chris Davis",[3.2,5.55,4.20]]]
+ddonors = {"Manny Machado": [12.2,2.51,3.20], "Adam Jones": [1024.14,22.21,323.45], "Chris Davis": [3.2,5.55,4.20]}
```
natasha-aleksandrova: 👍

students/astrawmyer/lesson04/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+    for name in ddonors:
+        f = open('{}.txt'.format(name), 'w')
+        donation = ddonors[name][0]
+        f.write(writeletter(name, donation))
+        f.close()
+    print("Letter files created.")
+
+
+if __name__ == "__main__":
+    main_switch_function = {"1": thankyou, "2": createreport, "3": all_letters, "4": exit}
```
natasha-aleksandrova: use `_` to separate words -> `thank_you` `create_report`

students/astrawmyer/lesson04/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+def thankyou():
+    loop_trigger = True
+    while loop_trigger == True:
+        input_name = input("Enter full name: ")
+        if input_name in ddonors.keys():
+            donation = float(input("Enter donation amount:"))
+            ddonors[input_name].append(donation)
+            print(writeletter(input_name,donation))
+            loop_trigger = False
+            break
```
natasha-aleksandrova: I don't think you need `loop_trigger` - you are using `break` to do exactly that

students/astrawmyer/lesson04/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+    print('-'*67)
+    for i in donors:
+        print('{1:27}${0:11.2f}{2:12}  ${3:12.2f}'.format(*i))
+
+
+# Function to send a thank you letter.
+
+def thankyou():
+    loop_trigger = True
+    while loop_trigger == True:
```
natasha-aleksandrova: True should use `is` comparison, this also applies to `is None` (for object comparison)

students/astrawmyer/lesson04/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+        elif input_name == 'list':
+            displaylist()
+            break
+        else:
+            print("Adding {} to donor database".format(input_name))
+            donation = float(input("Enter donation amount:"))
+            ddonors[input_name] = donation
+            print(writeletter(input_name,donation))
+            loop_trigger = False
+            break
```
natasha-aleksandrova: I think you can simplify this a bit:
```
while True:
    input_name = input("..")
    if input_name == "list":
        display_list()  # in instructions it says to re-prompt so no need to break out of loop
    else:
        donation = float(input("Enter donation amount:"))
        if input_name in ddonors:
            ddonors[input_name].append(donation)
        else:
            ddonors[input_name] = [donation]
        print(writeletter(input_name,donation))
        break

```
that's it!

students/astrawmyer/lesson04/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+        sum_donation = 0
+        avg_donation = 0
+        for i in amount:
+            sum_donation = sum_donation + i
+            num_donation = len(amount)
+        avg_donation = sum_donation/num_donation
+        #print(name, sum_donation, num_donation)
+        donors_report.append([sum_donation, name, num_donation, avg_donation])
+    donors_report.sort(reverse=True)
+    writereport(donors_report)
```
natasha-aleksandrova: nice job on modularizing your code.

students/astrawmyer/lesson04/mailroom.py
```diff
@@ -0,0 +1,80 @@
...
+
+#Function to write letters for everyone.
+def all_letters():
```
natasha-aleksandrova: for function docstrings use:
```
def some_func():
    """Function does function things"""
    pass
```

PR #914, opened 2018-08-22
Lesson 9 Assignment

cosmonautz:


students/KevinWaldron/lesson09/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+    def print_report(self):
+        """ Prints the report """
+        print(self.create_report())
+
+    def mail_everyone(self):
+        """ Handles the mail to everyone menu selection """
+        for donor in self._donors:
+            with open(donor.name.lower().replace(' ', "_") + ".txt", 'w') as f:
+                f.write(donor.create_thank_you(donor.total_donations))
+
```
natasha-aleksandrova: the classes look really great! 

minor nitpick - we technically should have print() functionality outside of data logic classes; consider use case where next feature is to have a UI instead of working with a console script. Now the print functionality is obsolete, but you have them nicely separated into separate functions so not a big deal.

PR #917, opened 2018-08-22
added sort function

ltrisolini:


students/ltrisolini/Lesson03/mailroom.py
```diff
@@ -32,12 +30,30 @@ def thank_you():
...
+            # Get total donation from each donor
+            for i in range(len(each_donor[1])):
+                    total_donation += each_donor[1][i]
+            report.append([each_donor[0], total_donation, len(each_donor[1])])
+
+    # sorting the report based on donations
+    r_sorted = sorted(report, key=lambda r: r[1], reverse=True)
+
+    # Create the report
+    for d in range(len(r_sorted)):
```
natasha-aleksandrova: it is best to loop over list items via simple loop:
`for item in r_sorted`

PR #926, opened 2018-08-27
Exercise6 For Review

ABartles:
Exercise 6 (mailroom & test_mailroom) are ready for review. The args and kwargs lab is still in work, all others are complete.


students/ABartles/Exercise6/mailroom.py
```diff
@@ -0,0 +1,146 @@
...
+
+def create_report():
+    dnr_names = sorted_donors()
+    print(f'{"Donor Name":<20}{"|":<1}{"Total Given":^15}{"|":<1}{"Num Gifts":^15}{"|":<1}{"Average Gift":>15}')
+    print("-" * 68)
+    i_count = 1
+    for i in dnr_names:
+        i_total_gift = sum_gifts(i)
+        i_count = count_gifts(i)
+        i_ave_gift = i_total_gift / i_count
```
natasha-aleksandrova: highly recommend you simplify this:
```
for donor, donations in donors.items():
    total_gift = sum(donations)
    count = len(donations)
    avg = total_gift / count
    # append to the list

# sort the list
# print report
```

students/ABartles/Exercise6/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+    menu
+
+Not sure how to test w/ pytest:
+    create_report
+    run_func
+    __name__ == __main__
+'''
+
+def test_1():
+    assert sum_gifts("Anita Bath") == 350
```
natasha-aleksandrova: your test names should be helpful, test_1 is too vague; a good name is `test_sum_gifts`

students/ABartles/Exercise6/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def test_2():
+    assert count_gifts("Seymour Butz") is 2
```
natasha-aleksandrova: `is` is for object comparison; use `==` here

students/ABartles/Exercise6/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def test_3():
+    name_list = ['Seymour Butz',
+                 'Anita Bath',
+                 'Isabelle Ringing',
+                 'John Smith']
+    name_list_sorted = sorted_donors()
+    for n in name_list:
+        assert n in name_list_sorted
```
natasha-aleksandrova: it should be more like `assert expected == name_list_sorted`

you're not actually testing your sorting logic.

students/ABartles/Exercise6/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+def test_5():
+    send_thanks_all()
+    ab = os.path.isfile('./Anita_Bath.txt')
+    ir = os.path.isfile('./Isabelle_Ringing.txt')
+    js = os.path.isfile('./John_Smith.txt')
+    sb = os.path.isfile('./Seymour_Butz.txt')
+    assert ab is True
+    assert ir is True
+    assert js is True
+    assert sb is True
```
natasha-aleksandrova: simplify to:
```
for fname in ('./Anita_Bath.txt', './Isabelle_Ringing.txt'):
    assert os.path.isfile(fname)
```

students/ABartles/Exercise6/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+    name_list_sorted = sorted_donors()
+    for n in name_list:
+        assert n in name_list_sorted
+
+
+sample_dict = {"Tom": [100, 200], "Jane": [500]}
+
+
+def test_4():
+    assert list_donors(sample_dict) is ['Tom', 'Jane'] or ['Jane', 'Tom']
```
natasha-aleksandrova: this won't do what you expect because of the order of operation; if `list_donors(sample_dict) is ['Tom', 'Jane']` evaluates to False then `or ['Jane', 'Tom']` is always true because it will test for truthiness:
```
>>> assert d is  ['Tom', 'Jane'] or ['Jane', 'Tom']
>>> assert d is ['Tom', 'Jane'] or ['Jane', 'Tom']
>>> d
['cat', 'dog']
```

it sounds like you wanted to handle different order in which case you can sort your actual return:
`assert sorted(actual) == ['Jane', 'Tom']`

students/ABartles/Exercise6/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+    *thanks letter
+    *send_thanks_all
+
+Items with user interaction: (can pytest test/simulate user imput?)
+    rec_donation
+    send_thanks_one
+    menu
+
+Not sure how to test w/ pytest:
+    create_report
```
natasha-aleksandrova: to test create_report you need to split it into two functions:
1. `generate_report` function which can return a list of rows or already formatted string
2. `print_report` function which consumes output from generate report and prints the report


you will write a unit test for `generate_report` function.

PR #932, opened 2018-08-30
Lesson04 Mailroom and Kata Ready for Review

srepking:
Hi, 
Please let me know if you want me to do more with the kata. I feel like I could really go down a rabbit hole working on punctuation and paragraphs and such.

Thanks!

### DIFF
--- a/students/srepking/Lesson04/mailroom2.py
+++ b/students/srepking/Lesson04/mailroom2.py
@@ -1,11 +1,12 @@
 def thank_you():
     print('\n''Ok, you chose 1, lets see what we can do.')
     while True:
-        name = input('\n''Type the full name to add entry, \'list\' to see list of names, or \'e\' to exit: > ')
+        name = input('\n''Type the full name to add entry, '
+                     '\'list\' to see list of names, or \'e\' to exit: > ')
         # See if they want a list of names.
         if name != 'list':
             break
-        for x in donors:    
+        for x in donors:
             print(x)
     # Add Name to list if not in list
     if name not in donors and name != 'e':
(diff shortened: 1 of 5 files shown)


students/srepking/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,122 @@
...
+def goodbye():
+    pass
+
+
+if __name__ == '__main__':
+    import os
+
+    donors = {'Joe Edgar Allen Poe the Third ': [1, 4, 200],
+              'Jack': [4, 5], 'Jill': [4], 'Jake': [.30],
+              'Jim': [1, 2, 1.04]}
```
natasha-aleksandrova: 👍

students/srepking/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,122 @@
...
+    while True:
+        response = input(
+            '\nChoose an Action:\n'
+            '1 - Send a Thank You\n'
+            '2 - Create a Report\n'
+
+            '3 - Send letters to everyone\n'
+            '4 - Quit\n'
+            '>>')
+        switch_dict = {'1': thank_you, '2': report, '3': letters_for_all}
```
natasha-aleksandrova: this should be outside of the loop, no need to redefine it every time.

students/srepking/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,122 @@
...
+        switch_dict.get(response, goodbye)()
+        if response == '4':
+            print('Goodbye!')
+            break
```
natasha-aleksandrova: you can avoid special logic if you map option `4` to a `quit` or `exit` type function.

students/srepking/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,122 @@
...
+        with open(donor + '.txt', 'w') as to_file:
+            to_file.write(letter)
+
+
+def goodbye():
+    pass
+
+
+if __name__ == '__main__':
+    import os
```
natasha-aleksandrova: imports go at the top of the file.

students/srepking/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,122 @@
...
+    donors_f = report()
+    for donor, donation in donors.items():
+        donation_summary = donors_f[donor]
+        letter = f'Dear {donor}, thank you so much for your last ' \
+                 f'contribution of ${donation[-1]:.2f}! ' \
+                 f'You have contributed a total of ${donation_summary[0]:.2f}, ' \
+                 f'and we appreciate your support!'
+        # Write the letter to a destination
+        with open(donor + '.txt', 'w') as to_file:
+            to_file.write(letter)
```
natasha-aleksandrova: 👍

students/srepking/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,122 @@
...
+        total = 0
+        count = 0
+        average = 0
+        i = 0
+        for i in donations:
+            total = total + i
+        # Calculate total number of donations, not counting $0
+        for i in donations:
+            if i != 0:
+                count = count + 1
```
natasha-aleksandrova: use `sum` for total and `len` for count logic

students/srepking/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,122 @@
...
+          f"{'Num Gifts':^{num_wi}}| {'Average Gift':^{ave_wi}}")
+    print(f"{'-':-^{(name_wi+tot_wi+ave_wi+num_wi+8)}}")
+    # print("-")
+    for key in list_sorted:
+        temp = donors_f[key]
+        print(f"{key:{name_wi}}${temp[0]:{tot_wi}}{temp[1]:^{num_wi}}   "
+              f"${temp[2]:>{ave_wi}.2f}")
+
+    print('\n')
+    return donors_f
```
natasha-aleksandrova: i think you need to modularize this function; it appears to be doing too much for both reporting and thank you letters functionality, when thank you letters doesn't need a full donors report printed as part of it.

students/srepking/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,122 @@
...
+        # See if they want a list of names.
+        if name != 'list':
+            break
+        for x in donors:
+            print(x)
+    # Add Name to list if not in list
+    if name not in donors and name != 'e':
+        print('Name is not there, added to the list')
+        donors[name] = []
+    if name != 'e':
```
natasha-aleksandrova: i think it would be cleaner to first check for `e`, like
```
if name == 'e':
    return
```
and then do donor update or additional logic.

students/srepking/Lesson04/mailroom2.py
srepking: Got it. Thanks!

students/srepking/Lesson04/mailroom2.py
srepking: I implemented the sum() function. However, I don't think the len() function will work for what I am doing. I am not counting donations of $0 as donations. If I use len(), I have to assume all numbers in my list are donations, and I thought that a donation of $0 should be disregarded. Or  is there another way to use len() I am overlooking? Thank you!

students/srepking/Lesson04/mailroom2.py
srepking: I agree. Fixed.

students/srepking/Lesson04/mailroom2.py
srepking: Change made!

students/srepking/Lesson04/mailroom2.py
natasha-aleksandrova: 👍

students/srepking/Lesson04/mailroom2.py
natasha-aleksandrova: 👍

students/srepking/Lesson04/mailroom2.py
natasha-aleksandrova: I see the new Pull Request and will take a look at the latest code and look into more details for this part, I will provide recommendation there.

PR #934, opened 2018-08-31
Updated file based on comments.

srepking:
I updated the file based on the previous pull comments. Thank you!

### DIFF
--- a/students/srepking/Lesson04/mailroom2.py
+++ b/students/srepking/Lesson04/mailroom2.py
@@ -1,40 +1,42 @@
+import os
+
 def thank_you():
     print('\n''Ok, you chose 1, lets see what we can do.')
     while True:
         name = input('\n''Type the full name to add entry, '
                      '\'list\' to see list of names, or \'e\' to exit: > ')
-        # See if they want a list of names.
+        if name == 'e':
+            return
         if name != 'list':
             break
         for x in donors:
             print(x)
     # Add Name to list if not in list
-    if name not in donors and name != 'e':
+    if name not in donors:
         print('Name is not there, added to the list')
         donors[name] = []
-    if name != 'e':
         amount = input('\n''What is the donation amount? or \'e\' to exit >')
-        if amount != 'e':
-            new_amount = [float(amount)]
-            old_amount = donors[name]
-            donors[name] = old_amount + new_amount
-            print('Thank you so much for the generous gift of ${0:.2f}, {1}!'
-                  .format(float(amount), name))
+        if amount == 'e':
+            return
+        new_amount = [float(amount)]
+        old_amount = donors[name]
+        donors[name] = old_amount + new_amount
+        print('Thank you so much for the generous gift of ${0:.2f}, {1}!'
+              .format(float(amount), name))
 
 
-def report():
-    print('\n''A summary of your donors donations:')
+def summary_donors() -> object:
+
     # Create a new dictionary with Total, number of donations,
     # and average donation amount
     donors_f = {}
     for name, donations in donors.items():
         # Calculate total donation amount
-        total = 0
+        total = sum(donations)
         count = 0
         average = 0
         i = 0
-        for i in donations:
-            total = total + i
+
         # Calculate total number of donations, not counting $0
         for i in donations:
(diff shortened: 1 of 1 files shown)


students/srepking/Lesson04/mailroom2.py
```diff
@@ -43,10 +45,15 @@ def report():
         if count > 0:
             average = total / count
         donors_f[name] = [total, count, average]
```
natasha-aleksandrova: I see what you're doing now. I overlooked at it first, this is the first version that's doing zero amount validation 🙂.

Since donation of zero doesn't really make sense as a valid amount in this program, it is best to prevent zero from being stored by adding a validation early on.

When user enters amount you can check that the amount is > 0.

students/srepking/Lesson04/mailroom2.py
```diff
@@ -115,8 +123,6 @@ def goodbye():
             '3 - Send letters to everyone\n'
             '4 - Quit\n'
             '>>')
-        switch_dict = {'1': thank_you, '2': report, '3': letters_for_all}
-        switch_dict.get(response, goodbye)()
-        if response == '4':
-            print('Goodbye!')
-            break
+
+        switch_dict.get(response, wrong_choice)()
```
natasha-aleksandrova: clever! (use of `wrong_choice`)

students/srepking/Lesson04/mailroom2.py
```diff
@@ -96,16 +102,18 @@ def letters_for_all():
             to_file.write(letter)
 
 
-def goodbye():
+def wrong_choice():
     pass
 
+def quit():
+    exit()
```
natasha-aleksandrova: `quit` is a built-in function (which is why it's highlighted in blue vs other functions you have).

you want to avoid over-writing built-ins 😉; rename it to something like `quit_program`

students/srepking/Lesson04/mailroom2.py
```diff
@@ -1,40 +1,42 @@
...
-            donors[name] = old_amount + new_amount
-            print('Thank you so much for the generous gift of ${0:.2f}, {1}!'
-                  .format(float(amount), name))
+        if amount == 'e':
+            return
+        new_amount = [float(amount)]
+        old_amount = donors[name]
+        donors[name] = old_amount + new_amount
+        print('Thank you so much for the generous gift of ${0:.2f}, {1}!'
+              .format(float(amount), name))
```
natasha-aleksandrova: the code is missing feature for adding donations to existing donors. 

Also, suggestion to simplify:
```
amount = float(input("..."))
# add validation here for amount if you'd like
if name in donors:
    donors[name].append(amount)
else:
    donors[name] = [amount]

print("thank you...")

```

students/srepking/Lesson04/mailroom2.py
srepking: Fixed!

students/srepking/Lesson04/mailroom2.py
srepking: I implemented the suggestion 'donors[name].append(float(amount))', and then caught the exception instead of using if/else.

natasha-aleksandrova: 👍 

there is a great library to make this even easier:
https://www.accelebrate.com/blog/using-defaultdict-python/

I don't mean keep hammering on your code but you seem eager to learn things 🙂

srepking: Hi Natasha, 
I looked over the default dictionary link you sent, and I see how I could have used it to create a blank entry for a key with no value instead of the if/else loop above. However, I kind of went a different  direction in the last mailroom exercise for Lesson05 and am handling this with an exception now, so that if someone tries to enter a new name, and exception is caught, and the user is prompted to verify that this is what they want to do. Is that OK? or is there still a way I could make this better with the defaultdic library? I am not seeing it! Thanks!

natasha-aleksandrova:
If you want to have this additional behaviour then using exceptions makes the most sense.

natasha-aleksandrova:
@srepking you're very welcome!

PR #935, opened 2018-09-01
Lesson 5 for Review

srepking:


srepking:
Yes and thanks!, though the file I pushed this morning is about it short of
PIP8 editing. I tried to break it down into more testible functions, but
think I am at the point where I need some tips. Happy Labor Day btw!

students/srepking/Lesson06/mailroom4.py
```diff
@@ -0,0 +1,176 @@
...
+
+            amount = input('\n''What is the donation amount? or \'e\' to exit >')
+            if amount == 'e':
+                return
+            else:
+                thank_you(donors, name, amount)
+
+
+
+def create_list(some_donors):
```
natasha-aleksandrova: no reason to pass in donors dict when it's already in global space.

students/srepking/Lesson06/mailroom4.py
```diff
@@ -0,0 +1,176 @@
...
+    # This prints the list of donors
+    for x in some_donors:
+        print(x)
+
+
+def thank_you(dict_donors, some_name, some_amount):
```
natasha-aleksandrova: no reason to pass in donors dict when it's already in global space.

students/srepking/Lesson06/mailroom4.py
```diff
@@ -0,0 +1,176 @@
...
+            print('\nYou entered an invalid amount!!\n')
+            return
+    except ValueError:
+        print('\nYou entered an invalid amount!!\n')
+        return ValueError
+
+    dict_donors[some_name].append(float(some_amount))
+    print('Thank you so much for the generous gift of ${0:.2f}, {1}!'
+          .format(float(some_amount), some_name))
+    return dict_donors
```
natasha-aleksandrova: return value in this function looks to be a test only code, since it isn't actually used - best to avoid code like that.

students/srepking/Lesson06/mailroom4.py
```diff
@@ -0,0 +1,176 @@
...
+    for donor, donation in donors.items():
+        donation_summary = donors_f[donor]
+        letter = f'Dear {donor}, thank you so much for your ' \
+                 f'last contribution of ${donation[-1]:.2f}! ' \
+                 f'You have contributed a total of $' \
+                 f'{donation_summary[0]:.2f}, ' \
+                 f'and we appreciate your support!'
+        # Write the letter to a destination
+        with open(donor + '.txt', 'w') as to_file:
+            to_file.write(letter)
```
natasha-aleksandrova: this would be a good function to unit test.

students/srepking/Lesson06/mailroom4.py
```diff
@@ -0,0 +1,176 @@
...
+    print(f"{'Donor Name':{name_wi}}| {'Total Given':^{tot_wi}}| "
+          f"{'Num Gifts':^{num_wi}}| {'Average Gift':^{ave_wi}}")
+    print(f"{'-':-^{(name_wi+tot_wi+ave_wi+num_wi+8)}}")
+    # print("-")
+    for key in list_sorted:
+        temp = donors_f[key]
+        print(f"{key:{name_wi}}${temp[0]:{tot_wi}.2f}{temp[1]:^{num_wi}}   "
+              f"${temp[2]:>{ave_wi}.2f}")
+
+    print('\n')
```
natasha-aleksandrova: to unit test this function, separate report generation by having a function return a list or a string, and then separate function to actually print the report.

srepking: I am not sure I am understanding. The 'sort_list()' function is returning a list of keys to print in order, and then the 'report()' function is printing. Isn't this what you are are suggesting in this comment?

natasha-aleksandrova: Here is an example of what I mean:
```
def get_report_data():
    # ... some code here
    for key in list_sorted:
        rows.append(f"{key:{name_wi}}${temp[0]:{tot_wi}.2f}{temp[1]:^{num_wi}}   "
              f"${temp[2]:>{ave_wi}.2f}")
    return rows


def generate_report():
    rows = get_report_data()
    print("\n".join(rows))
```

this allows you to unit test get_report_data() function without having to deal with "how do I ensure the right data is printed?" and also separates your "UI" code from logic code. This will be an actual requirement in Lesson9 version of mailroom.. but making mailroom unit testable in this lesson is the first step in that direction.

students/srepking/Lesson06/mailroom4.py
natasha-aleksandrova: ah I see why! it's because you define `donors` in `if __name__ == "__main__":` block and when you import the mailroom that block isn't imported or visible.

if you define `donors` at the top of the file then you should have no problems.

srepking: Great! I will try that. Thanks!

students/srepking/Lesson06/mailroom4.py
srepking: Fixed.

students/srepking/Lesson06/mailroom4.py
srepking: Sorry to keep asking question on this. Is it preferable to leave all the format in the 'generate_report' function and just pass the dictionary data from 'get_report_data'? Or is the idea to also test the formatting of the string in get_report_data()?

srepking: Ok, I put the test with formatting included. Please let me know if you intended otherwise.
Thank you!

natasha-aleksandrova: From a unit test point of view - you're benefiting from just having formatting in the get_report_data() so that you're getting most code coverage.

From a real case scenario.. it's kind of a grey area.. the formatting is kind of specific to print functionality which we wanted to separate. So i'm not inclined to say towards either option, I think for simplicity here it makes sense to have formatting in `get_report_data` and `generate_report` just handles the printing.

PR #937, opened 2018-09-02
Lesson 05, parts 1 and 2 complete.

rbillupsii:


students/RoyC/Lesson05/mailroom.py
```diff
@@ -0,0 +1,138 @@
...
+            record = {'donor_name' : donor_name, 'amt' : "{:.2f}".format(amt), 'total' : "{:.2f}".format(get_total(donations))}
+            print(ltr_template.format(**record))
+            break
+    
+def thank_all():
+    """
+    Write thank you letters for all donors to separate files
+    """
+    for donor, donations in donors.items():
+        f = open(donor.replace(' ',  '_') + ".txt", 'w')
```
natasha-aleksandrova: best to use `with` statement when working with files.

students/RoyC/Lesson05/mailroom.py
```diff
@@ -0,0 +1,138 @@
...
+def get_total(donations):
+    """
+    Return the total of the given list of donations
+    Argument:
+        donations - list of donation amounts (floating point nums)
+    """
+    total = 0;
+    for donation in donations:
+        total += donation
+    return total
```
natasha-aleksandrova: the whole thing can be replaced with `sum` function... this was also a good candidate for comprehensions (which are part of this lesson).

PR #952, opened 2018-09-11
Lesson 2 - Exercises

aguardia:
I built three functions with the different goals

### DIFF
--- a/students/AleGuardia/lesson2/FizzBuzz.py
+++ b/students/AleGuardia/lesson2/FizzBuzz.py
@@ -1,0 +1,15 @@
+# FizzBuzz solution by Alejandro Guardia
+
+for i in range(1,101):
+    word = ""
+    if i % 3 == 0 or i % 5 == 0:
+        if i % 3 == 0:
+            word += "Fizz"
+        if i % 5 == 0:
+            word += "Buzz"
+        print(word)
+        continue
+    print(i)
+
+
+
--- a/students/AleGuardia/lesson2/print_grid.py
+++ b/students/AleGuardia/lesson2/print_grid.py
@@ -1,0 +1,22 @@
+# collection of functions to print grids by Alejandro Guardia
+
+
+def print_grid_std():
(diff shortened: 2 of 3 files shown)


aguardia:
Added the fibonacci and Lucas sequence exercise

students/AleGuardia/lesson2/FizzBuzz.py
```diff
@@ -0,0 +1,15 @@
...
+
+for i in range(1,101):
+    word = ""
+    if i % 3 == 0 or i % 5 == 0:
+        if i % 3 == 0:
+            word += "Fizz"
+        if i % 5 == 0:
+            word += "Buzz"
+        print(word)
+        continue
```
charlesrobison: continue is not necessary here.  Solution works well.

students/AleGuardia/lesson2/print_grid.py
```diff
@@ -0,0 +1,65 @@
...
+    for i in range(number):
+        print(row)
+        print()
+        for j in range(units):
+            print(column)
+            print()
+    print(row)
+
+
+print_grid_custom_two(5,3)
```
charlesrobison: Works well - nice job.

students/AleGuardia/lesson2/series.py
```diff
@@ -0,0 +1,43 @@
...
+        print("The expected " + str(i) + " element in the " + name + " sequence is: " + str(series[i]))
+        print("The result from the function is " + str(sum_series(i,2,1)))
+        if sum_series(i,2,1) == series[i]:
+            print("Test Passed")
+        else:
+            print("Test Failed)")
+
+
+test_series(fib_series,"Fibonacci")
+test_series(luc_series,"Lucas")
```
charlesrobison: Tests look good but fibonacci and lucas functions don't seem to be functioning as expected.  Recommend reviewing assignment requirements.  Should be able to run each function and return expected results - e.g., fibonacci(4) should return 3 etc.

PR #955, opened 2018-09-12
mailroom part 2 with option to save emails in same directory

ltrisolini:
Lesson04 assignment1 mailroom part 2


ltrisolini:
Here is assignment 2: trigram, thanks!

PR #957, opened 2018-09-14
FizzBuzz

chelseamsmith:
Please look at fizzbuzz.py for my solution to the lesson 2 fizzbuzz problem.


students/ChelseaSmith/Lesson3/slicingfunctions.py
```diff
@@ -0,0 +1,23 @@
+def flip_first_last(seq):
+    orig = tuple(seq)
+    new = list(seq)
+    new[0] = orig[-1]
+    new[-1] = orig[0]
+    return new
```
natasha-aleksandrova: you should only use slicing for this assignment; right now it is using indexing.

here is how you would do it using slicing:
```
def flip_first_last(seq):
```

PR #959, opened 2018-09-15
Submitting tasks 1 and 2 of lesson05

ptfbanks:
Exception exercise and Mailroom-3 submitted.
Two issues with mailroom acknowledged.
1) formatting in Write statements - I could not figure how to establish 2 decimal places.
2) Exception logic introduced a dual step to quit..  I could not figure how to prevent this

### DIFF
--- a/students/ptfbanks/Lesson05/pf_mailroom-3.py
+++ b/students/ptfbanks/Lesson05/pf_mailroom-3.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python3
+#Instructions:\ Add exceptions and use comprehensions to improve your mailroom project.
+#Exceptions: update your code to handle errors better – like when a user inputs bad data.
+#Comprehensions: Can you use comprehensions to clean up your code a bit?
+#---------------------Set-up----------------------------------#
+#  MailRoom III exercise lesson 05 Include Compression and exception handling
+print('Your desktop has been up-dated with the Donations Management II applicaton.\n'
+'We trust you will appreciate the added menue options and snappy output.\n'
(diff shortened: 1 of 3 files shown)


students/ptfbanks/Lesson05/pf_mailroom-3.py
```diff
@@ -0,0 +1,123 @@
...
+    try:
+        while True:
+            pick=input(choice)
+            if pick =='4':
+                print('So long.....')
+                break
+            menu[pick]()
+    except KeyError:
+        print('command not understood.  Plese try again.\n')
+        menu_select(choice)
```
natasha-aleksandrova: your KeyError is supposed to be only around the code that will throw that exception, in this case around line 61.

The main reason why your loop logic is now behaving differently is because you have try/except around the actual while True; which isn't the case almost always.

students/ptfbanks/Lesson05/pf_mailroom-3.py
```diff
@@ -0,0 +1,123 @@
...
+    if whom.lower() == 'all':   #Prepare text of T/U letters for all donors showing summary amounts
+        for name in d_dict:
+            send_where (name, send, direct)
+    else:                                       #(Excepton check incluced)
+        try:
+            #print("test", whom, " - ", send, "---", direct, "+++") 
+            send_where(whom, send, direct)
+        except KeyError:
+            print("\n There is no record of donations by: ", whom, ".  Refresh donor report and try again.")
+            menu_select(choice)
```
natasha-aleksandrova: don't call the main loop manually; you're in a while True so it will loop naturally

PR #963, opened 2018-09-17
Corrections made to lesson 2 assignments

jayjonesseattle:
I have made the suggested corrections to the fizzbuzz.py and series.py assignments. I will use an IDE moving forward to prevent the indenting errors if that is okay

### DIFF
--- a/students/terrance_jones/lesson2/series.py
+++ b/students/terrance_jones/lesson2/series.py
@@ -1,5 +1,4 @@
 def fibonacci(n):
-	assert (n > 0), "Must be greater than 0"
 	counter = 0
 	a = 0
 	b = 1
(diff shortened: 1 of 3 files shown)


students/terrance_jones/lesson2/series.py
```diff
@@ -60,16 +73,20 @@ def sum_series(n, a=0, b=1):
...
+			return True
+		else:
+			return False
+
+
+
+if __name__ == '__main__':
+	print("testing")
+	print()
+	assert fibonacci(6) == True
```
natasha-aleksandrova: from instructions 

> The function should return the nth value in the fibonacci series (starting with zero index)


so it should be `assert fibonacci(3) == 2`

natasha-aleksandrova:
Yes I recommend using an IDE like Pycharm, which is great for improving code style and catch violations early.

jayjonesseattle:
Thank you for your feedback. I didn’t quite understand the asset part of
the exercise but you helped make it clear. I apologize for not correcting
the indentation. I totally forgot. I will make the corrections and push
them later today. Thank you!

natasha-aleksandrova:
sounds good!

PR #966, opened 2018-09-18
Lesson 6 submission.

astrawmyer:


students/astrawmyer/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+    assert m.ddonors == {"Manny Machado": [12.2,2.51,3.20,2131],
+                        "Adam Jones": [1024.14,22.21,323.45],
+                        "Chris Davis": [3.2,5.55,4.20]}
+    
+
+def test_thank_you_new_donor():
+    input_val = ["Robert Wickens", 9000.1]
+    def mock_input(s):
+        return input_val.pop(0)
+    m.input = mock_input
```
natasha-aleksandrova: just FYI - the idea here is correct to use mocking but you want to use libraries to handle that for you in a safe way (`mock` library)

students/astrawmyer/lesson06/mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+#    ddonors[name].append(donation)
+
+
+#def add_donor():
+
+
+
+
+
+def thank_you():
```
natasha-aleksandrova: make sure functions have 2 line spaces between them

PR #967, opened 2018-09-18
Submission for slicing assignment

chelseamsmith:


students/ChelseaSmith/Lesson3/slicingfunctions.py
```diff
@@ -1,17 +1,17 @@
 def flip_first_last(seq):
-    orig = tuple(seq)
-    new = list(seq)
-    new[0] = orig[-1]
-    new[-1] = orig[0]
+    orig = tuple(seq)  # make a tuple so this stays safe
+    new = list(seq)  # make a list to mess around with
+    new[0] = orig[-1]  # make first value of new list last value of old list
+    new[-1] = orig[0]  # make last value of new list first value of old list
     return new
```
charlesrobison: Works well for tuples, but not quite getting the expected results for strings.

students/ChelseaSmith/Lesson3/slicingfunctions.py
```diff
@@ -1,17 +1,17 @@
...
 
 
 def remove_ev_other(seq):
-    return seq[::2]
+    return seq[::2]  # skips every other value
```
charlesrobison: These and the rest work well - nice work!

students/ChelseaSmith/Lesson3/mailroom.py
```diff
@@ -0,0 +1,57 @@
...
+          ["Kate Smith", 100],
+          ["Kate Jones", 175, 50],
+          ["Alex Smith", 200, 200, 200]]  # creates initial donor record
+
+choose = 0  # creates variable for selecting from main menu
+
+def report(listin):  # prints a report of all donors
+    print(f"Donor name" + " " * 17 + " | Total Given | Num Gifts | Average Gift\n")
+    for row in listin:
+        sum = 0
```
natasha-aleksandrova: this variable is overwriting the built-in function `sum` which actually should use below ... but it would error out because of this var 😉

students/ChelseaSmith/Lesson3/mailroom.py
```diff
@@ -0,0 +1,57 @@
+record = [["James Smith", 85, 100, 75],
```
natasha-aleksandrova: should be `records` to properly identify what the data is (plural vs singular)

students/ChelseaSmith/Lesson3/mailroom.py
```diff
@@ -0,0 +1,57 @@
...
+
+def report(listin):  # prints a report of all donors
+    print(f"Donor name" + " " * 17 + " | Total Given | Num Gifts | Average Gift\n")
+    for row in listin:
+        sum = 0
+        num = 0
+        name = row[0]
+        for item in row[1:]:
+            sum = sum + int(item)
+            num = num + 1
```
natasha-aleksandrova: no need for the `for` loop, instead:
```
donations = row[1:]
total = sum(donations)
num_of_donations = len(donations)
```

students/ChelseaSmith/Lesson3/mailroom.py
```diff
@@ -0,0 +1,57 @@
...
+        print(f"{name:<28}${sum:12.2f}{num:10}   ${(sum / num):13.2f}")
+
+
+def checknew(listin, name):  # returns true if name is not already in list input
+    new = True
+    for row in listin:
+        if name == row[0]:
+            new = False
+    return new
```
natasha-aleksandrova: simplify (no need for temp var):
```
for row in listin:
        if name == row[0]:
            return False
    return True

```

students/ChelseaSmith/Lesson3/mailroom.py
```diff
@@ -0,0 +1,57 @@
...
+                print(item[0])
+            name = input("Please enter the donor's full name or list for a list of donors.")
+        if checknew(record, name):  # checks if new donor
+            record.append([name])  # adds new donor to database
+        donation = int((input("Enter donation amount")))  # gets donation amount
+        record[getrow(record,name)].append(donation)  # uses getrow function to find correct index in record and appends a donation to the end of the list at that index
+        print(f"Dear {name},\nThank you for your generous donation of ${donation}.\n"   # makes a nice letter
+              f"Sincerely,\n Generic Charities")
+    elif choose == 2:  # prints report
+        report(record)
```
natasha-aleksandrova: this code needs to be inside `if __name__ == "__main__"` block (it will be even more important in further lessons that revision the mailroom)

natasha-aleksandrova: this block should be simplified to something like:
```
while True:
    choose = input("..")
    if choose == 1:
        thank_you()
    elif choose == 2:
        create_report()
    elif choose == 3:
        quit_program()
```

the entry should be clean and readable and redirect to feature functions (this will be pretty much a requirement in lesson 4, mailroom part 2)

students/ChelseaSmith/Lesson3/slicingfunctions.py
natasha-aleksandrova: this function should only use slicing, not indexing (which is what you have); there is also no need for special type handling or casting - if using slicing it will work for tuples, strings and lists.

PR #971, opened 2018-09-19
added GridPrinterExercise file: Lesson 2 Assignment

dreifler:
Finished the assignment from Lesson 2, the Grid Printer Exercise.


students/dreifler/Lesson2/FizzBuzz.py
```diff
@@ -0,0 +1,15 @@
+#FizzBuzz
+
+def fizzBuzz():
+    for a in range (1, 100):
```
natasha-aleksandrova: should be `range(1, 101)` to include 100 number.

also per pep8 style guidelines variable and function names follow snake_case convention -> `def fizz_buzz`

PR #974, opened 2018-09-19
please reveiw

jayjonesseattle:
please review fizzbuzzUPDATED and series.py. I think i have corrected the indentation errors, at least it seems that way on my end =).

### DIFF
--- a/students/terrance_jones/lesson2/fizzbuzzUPDATED.py
+++ b/students/terrance_jones/lesson2/fizzbuzzUPDATED.py
@@ -1,0 +1,22 @@
+# To change this license header, choose License Headers in Project Properties.
+# To change this template file, choose Tools | Templates
+# and open the template in the editor.
+def fizzbuzz():
+	''' Fizzbuzz prints fizz for numbers
+	divisible by 3, prints buzz for numbers divisible by 5
+	and prints fizzbuzz for numbers divisible by 3 and 5. All others 
+	it prints the number'''
+        
+    for i in range(100):
+	if i == 0:
+            print()
+		elif(i%3 == 0 and i%5!=0):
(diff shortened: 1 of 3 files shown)


students/terrance_jones/lesson2/fizzbuzzUPDATED.py
```diff
@@ -0,0 +1,22 @@
...
+	''' Fizzbuzz prints fizz for numbers
+	divisible by 3, prints buzz for numbers divisible by 5
+	and prints fizzbuzz for numbers divisible by 3 and 5. All others 
+	it prints the number'''
+        
+    for i in range(100):
+	if i == 0:
+            print()
+		elif(i%3 == 0 and i%5!=0):
+            print("Fizz")
```
natasha-aleksandrova: these indentations are off here as well

jayjonesseattle:
I’m not really sure what’s going on. Originally I was using sublime as it
was recommended by the course.  Since you suggested I use an IDE I tried it
using Netbeans as that is the one I’m most familiar with. Any suggestions
as to what I should do? Is there a better IDE or editor that you would
suggest? Are you noticing these errors in each of the assignments or is it
just the fizz buzz?

natasha-aleksandrova:
@jayjonesseattle sorry I missed this message. Yes it's not only fizz buzz.

Sublime is a great editor .. BUT.. it needs to be configured properly for python because it's a generic editor that can be used with any language. 

I have not used Netbeans for python (didn't even know it can support it) so not sure how good it is.. but I **highly** recommend Pycharm. You can get free edition or professional with your uw email. 

Pycharm requires pretty much no initial configuration and everything is ready to go after install including the correct indentation.

PR #976, opened 2018-09-20
Lesson 09

andykhkwok:
Ready for review.

### DIFF
--- a/students/AndyKwok/lesson09/mailroom_pt9.py
+++ b/students/AndyKwok/lesson09/mailroom_pt9.py
@@ -1,0 +1,148 @@
+# Description: Mailroom Program - OOP
+# Author: Andy Kwok
+# Last Updated: 9/18/18
+# ChangeLog: 
+# v1.0 - Initialization
+
+#!/usr/bin/env python3
+ 
+
+class Donor():
+    ''' Class to hold donor name and donation information'''
+    def __init__(self, First = None, Last = None):
+        self.__First = First
+        self.__Last = Last
+        self.Donation_Record = []
+        self.__Fullname = None
+        
+    @property
+    def First(self):
+        return self.__First
+
+    @First.setter
+    def First(self, name):
+        self.__First = name
+        
+    @property
+    def Last(self):
+        return self.__Last       
+    
+    @Last.setter
+    def Last(self, name):
+        self.__Last = name
+    
+    @property
+    def Fullname(self):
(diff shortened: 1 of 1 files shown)


students/AndyKwok/lesson09/mailroom_pt9.py
```diff
@@ -0,0 +1,148 @@
+# Description: Mailroom Program - OOP
+# Author: Andy Kwok
+# Last Updated: 9/18/18
+# ChangeLog: 
+# v1.0 - Initialization
+
+#!/usr/bin/env python3
+ 
+
+class Donor():
```
natasha-aleksandrova: no need for parens -> `class Donor:`

students/AndyKwok/lesson09/mailroom_pt9.py
```diff
@@ -0,0 +1,148 @@
...
+    ''' Class to hold donor name and donation information'''
+    def __init__(self, First = None, Last = None):
+        self.__First = First
+        self.__Last = Last
+        self.Donation_Record = []
+        self.__Fullname = None
```
natasha-aleksandrova: all of your var names should be lower case:
```
def __init__(self, first=None, last=None):
    self.first = first
    self.last = last
....
```
also, there is usually no need for vars to be private - so no need for double under scores.

There are cases when it is useful but not here.

students/AndyKwok/lesson09/mailroom_pt9.py
```diff
@@ -0,0 +1,148 @@
...
+    def First(self, name):
+        self.__First = name
+        
+    @property
+    def Last(self):
+        return self.__Last       
+    
+    @Last.setter
+    def Last(self, name):
+        self.__Last = name
```
natasha-aleksandrova: no need for above to be properties

students/AndyKwok/lesson09/mailroom_pt9.py
```diff
@@ -0,0 +1,148 @@
...
+    two.AddDonation(15)
+    two.AddDonation(153)
+    two.AddDonation(20)
+    DonorInfo.AddDonor(two)
+
+    three = Donor()
+    three.First = 'A'
+    three.Last = 'Character'
+    three.AddDonation(1200)
+    DonorInfo.AddDonor(three)
```
natasha-aleksandrova: you should have a separate file with unit tests.

PR #977, opened 2018-09-20
adding correct fizzbuzz.py with indentions fixe

jayjonesseattle:
Please review fizzbuzzUPDATED.  i finally believe i corrected the indention errors. Thank you for your help.

### DIFF
--- a/students/terrance_jones/lesson2/fizzbuzzUPDATED.py
+++ b/students/terrance_jones/lesson2/fizzbuzzUPDATED.py
@@ -1,22 +1,17 @@
-# To change this license header, choose License Headers in Project Properties.
-# To change this template file, choose Tools | Templates
-# and open the template in the editor.
 def fizzbuzz():
-	''' Fizzbuzz prints fizz for numbers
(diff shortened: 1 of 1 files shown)


students/terrance_jones/lesson2/fizzbuzzUPDATED.py
```diff
@@ -1,22 +1,17 @@
...
-	elif(i%3==0 and i%5==0):
-		print("FizzBuzz")
-	else:
-            print(i)
+    '''Fizzbuzz prints fizz for numbers divisible by 3, prints buzz for numbers
+    divisible by 5 and prints fizzbuzz for numbers divisible by 3 and 5. All 
+    other it prints the number'''
 
-        i = i + 1
+        for i in range(100):
```
natasha-aleksandrova: the indentation is now 4 spaces. but there should be no indentation after your docstrings, like:
```
def some_method():
    """this is a function
    second line"""
    for i in range():
        pass
```
also notice that docstrings use triple double quotes.

PR #980, opened 2018-09-23
mailroom dict

au-gold:
lesson04

### DIFF
--- a/students/Mintae_K/lesson04/mailroom2.py
+++ b/students/Mintae_K/lesson04/mailroom2.py
@@ -1,0 +1,57 @@
+# Sending a Thank you
+def sendingthank(donors):
+    f_name = input('Please type the Full Name or list> ')
+    while f_name == 'list':
+        print(' ')
+        for x in donors:
+            print(x)
+        print(' ')
+        f_name = input('Please type the Full Name or list> ')
+    if donors.get(f_name, False):
+        s = 'Existing donor, please type the donation amount from {}> '
+        amount = input(s.format(f_name))
+        donors[f_name].append(float(amount))
+    else:
+        s = 'New donor, please type the donation amount from {}> '
+        amount = input(s.format(f_name))
+        donors[f_name] = [float(amount)]
+    print('\n', "Thank you {} for the generous donation!".format(f_name), '\n')
+    return donors
+
+
+# Creating Report
+def createreport(donors):
+    s = '\n, {:<20}| {:<10} | {:<10}| {:<20}'
+    print(s.format('Donor Name', 'Total Given', "Num Gift", "Average Gift"))
+    print('-' * 60)
+    total =[]
+    for keys in donors:
+        num_gift = len(donors[keys])
+        total.append([sum(donors[keys]), num_gift, keys])
+    total.sort(reverse=True)
+    s = '{:<20}  $ {:<10.2f}  {:^12} $ {:<20.2f}'
+    for i in range(len(total)):
+        print(s.format(total[i][2], total[i][0], total[i][1], total[i][0] / total[i][1]))
+    print(' ')
+    return
+
+
+def main():
+    # Mailroom
+    donors={'William Gates, III':[54842.49,48965.25],'Mark Zuckerberg':[7852.25,48652.0,3548.0],'Jeff Bezos':[5486.0,58794.02,7412.1],'Paul Allen':[46872.02]}
+    response = '0'
+    # menu_prompt
+    while not (response == '3'):
+        print('Menu')
+        print('1. Send a Thank You')
+        print('2. Create a Report')
+        print('3. Quit')
+        response = input("Please type number from menu > ")
+        if response == '1':
+            donors = sendingthank(donors)
+        if response == '2':
+            createreport(donors)
+
+
+if __name__ == "__main__":
+    main()
--- a/students/Mintae_K/lesson04/mailroom2.py
+++ b/students/Mintae_K/lesson04/mailroom2.py
@@ -39,14 +39,13 @@ def createreport(donors):
 def sendletters(donors):
     for keys in donors:
         s = "{}.txt".format(keys)
-        f = open(s,"w+")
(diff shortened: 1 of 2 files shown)


students/Mintae_K/lesson04/mailroom2.py
```diff
@@ -0,0 +1,69 @@
...
+    response = '0'
+    # menu_prompt
+    while not (response == '4'):
+        print('Menu')
+        print('1. Send a Thank You')
+        print('2. Create a Report')
+        print('3. Send letters to everyone')
+        print('4. Quit')
+        response = input("Please type number from menu > ")
+        switch_func_dict = {'1': sendingthank, '2': createreport, '3': sendletters}
```
natasha-aleksandrova: * this switch dict should be defined outside of your while loop; it is currently being redefined on each while loop iteration and is wasteful
* map "4" option to a function that quits the program so that you don't have to special case handle that option
* per style guide - write out your function names separated by undrescores -> `send_thank_you`

students/Mintae_K/lesson04/mailroom2.py
```diff
@@ -0,0 +1,69 @@
...
+    for i in range(len(total)):
+        print(s.format(total[i][2], total[i][0], total[i][1], total[i][0] / total[i][1]))
+    print(' ')
+    return
+
+
+def sendletters(donors):
+    for keys in donors:
+        s = "{}.txt".format(keys)
+        f = open(s, "w+")
```
natasha-aleksandrova: as mentioned in the course material, use `with` statement when working with files; it is the preferred way and ensures the files is closed when you're done

students/Mintae_K/lesson04/mailroom2.py
```diff
@@ -0,0 +1,69 @@
...
+    total.sort(reverse=True)
+    s = '{:<20}  $ {:<10.2f}  {:^12} $ {:<20.2f}'
+    for i in range(len(total)):
+        print(s.format(total[i][2], total[i][0], total[i][1], total[i][0] / total[i][1]))
+    print(' ')
+    return
+
+
+def sendletters(donors):
+    for keys in donors:
```
natasha-aleksandrova: best to name variable that matches closely to what it holds - in this case it should be either singular `key` or `donor`, not plural

students/Mintae_K/lesson04/mailroom2.py
```diff
@@ -0,0 +1,69 @@
...
+    total = []
+    for keys in donors:
+        num_gift = len(donors[keys])
+        total.append([sum(donors[keys]), num_gift, keys])
+    total.sort(reverse=True)
+    s = '{:<20}  $ {:<10.2f}  {:^12} $ {:<20.2f}'
+    for i in range(len(total)):
+        print(s.format(total[i][2], total[i][0], total[i][1], total[i][0] / total[i][1]))
+    print(' ')
+    return
```
natasha-aleksandrova: no need for `return` here

students/Mintae_K/lesson04/mailroom2.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+# Creating Report
+def createreport(donors):
+    s = '\n, {:<20}| {:<10} | {:<10}| {:<20}'
+    print(s.format('Donor Name', 'Total Given', "Num Gift", "Average Gift"))
+    print('-' * 60)
+    total = []
+    for keys in donors:
```
natasha-aleksandrova: it is preferred to use `items()` when you need both keys and values; why? as you are creating variables for key and value in a loop, you are assigning good variable names which help the reader.

for example
```
for donor, donations in donors.items()
```
or 
```
for name, donations in donors.items()
```
are clearer than `donors[key]`

students/Mintae_K/lesson04/mailroom2.py
```diff
@@ -0,0 +1,69 @@
+# Sending a Thank you
+def sendingthank(donors):
+    f_name = input('Please type the Full Name or list> ')
+    while f_name == 'list':
+        print(' ')
+        for x in donors:
+            print(x)
+        print(' ')
+        f_name = input('Please type the Full Name or list> ')
```
natasha-aleksandrova: move this to line 5 (right after while loop) and you can remove line 3.

students/Mintae_K/lesson04/mailroom2.py
```diff
@@ -0,0 +1,69 @@
...
+    if donors.get(f_name, False):
+        s = 'Existing donor, please type the donation amount from {}> '
+        amount = input(s.format(f_name))
```
natasha-aleksandrova: you can move this line and line 16 to above line 10; they are duplicated

students/Mintae_K/lesson04/mailroom2.py
```diff
@@ -0,0 +1,69 @@
+# Sending a Thank you
+def sendingthank(donors):
+    f_name = input('Please type the Full Name or list> ')
+    while f_name == 'list':
+        print(' ')
+        for x in donors:
+            print(x)
+        print(' ')
+        f_name = input('Please type the Full Name or list> ')
+    if donors.get(f_name, False):
```
natasha-aleksandrova: when using `.get` there is no need to provide the default value unless you're going to use that default value; in this case it isn't being used.

so you can just do `donors.get(f_name)` and `.get` will return None if it cannot find the key.

students/Mintae_K/lesson04/mailroom2.py
au-gold: the prompts (my text to display for user) are different for line 12 and line 16. Additionally, if statement is dictating which prompt to show. Thus cannot just simply move them outside of if statement.

students/Mintae_K/lesson04/mailroom2.py
au-gold: Can't remove line 3. Because then 'while' loop will fail since it is testing for 'f_name == 'list'. If line 3 is removed, f_name is undefined.

PR #983, opened 2018-09-24
Lesson 3

lauraannf:
All assignments for Lesson 03


students/lauraannf/Lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Sep 24 13:06:38 2018
+
+@author: Laura.Fiorentino
+"""
+DonorList = [
```
charlesrobison: Lowercase, separated by underscores:  donor_list =

students/lauraannf/Lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
...
+        'Frank Reynolds', [10, 20, 50], 
+        'Dee Reynolds', [25, 100], 
+        'Dennis Reynolds', [10, 50], 
+        'Mac McDonald', [25, 35, 20], 
+        'Charlie Kelly', [0.25]
+        ]
+
+def namefunc(Name):
```
charlesrobison: Same here: name_func(name):

students/lauraannf/Lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
...
+    while Name == 'list' or Name == 'List':
```
charlesrobison: You can use lower() here

students/lauraannf/Lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
...
+        if Task == '1':
+            Name = input("Type full donor name, or ''list'' for choices?>")
+            namefunc(Name)
+            Task = input("Choose an action: [1] Send a Thank You; [2] Create a Report; [3] Quit)>")
+        elif Task == '2':
+            reportfunc()
+            Task = input("Choose an action: [1] Send a Thank You; [2] Create a Report; [3] Quit)>")
+
+if __name__ == '__main__':
+    main()
```
charlesrobison: Nice work - the functions well!

students/lauraannf/Lesson03/strformat_lab.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Tue Sep 18 17:00:52 2018
+
+@author: Laura.Fiorentino
+"""
+print('----------Task One--------')
+testtuple = (2, 123.4567, 10000, 12345.67)
```
charlesrobison: Separate words with underscores per PEP8: test_tuple = for this as well as below

students/lauraannf/Lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
...
+    while Task not in ('1','2','3'):
+        Task = input("Choose an action: [1] Send a Thank You; [2] Create a Report; [3] Quit)>")
+    while Task !='3':  
+        if Task == '1':
+            Name = input("Type full donor name, or ''list'' for choices?>")
+            namefunc(Name)
+            Task = input("Choose an action: [1] Send a Thank You; [2] Create a Report; [3] Quit)>")
+        elif Task == '2':
+            reportfunc()
+            Task = input("Choose an action: [1] Send a Thank You; [2] Create a Report; [3] Quit)>")
```
natasha-aleksandrova: here is a modified entry point based on your code:
```
while True:
    task = input("choose an action: ...")
    if task == '1':
        send_thank_you()
    elif task == '2':
        create_report()
    elif task == '3':
        break
```
notice how much simpler the code is ^^^

here are the reasons:
* you have lines 43, 45, 50, and 53 all repeated - although you could use a variable for the string it is still an overkill; 
* line 48 should be moved inside of thank you functionality, in lesson4 it will be pretty much a requirement
* namefunc and reportfunc don't need "func" as part of it.. it is similar to naming "task_var" / "name_var"; it is known already that you are calling a function, therefore redundant.

lauraannf:
Does this work if I just hit reply no the email? We'll see I guess.

Thanks for the input. I definitely need to work on styling and syntax. I'm
so used to matlab, where none of that really matters, and I've always been
self-taught and rarely share my code, so it's not something I ever worry
about. So I really appreciate all of the advice.

students/lauraannf/Lesson03/mailroom.py
lauraannf: Thank you, I was having a difficult time with the "while" but I think it's starting to become more clear.

natasha-aleksandrova:
Yes you can reply via email but if you reply via github then you can add comments directly to our comments.

In your email you should see "view it on github" link at the bottom.

And yes totally make sense - it may be challenging at first to remember all of the styling rules.. but trust me quickly enough it becomes so easy you won't even think about it anymore 🙂. This is why I love Pycharm, it is much quicker to learn the styling this way.

students/lauraannf/Lesson03/mailroom.py
natasha-aleksandrova: 👍 

what helps me is to write the code first just to make it work and then iterate on it - by that I mean, look over it again and ask yourself: "is there a duplication?" "can the logic be simplified". Almost always it is the case.

PR #987, opened 2018-09-25
Update for Lesson 03

lauraannf:
I've redone lesson 3 according to advice. No need to review, mostly for my own benefit.

### DIFF
--- a/students/lauraannf/Lesson03/mailroom.py
+++ b/students/lauraannf/Lesson03/mailroom.py
@@ -4,53 +4,73 @@ Created on Mon Sep 24 13:06:38 2018
 
 @author: Laura.Fiorentino
 """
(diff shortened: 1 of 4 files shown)


students/lauraannf/Lesson03/mailroom.py
```diff
@@ -4,53 +4,73 @@
...
+        task = input("Choose an action: [1] Send a Thank You; [2] Create a \
+Report; [3] Quit>")
+        if task == '1':
+            create_email()
+        elif task == '2':
+            create_report()
+        elif task == '3':
+            break
+        else:
+            continue
```
natasha-aleksandrova: no need for continue here as it s redundant, your loop will continue looping on its own.

PR #990, opened 2018-09-27
Lesson 9 Ready for Review

srepking:
Thank you!

### DIFF
--- a/students/srepking/Lesson09/donors.py
+++ b/students/srepking/Lesson09/donors.py
@@ -35,15 +35,84 @@ class Group:
                     for some_name, donations in self._donor_raw.items()}
         return donors_f
 
+    @staticmethod
+    def column_name_width(donor_summary):
+        name_list = list(donor_summary.keys())  # creates a list of keys
+        name_wi = 11  # Establish minimum column width
+        for i in name_list:
+            if len(i) > name_wi:
+                name_wi = (len(i))  # width of name column
+        return name_wi
+
+    @staticmethod
+    def column_total_width(donor_summary):
+        tot_wi = 12
+        for name, summary in donor_summary.items():
+            if len(str(summary[0])) > tot_wi:
+                # width of total column
+                tot_wi = (len(str(summary[0]))) + 3
+                # width of number of donations column
+        return tot_wi
+
+    @staticmethod
+    def column_average_width(donor_summary):
+        ave_wi = 12
+        for name, summary in donor_summary.items():
+            if len(str(summary[2])) > ave_wi:
+                # width of total column
+                ave_wi = (len(str(summary[2]))) + 3
+                # width of number of donations column
+        return ave_wi
+
+    @staticmethod
+    def column_number_width(donor_summary):
+        num_wi = 12
+        for name, summary in donor_summary.items():
+            if len(str(summary[1])) > num_wi:
+                # width of total column
+                num_wi = (len(str(summary[1]))) + 3
+                # width of number of donations column
+        return num_wi
+
+    @staticmethod
+    def sort_list(donor_summary):
+        list_sorted = sorted(donor_summary, key=donor_summary.__getitem__, reverse=True)
+        return list_sorted
+
+    @staticmethod
+    def report(donor_summary):
+        """Return a report on all the donors"""
+        name_wi = Group.column_name_width(donor_summary)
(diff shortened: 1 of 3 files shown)


students/srepking/Lesson09/donors.py
```diff
@@ -0,0 +1,140 @@
+
+
+class Group:
+
+    def __init__(self):
+
+        self._donor_raw = {}
+
+    def get_donors(self):
```
natasha-aleksandrova: common practice is to use a decorator for property:
```
@property
def get_donors(self):
```

natasha-aleksandrova: however.. no need for decorator here if you're not applying any special logic to that attribute.

students/srepking/Lesson09/donors.py
```diff
@@ -0,0 +1,140 @@
...
+        return self._donor_raw
+
+    donors = property(get_donors)
+
+    def search(self, donor):
+        if self._donor_raw.get(donor) is None:
+            return
+        else:
+            return donor
```
natasha-aleksandrova: can be simplified to:
```
def search(self, donor):
    return self._donor_raw.get(donor)
```
it's identical to your code as far as results.

students/srepking/Lesson09/donors.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def add(self, donor, donation):
+        if self.search(donor) is None:
+            self._donor_raw[donor] = [donation]
+        else:
+            self._donor_raw[donor].append(donation)
+
+    def create_list(self):
```
natasha-aleksandrova: more appropriate name would be -> `def print_donors(self):`

students/srepking/Lesson09/donors.py
```diff
@@ -0,0 +1,140 @@
...
+    def column_name_width(donor_summary):
+        name_list = list(donor_summary.keys())  # creates a list of keys
+        name_wi = 11  # Establish minimum column width
+        for i in name_list:
+            if len(i) > name_wi:
+                name_wi = (len(i))  # width of name column
+        return name_wi
+
+    @staticmethod
+    def column_total_width(donor_summary):
```
natasha-aleksandrova: nice use of staticmethods

students/srepking/Lesson09/donors.py
```diff
@@ -0,0 +1,140 @@
...
+
+class Individual:
+    def __init__(self):
+        pass
+
+    @staticmethod
+    def thank_you(donor, donation):
+        """Add a donation to a donors records and print a report."""
+        return ('Thank you so much for the generous gift of ${0:.2f}, {1}!'
+              .format(float(donation), donor))
```
natasha-aleksandrova: This class should be utilized in your `Group` class.

The relation ship here is one to many; you have `Donor` and `DonorCollection` - ideally `DonorCollection` would reference many `Donor` objects.

Here is an example of what I mean:
https://repl.it/@naaleksandrova/Lesson9Example

students/srepking/Lesson09/donors.py
srepking: Yes! I fixed it and Thank you!

students/srepking/Lesson09/donors.py
srepking: I looked at your example showing how you made the relationship between single donor objects and the collection. Very nice! I incorporated this approach that you used, and learned from it. Thank you!

I deleted the @property for this def, but used it elsewhere.

students/srepking/Lesson09/donors.py
srepking: Updated my code.

srepking: I made the change.

students/srepking/Lesson09/donors.py
srepking: In your example, you also have the thank you letter in the Donor Class. Let me know if this comment still stands after my next pull request. I have incorporated your example, and now my DonorCollection is referencing many Donor objects. This is a much nicer way of doing this! Thank you!

students/srepking/Lesson09/donors.py
natasha-aleksandrova: glad it helped!

PR #993, opened 2018-09-29
Josh Hoff Grid Printer Assignment

josh1209:
Parts 1, 2 and 3 complete. Do you have any tips to make part 3 a bit cleaner? I feel like I could reduce the number of lines or at least simplify it.

### DIFF
--- a/students/Josh_HOff/lesson02/GridPrinter.py
+++ b/students/Josh_HOff/lesson02/GridPrinter.py
@@ -35,3 +35,42 @@ def print_grid(n):
         if mid == x + 1:
             print(plus_ + (dash_ * mid) + plus_ + (dash_ * mid) + plus_)
     print(plus_ + (dash_ * mid) + plus_ + (dash_ * mid) + plus_)
+
+#custom grid depending on number of rows/columns and grid cell size
+def print_grid2(r,n):
+    plus_ = '+ '
+    dash_ = '- '
+    line_ = '|'
+    spaces_ = ' ' * (n * 2 + 1)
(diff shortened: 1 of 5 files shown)


students/Josh_HOff/lesson02/GridPrinter.py
```diff
@@ -0,0 +1,76 @@
...
+        if mid == x + 1:
+            print(plus_ + (dash_ * mid) + plus_ + (dash_ * mid) + plus_)
+    print(plus_ + (dash_ * mid) + plus_ + (dash_ * mid) + plus_)
+
+#custom grid depending on number of rows/columns and grid cell size
+def print_grid2(r,n):
+    plus_ = '+ '
+    dash_ = '- '
+    line_ = '|'
+    spaces_ = ' ' * (n * 2 + 1)
```
natasha-aleksandrova: why use underscores in each variable?

students/Josh_HOff/lesson02/series.py
```diff
@@ -0,0 +1,74 @@
...
+        for x in range(2, n - 1):
+            if (x % 2 == 0):
+                num0 = num0 + num1
+            else:
+                num1 = num0 + num1
+        return num1 + num0
+    else:
+        #run other series here
+        #i'm not sure what series to add here - do I make up my own?
+        return 0
```
natasha-aleksandrova: the idea is to use the same formula for both series and fib series; therefore your sum_series shouldn't have separate code for each - you will only need to special case the first two values in the series if indexes 0 and 1 are requested.

PR #994, opened 2018-09-30
mailroom.py complete

jayjonesseattle:
Please grade mailroom.py. I appreciate any feedback you may have. thank you!


students/terrance_jones/lesson3/mailroom.py
```diff
@@ -0,0 +1,126 @@
...
+    while donorName == "list":
+        show_donor_list()
+        donorName =get_name()
+
+    if check_list(donorName)==True: 
+        donation_amount = get_amount()
+        add_to_existing_donor(donorName, donation_amount)
+        print_email(donorName, donation_amount)
+
+    elif check_list(donorName) == False:
```
natasha-aleksandrova: to simplify:

```
if check_list(donor_name):
else:
```
since it's a true false return you only need to check once.

however, it might be better to have a method `find_donor` so that it returns the donor reference - otherwise right now you're always searching the list twice, in check_list and add_to_existing_donor

students/terrance_jones/lesson3/mailroom.py
```diff
@@ -0,0 +1,126 @@
...
+        print("Good-bye")
+        SystemExit(0)
+    else:
+        print("Invalid entry")
+        menu()
+
+
+#thank you function gets a donors name and amount of donation. then prints out a thank you email
+def thankyou():
+    donorName = get_name()
```
natasha-aleksandrova: per pep8 style guide use snake_case style -> `donor_name`

students/terrance_jones/lesson3/mailroom.py
```diff
@@ -0,0 +1,126 @@
...
+    if response == "1":
+        thankyou()
+    elif response =="2":
+        create_report()
+    elif response == "3":
+        print("Good-bye")
+        SystemExit(0)
+    else:
+        print("Invalid entry")
+        menu()
```
natasha-aleksandrova: it is best to use `while True` loop for this entry point that way you don't have to manage the program flow outside of the main entry.

PR #1000, opened 2018-10-02
Mailroom_fp Submittal

tfbanks:


tfbanks:
Submitting lesson 10 assignment - Mailroom_fp file.
Note: I'm still reworking the object oriented file based on your feedback, I'm still running into a small difficulty with pulling in the report and mass mail letters, but I think I'm starting to figure out what I am missing.  I went ahead and worked lesson 10 material based on the last mailroom version prior to the object oriented lesson to make sure I got through all the lessons.  I'll continue to review the other file and resubmit at a later date.
Thanks.

PR #1003, opened 2018-10-03
lesson 7 html renderer

rbillupsii:


students/RoyC/Lesson07/html_render.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+    tag = "ul"
+    
+class Li(Element):
+    """
+    A list item tag element
+    """
+    tag = "li"
+    
+class TextEntry(Element):
```
natasha-aleksandrova: pep8: 2 lines space between classes

PR #1005, opened 2018-10-03
Lesson 10

andykhkwok:
Experimented with branching, but do not completely understand how it would be applicable locally unless I am working with multiple changes to different sections of the same file.

### DIFF
--- a/students/AndyKwok/lesson10/mailroom_fp.py
+++ b/students/AndyKwok/lesson10/mailroom_fp.py
@@ -1,0 +1,168 @@
+# Description: Mailroom Program - FP
+# Author: Andy Kwok
+# last Updated: 10/2/18
+# ChangeLog:
+# v1.0 - Initialization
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
In real life scenario, branching is used for "features" in a team setting (sometimes can help in solo projects too) .

Let's say we have a project (no forking involved) and in master branch we have the original mailroom version. I am tasked with adding try/except logic and you are tasked with adding send letters feature. We would have two separate branches: `add_try_except` and `add_send_letters`, we can work in our separate branches and then create a PR for each feature separately.

-----
In his class, here is what you get with branches: you can work on Lesson9 in `lesson9` branch and at the same time work on Lesson10 in `lesson10` branch. You can complete lesson 9 and create a PR while still working and adding commits to `lesson10` branch - this will not affect your PR for lesson 9 because it is in the isolated branch.

hopefully this helps.

PR #1010, opened 2018-10-04
modify mailroom_part3

pythonwa:
Hi Professor,
I re-submit mailroom_part3. Thanks.


students/tonylee/mailroom_part3.py
```diff
@@ -1,46 +1,49 @@
...
                     2 - Create a Report
                     3 - Send letters to everyone
                     4 - Quit\n''')
-        choice = int(choice)
-        if choice >=1 and choice < 4:
-            OptionSelect[choice]()
+            choice = int(choice)
+            if choice >=1 and choice < 4:
+              OptionSelect[choice]()
+        except ValueError:
```
natasha-aleksandrova: remember that it is best to put try/except around code that can throw that exception; in this case line 45.

PR #1011, opened 2018-10-04
adding solution to list_lab from lesson03

elnurad:


natasha-aleksandrova:
@elnurad don't forget to submit on the course website as well.

elnurad:
Thank you for reviewing and for the feedback. Will make sure to remember these style comments next time!

PR #1013, opened 2018-10-06
lesson03 HW submission - Mark Luckeroth

mluckeroth:
all Lesson03 exercises


students/mark_luckeroth/lesson03/mailroom.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+
+Names = ['Peter','Paul','Mary','Jake','Raj']
+Donations = [[10., 10., 10.],
+            [5., 50000., 5., 5.],
+            [100.],
+            [123., 456., 789.],
+            [60., 600000.]]
```
natasha-aleksandrova: per pep8 var names should use lower case snake_case -> `names`, `donations`

also, an alternate data structure:
```
donors = [('Peter', [10, 20, 30]), ('Paul', [29])]
```
it usually simplifies the code

students/mark_luckeroth/lesson03/mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+
+def Thanks():
```
natasha-aleksandrova: function names should also use snake_case lower case -> `thanks`

students/mark_luckeroth/lesson03/mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+    for item in sort_given:
+        index = Total_Given.index(item)
+        print(entry_str.format(*(Names[index], Total_Given[index], Num[index], avg[index])))
+
+def totals(Donations):
+    totals = []
+    for i, row in enumerate(Donations):
+        row_total = 0
+        for item in row:
+            row_total = row_total+item
```
natasha-aleksandrova: no need to compute this manually, you can just use `sum`

students/mark_luckeroth/lesson03/mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+    for row in Donations:
+        count.append(len(row))
+    return count
+
+def averages(totals, count):
+    return [x/y for x,y in zip(totals,count)]
+
+if __name__ == '__main__':
+    while True:
+        while True:
```
natasha-aleksandrova: you don't need the two while True loops, only one is needed.

PR #1015, opened 2018-10-06
submitted materials for Lesson 7

ptfbanks:
Materials represent a second effort.  I found this a more confusing exercise than instructive .  My response, having followed instructions appears much different from that of fellow students.  They have been able to greatly and neatly compress code.  I had hope to show work progression, but found myself conflicted between class specifications. I could not quite line up my indents correctly so my last sets of tests fail assertion tests.

### DIFF
--- a/students/ptfbanks/Lesson07/html_render.py
+++ b/students/ptfbanks/Lesson07/html_render.py
@@ -1,0 +1,152 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+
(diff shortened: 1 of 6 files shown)


ptfbanks:
I appear to have conflicted my git submittals for lessons 7 and 8.  The directories from which I would have submitted them were both empty, making me concerned I had loaded incomplete copies.  Additionally, my submittal of lesson 8 picked up a host of data that was not mine.

 I've tried to re-commit and merge the proper files. Please advise what may be the best correction.

PR #1019, opened 2018-10-08
Lesson 8

toddlikesbeer:
This one was actually .... fun, despite the fact that I was at work all weekend working on it.    :-)

This is ONLY for Session 8. I still need to finish 7.

### DIFF
--- a/students/toddschultz/session08/circle.py
+++ b/students/toddschultz/session08/circle.py
@@ -30,6 +30,7 @@ class Circle:
         if diameter <= 0:
             raise ValueError
         self = cls(diameter / 2)
+        self = self
         self.diameter = diameter
         return self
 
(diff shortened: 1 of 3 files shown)


students/toddschultz/session08/test_circle.py
```diff
@@ -0,0 +1,88 @@
...
+def test_diameter_2():
+	c = Circle(4)
+	c.diameter = 10
+	assert c.diameter is 10
+
+def test_add():
+	e = Circle(4)
+	f = Circle(6)
+	assert e.radius + f.radius is 10
+	assert e.diameter + f.diameter is 20
```
natasha-aleksandrova: `is` is used for object comparison; for example `is True` `is None` etc

here you should use `==` operators

students/toddschultz/session08/test_circle.py
```diff
@@ -0,0 +1,88 @@
...
+
+def test_ne():
+	t = Circle(8)
+	u = Circle(4)
+	assert t != u
+
+def test_sub():
+	v = Circle(2)
+	w = Circle(4)
+	assert w - v is 2
```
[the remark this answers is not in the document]
natasha-aleksandrova: your test should look something like:
```
    c1 = Circle(2)
    c2 = Circle(4)
    c3 = c2 + c1
    assert c3.radius == 2
```

students/toddschultz/session08/circle.py
```diff
@@ -0,0 +1,74 @@
...
+        return self._area
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        if diameter <= 0:
+            raise ValueError
+        self = cls(diameter / 2)
+        self = self
+        self.diameter = diameter
+        return self
```
natasha-aleksandrova: this can be simplified:
```
@classmethod
        if diameter <= 0:
            raise ValueError
        return cls(diameter / 2.0)
```

PR #1022, opened 2018-10-09
Completed Slicing Exercise in Lesson 3

josh1209:
Finished assert statements and used simplified code in all functions.

### DIFF
--- a/students/Josh_HOff/lesson03/list_lab.py
+++ b/students/Josh_HOff/lesson03/list_lab.py
@@ -1,0 +1,7 @@
+fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+def series1(fruits)
+    print(fruits)
+    response = input('Please add a fruit:')
+    fruits.append(response)
+    print(fruits)
+    response = input('Please indicate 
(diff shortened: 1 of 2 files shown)


students/Josh_HOff/lesson03/list_lab.py
```diff
@@ -0,0 +1,54 @@
...
+#    print(fruits)
+#bonus for series 2
+fruits = (fruits * 2)
+x = 1
+while x == 1:
+    response = input('Please delete a fruit: ')
+    for i in fruits:
+        if i == response:
+            fruits.remove(response)
+            x = 2
```
natasha-aleksandrova: the `x` seems mysterious 🙂 

if you would like to exit on the remove condition then this is a better pattern:
```
while True:
    ....
    if fruit == response:
        ...
        break
```
use `break`

students/Josh_HOff/lesson03/list_lab.py
```diff
@@ -0,0 +1,54 @@
...
+#series 3
+for i in fruits3:
+    x = 1
+    response = input('Do you like {}? '.format(i))
+    if response != 'yes' and response != 'no':
+        while x == 1:
+            response = input('Please choose "yes" or "no": ')
+            if response == 'yes' or response == 'no':
+                x = 2
```
natasha-aleksandrova: same here about the while loop pattern.

josh1209:
Cool, cool. I'll use that style for while loops in the future.

natasha-aleksandrova:
sounds good!

PR #1023, opened 2018-10-09
Lesson 8 assignment

ltrisolini:


students/ltrisolini/Lesson08/Circle.py
```diff
@@ -0,0 +1,94 @@
...
+import math
+
+class Circle:
+    def __init__(self, r):
+        self._r = r
+				
+    @classmethod
+    def from_diameter(self, d):
+        #return cls(d/2)
+        return self(d/2)
```
natasha-aleksandrova: use `cls` instead of `self` as that is a python pattern for classmethods.

students/ltrisolini/Lesson08/Circle.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # DUNDER METHOD
+    def __str__(self):
+        return "Circle with radius {}".format(self.radius)
+
+    def __repr__(self):
+        return "'Circle ({})'".format(self.radius)
+
+    
+    #comparison
```
natasha-aleksandrova: you'll want 1 line space between methods of the class

students/ltrisolini/Lesson08/Circle_Test.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python
+
+import Circle
+import unittest
+
+class Test_Circle(unittest.TestCase):
```
natasha-aleksandrova: class names use CamelCase naming convention -> `TestCircle`

students/ltrisolini/Lesson08/Circle_Test.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python
+
+import Circle
```
natasha-aleksandrova: module name which is equivalent to just a python file that ends in `*.py` should use all lower case naming; then you can:
`import circle` or `from circle import Circle`

it makes it clear if you're importing modules or classes

PR #1029, opened 2018-10-11
adding files

pythonwa:
I am submitting all the python files

### DIFF
--- a/students/tonylee/circle.py
+++ b/students/tonylee/circle.py
@@ -1,0 +1,55 @@
+import math
+
+class Circle:
+  def __init__(self, radius):
+      self._radius = radius
+      self._area = round(math.pi * radius ** 2, 6)
+
+  @property
+  def radius(self):
+      return self._radius
+
+  @property
+  def diameter(self):
+      return self._radius * 2
+
+  @diameter.setter
+  def diameter(self, diameter):
+      self._radius = diameter/2
+
+  @property
+  def area(self):
+    return self._area
+
+  @classmethod
+  def from_diameter(class_object, diameter):
+      return class_object(diameter/2)
+
+  def __str__(self):
+      return "Circle with radius: %.6f"% (self._radius)
+
+  def __repr__(self):
+      return "Circle({})".format(int(self._radius))
+
+  def __add__(self, other):
+      return Circle(self._radius + other._radius)
+
+  def __mul__(self, other):
+      return Circle(self._radius * other)
+
+  def __lt__(self, other):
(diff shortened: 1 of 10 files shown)


students/tonylee/circle.py
```diff
@@ -0,0 +1,55 @@
...
+  @diameter.setter
+  def diameter(self, diameter):
+      self._radius = diameter/2
+
+  @property
+  def area(self):
+    return self._area
+
+  @classmethod
+  def from_diameter(class_object, diameter):
```
natasha-aleksandrova: the convention is to name class instance as `cls` (instead of `class_object`)

students/tonylee/mailroom_oo.py
```diff
@@ -0,0 +1,110 @@
...
+    def add_donor(self, donor):
+        self.donors.append(donor)
+
+    def get_all_donor_names(self):
+        return [donor.full_name for donor in self.donors]
+
+    def thank_you(self):
+        donor_name = None
+        while not donor_name:
+            donor_name = name_input()
```
natasha-aleksandrova: remember that there should be no input functions in the data holder classes (even though it's not directly used here)

students/tonylee/test_mailroom_part4.py
```diff
@@ -0,0 +1,22 @@
+import unittest
+import os.path
+
+from mailroom_part4 import donation_history, thank_you_message, send_letter
+
+class TestMailroomPart4(unittest.TestCase):
+    def test_all_donor_names(self):
+        self.assertEqual(['Tony Lee', 'Michelle Cao', 'Andy Arko', 'Tom Ludwinski'], list(donation_history.keys()))
```
natasha-aleksandrova: this test is testing python functionality (like contents of keys)

your unit tests should test your code logic.

students/tonylee/test_mailroom_part4.py
```diff
@@ -0,0 +1,22 @@
...
+    def test_thankyou_message(self):
+        message = thank_you_message('Tony Lee')
+        self.assertEqual('Thank you Tony Lee for your donation of 100.0', message)
+
+    def test_send_letter(self):
+        send_letter()
+        assert os.path.isfile("Tony Lee.txt")
+        assert os.path.isfile("Michelle Cao.txt")
+        assert os.path.isfile("Andy Arko.txt")
+        assert os.path.isfile("Tom Ludwinski.txt")
```
natasha-aleksandrova: the other two tests look good!

PR #1037, opened 2018-10-12
Lesson 4

lauraannf:
I really want to keep working on the trigram assignment, but then I'll never move on, so I'm going to end it for now and hopefully work on it again later.  It should meet the assignment requirements for now though.


students/lauraannf/Lesson04/mailroom.py
```diff
@@ -0,0 +1,91 @@
...
+from pathlib import Path
+my_path = 'C:\\Users\\Laura.Fiorentino\\Documents\\MyPython\\\
+Self_Paced-Online\\students\\lauraannf\\Lesson04\\'
+donor_list = {
+    'Frank Reynolds': [10, 20, 50],
+    'Dee Reynolds': [25, 100],
+    'Dennis Reynolds': [10, 50],
+    'Mac McDonald': [25, 35, 20],
+    'Charlie Kelly': [0.25]
+    }
```
natasha-aleksandrova: 👍

students/lauraannf/Lesson04/mailroom.py
```diff
@@ -0,0 +1,91 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Sep 24 13:06:38 2018
+
+@author: Laura.Fiorentino
+"""
+import datetime
+from pathlib import Path
+my_path = 'C:\\Users\\Laura.Fiorentino\\Documents\\MyPython\\\
+Self_Paced-Online\\students\\lauraannf\\Lesson04\\'
```
natasha-aleksandrova: note that you can also get current path programmatically so you don't have to hard code

students/lauraannf/Lesson04/mailroom.py
```diff
@@ -0,0 +1,91 @@
...
+    new_path = my_path + 'letters_' + now.strftime('%d%m%Y')
+    Path(new_path).mkdir(exist_ok=True)
+    for key in donor_list.keys():
+        file = open(new_path + '\\' + key + '_' + now.strftime('%d%m%Y') +
+                    '.txt', 'w')
+        file.write('Dear {}, \n'
+                   'Thank you so much for your generous donation of ${:.2f} \n'
+                   'Sincerely, \n'
+                   'Laura F'.format(key, sum(donor_list[key])))
+        file.close()
```
natasha-aleksandrova: couple notes here:
* `file` is a python built-in so you don't want to overwrite it (notice github uses different color vs normal vars)
* use `with` statement when working with files which ensures to close the file.

students/lauraannf/Lesson04/mailroom.py
```diff
@@ -0,0 +1,91 @@
...
+    while True:
+        task = input("Choose an action: [1] Send a Thank You; [2] Create a \
+Report; [3] Send a Thank You to Everyone; [4] Quit>")
+        arg_dict = {'1': create_email, '2': create_report, '3': all_letters}
+        if task in arg_dict.keys():
+            arg_dict.get(task)()
+        elif task == '4':
+            break
+        else:
+            continue
```
natasha-aleksandrova: you should use a switch dict here for the main entry point.

students/lauraannf/Lesson04/write_a_book.py
```diff
@@ -0,0 +1,40 @@
...
+for it in range(len(word_list_new)):
+    # for it in range(300):
+    file_new.write(word_list_new[it] + ' ')
+    if word_list_new[it][-1] is '.' and len(word_list_new[it]) > 2 and \
+            word_list_new[it] not in ['Mr.', 'Dr.', 'Mrs.']:
+        n_period = n_period + 1
+    if n_period == 5:
+        file_new.write('\n     ')
+        n_period = 0
+file_new.close()
```
natasha-aleksandrova: nice work! couple notes:
* same here, use `with` with files
* would be best to break the code up into smaller functions to improve readability

students/lauraannf/Lesson04/mailroom.py
lauraannf: Can you explain this to me? I was having a hard time figuring out how using paths in python. 
For this I intended to name the path (so that I could it could be any folder). What would be the most efficient way to have written this?

students/lauraannf/Lesson04/mailroom.py
lauraannf: Can you explain this? Thanks!

students/lauraannf/Lesson04/mailroom.py
natasha-aleksandrova: My initial response was to use `os.getcwd()` which gives you current working directory. This would give you the correct path if you are always running it from Lesson4 directory.

If you want to run this from any directory then you need to use this:
```
os.path.dirname(os.path.abspath(__file__)
```

Note: I tested this on unix based system; sometimes windows is a little weird on working with file paths. 

But the idea here is that you can get the path of your current file programmatically.

students/lauraannf/Lesson04/mailroom.py
natasha-aleksandrova: @lauraannf f from assignment:
> See if you can use a dict to switch between the users selections. See Using a Dictionary to switch for what this means.

it has a link there that takes you to:
https://uwpce-pythoncert.github.io/PythonCertDevel/modules/DictionaryAsSwitch.html

please review the material and if it is still not clear let me know, thanks.

lauraannf: I just can't figure out how to quit or re-ask the question using the switch. Should I make them their own functions?

natasha-aleksandrova: this is what it would look like:
```
def quit_program():
    sys.exit()


switch_dict = {"1": send_thank_you, "4": quit_program}
while True:
    resp = input("...")
    switch_dict[resp]()
```
does that help?

PR #1044, opened 2018-10-14
String Formatting Lab Exercise 

josh1209:
Finished String Formatting Exercise


students/Josh_HOff/lesson03/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+if __name__ == "__main__":
+    donors_list = (['Ralph Anders', 5, 10], ['Andrei Wasinski', 101, 151, 75], ['Stalk Holmes', 40], ['Traci Johnston', 20], ['James Hendrick', 60], ['Angelica Kisel', 45, 25, 55.60])
+    while True:
+        print('')
+        response = input('1: Send a Thank You \n2: Create a Report \n3: Quit \nChoose an Option: ')
+        if response == '1':
+            donors_list = response1(donors_list)
+        if response == '2':
+            response2(donors_list)
+        if response == '3' or response == 'quit':
```
natasha-aleksandrova: you should be using if/elif/else statements here as only one choice can happen here

students/Josh_HOff/lesson03/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+                if a == sum(item[1:]):
+                    gift = len(item[1:])
+                    average = (a / gift)
+                    print(f'{item[0]:<23} ${a:>11.2f} {gift:>11}  ${average:>11.2f}')
+                    break
+        break
+   
+#main function
+if __name__ == "__main__":
+    donors_list = (['Ralph Anders', 5, 10], ['Andrei Wasinski', 101, 151, 75], ['Stalk Holmes', 40], ['Traci Johnston', 20], ['James Hendrick', 60], ['Angelica Kisel', 45, 25, 55.60])
```
natasha-aleksandrova: * tuple is **not** a good choice here for a data structure as tuples are *immutable*. something like this is better: `[("Bob", [299]), ("Kat": [399, 3999])]`
* you can define the `donors_list` at the module level (at the top of the file) so you don't have to pass it into each function to simplify

students/Josh_HOff/lesson03/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+    while True:
+        print('')
+        response = input('1: Send a Thank You \n2: Create a Report \n3: Quit \nChoose an Option: ')
+        if response == '1':
+            donors_list = response1(donors_list)
+        if response == '2':
+            response2(donors_list)
```
natasha-aleksandrova: prefer to use good readable function names that help the reader understand your code:
* `response1` -> `send_thank_you`
* `response2` -> `create_report`

students/Josh_HOff/lesson03/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+            a = max(temp)
+#            print(max(temp))
+            temp.remove(max(temp))
+            for item in donors_list:                
+                if a == sum(item[1:]):
+                    gift = len(item[1:])
+                    average = (a / gift)
+                    print(f'{item[0]:<23} ${a:>11.2f} {gift:>11}  ${average:>11.2f}')
+                    break
+        break
```
natasha-aleksandrova: your `while` loop *always* runs once - why use a while loop?

natasha-aleksandrova: the report should be sorted - please add the feature.

students/Josh_HOff/lesson03/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+                float(don_amt)
+                count = 0
+                for i in donors_list:
+                    if i[0] == response:
+                        donors_list[count].append(float(don_amt))
+                        email(response, don_amt)
+                        x = 2
+                        break
+                    count += 1
+                if x != 2:
```
natasha-aleksandrova: what is `x` and the significance of 2?

natasha-aleksandrova: here is an example of a more concise version using a list for a data structure:
```
def send_thank_you():
    while True:
        resp = input("Please type 'list', 'quit' or name of the donor")
        if resp == "list":
            list_donors()
        elif resp == "quit":
            return
        else:
            new_donation = input("Please enter amount")
            new_donor = False
            for donor_name, donations in donors_list:
                if donor_name == resp:
                    donations.append(new_donation)
                    break
            else:
                new_donor = True
                donors_list.append((resp, [new_donation]))
            
            print_email(new_donor)  # passed bool for custom message
            return
```

students/Josh_HOff/lesson03/mailroom.py
josh1209: Fixed!

students/Josh_HOff/lesson03/mailroom.py
josh1209: When I defined donors_list at the top, I was having issues using it in other functions. I'll try messing with it some more.

students/Josh_HOff/lesson03/mailroom.py
josh1209: Fixed!

students/Josh_HOff/lesson03/mailroom.py
josh1209: Removed the while loop. The report already sorts itself by total historical donation amount. Not very cleanly, though. I used a couple for loops to accomplish it.

students/Josh_HOff/lesson03/mailroom.py
josh1209: I was trying to avoid executing this block of code if the previous block of code was executed, so I used the variable x.

>                 for i in donors_list:
>                     if i[0] == response:
>                         donors_list[count].append(float(don_amt))
>                         email(response, don_amt)
>                         x = 2
>                         break
>                     count += 1
>                 if x != 2:
>                     donors_list += ([response],)
>                     donors_list[count].append(float(don_amt))
>                     email(response, don_amt)
>                 return(donors_list)
>

josh1209: Your block of code makes a LOT more sense and is much cleaner. I'll put it in comments in my code for future reference.

students/Josh_HOff/lesson03/mailroom.py
natasha-aleksandrova: 👍

students/Josh_HOff/lesson03/mailroom.py
natasha-aleksandrova: 👍

students/Josh_HOff/lesson03/mailroom.py
natasha-aleksandrova: hm shouldn't be having issues - if you still do reach out to me via MS Teams or email, I would like to take a look at what you have.

students/Josh_HOff/lesson03/mailroom.py
natasha-aleksandrova: Interesting - yeah I can't tell from the code 🙂 

I look for `sorted` or `sort` function calls - I would **highly** recommend using built-in functions. Very easy to use and readable.

natasha-aleksandrova:
👍 for leaving comments on github! this makes a great discussion on improving the code or asking questions.

students/Josh_HOff/lesson03/mailroom.py
josh1209: In mailroom2, I defined donors_list at the top like you suggested and didn't have any issues.

students/Josh_HOff/lesson03/mailroom.py
josh1209: I used the sorted function in mailroom2 like you suggested

students/Josh_HOff/lesson03/mailroom.py
josh1209: Nice, I understand now. If I need to use control flow like that in the future, I'll use boolean flags. I didn't have to use them in mailroom2.

natasha-aleksandrova:
glad it was helpful! you're welcome 🙂

PR #1045, opened 2018-10-14
Lesson 3: Slicing Lab Exercise and List lab Exercise

Pjurek3:
submission of Lesson 3 slicing lab exercise.  As general question, do you care if we do in file testing with pytest rather than just the assert statements?


natasha-aleksandrova:
@Pjurek3 your question:
> Do you want us to merge multiple exercises into one pull request?

You can create a single pull request for multiple exercises in a single Lesson, but it is often easier for us to review single exercise and give quicker feedback - however either works!

PR #1048, opened 2018-10-15
Made a couple of changes as per Natasha's feedback. 

elnurad:
Corrected indentation in "create_report" function and added docstrings to functions. Thank you for the feedback!


students/eldonab/lesson03/mailroom.py
```diff
@@ -1,25 +1,35 @@
...
 # to add a new donor name&donation amount:
 def add_new_donor(new_donor):
+    """Add a new donor name and donation amount."""
     amount = float(input("How much is the new donation amount? "))
     donors_list.append([new_donor, amount])
     
+
 # to check whether donor already exists in the list.
 def is__in_list(donor_name):
+    """Check whether donor already exists in the list of donors."""
```
natasha-aleksandrova: much better! (because of your docstrings you no longer need comments above the function definition.

elnurad:
Thanks!

PR #1054, opened 2018-10-17
Mailroom2 Completed

josh1209:
Need to find out how to choose a directory to save the files to, and how to ask a user for a path to save the files to.

### DIFF
--- a/students/Josh_HOff/lesson04/mailroom2.py
+++ b/students/Josh_HOff/lesson04/mailroom2.py
@@ -1,6 +1,6 @@
-arg_dict = {0:'zero', 1:'one', 2: 'two'}
 donors_list = {'Ralph Anders': [5, 10], 'Andrei Wasinski': [101, 151, 75], 'Stalk Holmes': [40], 'Traci Johnston': [20], 'James Hendrick': [60], 'Angelica Kisel': [45, 25, 55.60]}
-donor = 'Ralph Anders'
+donor = ''
+donation = 0
 
 def send_email(donor):
     donation = sum(donors_list.get(donor))
@@ -8,11 +8,52 @@ def send_email(donor):
 
 def send_thank_you():
     print('thank you')
-
-#    print((donors_list['Ralph Anders'])
+    while True:
+        print('')
+        donor = input('Type a Full Name: ')
+        if switch_func_dict.get(donor, continue_func)() == False:
+            return
+        elif donor != 'list':
+            print('')
+            donation = input('What is the donation amount?: ')
+            if donation == 'quit':
+                return
+            donation = float(donation)
+            if donor in donors_list:
+                donors_list[donor] += [donation]
+            else:
+                donors_list[donor] = [donation]
+            print(donors_list)
+            print('')
+            print(f'Hello, {donor}! Thank you very much for your generous donation of ${donation}! Your contribution is essential and will be well utilized.')
+            
+            
+def print_list():
+    print('')
+    for i in donors_list:
+        print(i)
+    return True
     
 def create_report():
     print('created report')
+    print('')
+    y = '|'
+    print(f'Donor Name{y:>14} Total Given {y} Num Gifts {y} Average Gift')
+    print('-' * 63)
+    
+    
+    donation = sum(donors_list.get(donor))
+
+    
+    
+    
+    
+    
+    
+    
+    
+    
+    
     
 def quitting():
     print('time to quit')
@@ -21,11 +62,15 @@ def quitting():
 def letters_to_everyone():
     print('create txt files with thank you letters in each one')
     
-switch_func_dict = {1:send_thank_you, 2:create_report, 3:letters_to_everyone, 4:quitting}
-
+def continue_func():
+    print('continuing the function')
+    return True
+    
+switch_func_dict = {'1':send_thank_you, '2':create_report, '3':letters_to_everyone, '4':quitting, 'quit':quitting, 'list':print_list}
 #main function
 if __name__ == "__main__":
     while True:
-        a = int(input('1: Send a Thank You \n2: Create a Report \n3: Send Letters to Everyone \n4: Quit \n Choose an Option: '))
-        if switch_func_dict.get(a)() == False:
+        send_thank_you_set = set(donors_list)
+        a = input('1: Send a Thank You \n2: Create a Report \n3: Send Letters to Everyone \n4: Quit \n Choose an Option: ')
+        if switch_func_dict.get(a, continue_func)() == False:
             break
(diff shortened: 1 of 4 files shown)


students/Josh_HOff/lesson04/mailroom2.py
```diff
@@ -0,0 +1,79 @@
+import pathlib
+import os
+donors_list = {'Ralph Anders': [5, 10], 'Andrei Wasinski': [101, 151, 75], 'Stalk Holmes': [40], 'Traci Johnston': [20], 'James Hendrick': [60], 'Angelica Kisel': [45, 25, 55.60]}
```
natasha-aleksandrova: 👍

students/Josh_HOff/lesson04/mailroom2.py
```diff
@@ -0,0 +1,79 @@
...
+
+#this function will continue the previous while loop and prevents errors if a user enters an unexpected input.
+def continue_func():
+    return True
+    
+switch_func_dict = {'1':send_thank_you, '2':create_report, '3':letters_to_everyone, '4':quitting, 'quit':quitting, 'list':print_list}
+#main function
+if __name__ == "__main__":
+    while True:
+        send_thank_you_set = set(donors_list)
```
natasha-aleksandrova: doesn't look like this is being used?

students/Josh_HOff/lesson04/mailroom2.py
```diff
@@ -0,0 +1,79 @@
...
+        a = input('1: Send a Thank You \n2: Create a Report \n3: Send Letters to Everyone \n4: Quit \nChoose an Option: ')
+        if switch_func_dict.get(a, continue_func)() == False:
```
natasha-aleksandrova: to make this a little cleaner so that you don't have to add a bool return to each function just to do this check, change your `quitting` function:
```
def quitting():
    sys.exit()   # or quit()
```

Then you can get rid of all of the returns and remove the check for False. why is this important? Imagine you're on a team and you guys continue to add features to the mailroom - each time you add a feature, a dev has to remember to add the dummy return True to their functions.

students/Josh_HOff/lesson04/mailroom2.py
```diff
@@ -0,0 +1,79 @@
...
+def quitting():
+    return False
+
+#this function creates files for all donors in the donor list.
+def letters_to_everyone():
+#    pth = pathlib.Path('./donations/')
+#    p = pth.absolute()
+#    print(p)
+    for i, val in donors_list.items():
+        outfile = open(f'{i}.txt', 'w')
```
natasha-aleksandrova: prefer to use `with` statement when working with files, it ensures that the file is always closed automatically

students/Josh_HOff/lesson04/mailroom2.py
```diff
@@ -0,0 +1,79 @@
...
+        elif donor != 'list':
+            print('')
+            donation = input('What is the donation amount?: ')
+            if donation == 'quit':
+                return
+            donation = float(donation)
+            if donor in donors_list:
+                donors_list[donor] += [donation]
+            else:
+                donors_list[donor] = [donation]
```
natasha-aleksandrova: 👍

students/Josh_HOff/lesson04/mailroom2.py
```diff
@@ -0,0 +1,79 @@
...
+    print(f'Donor Name{y:>14} Total Given {y} Num Gifts {y} Average Gift')
+    print('-' * 63)
+    temp_set = donors_list.copy()
+    for i, val in donors_list.items():
+        temp_set[i] = sum(val)        
+    temp_set = dict(sorted(temp_set.items(), key=lambda kv: kv[1], reverse=True))
+    for i, val in temp_set.items():
+        gift = len(donors_list[i])
+        average = ( val / gift)
+        print(f'{i:<23} $ {val:>11.2f} {gift:>11} {average:>11.2f}')
```
natasha-aleksandrova: 👍 on using sorted

we can simplify the code further, replace lines 41 through 45 with:
```
    # no need for temp list to compute the sum, you can do it on the fly here
    sorted_donors = sorted(donors_list.items(), key=lambda k: sum(k[1]), reverse=True)
    # now loop directly over the sorted structure
    for donor_name, donations in sorted_donors:
        pass
```

students/Josh_HOff/lesson04/mailroom2.py
josh1209: My mistake, I'll remove it

students/Josh_HOff/lesson04/mailroom2.py
josh1209: That makes sense. I imported sys, updated quitting function, and removed the True/False returns.

students/Josh_HOff/lesson04/mailroom2.py
josh1209: Got it. I'll use that in the next assignments.

students/Josh_HOff/lesson04/mailroom2.py
josh1209: Got it, I've implemented it in my code.

students/Josh_HOff/lesson04/mailroom2.py
josh1209: I had to implement another "If" statement to send the user back to the main menu when they use 'quit' when asked for a name.

> def send_thank_you():
>     while True:
>         print('')
>         donor = input('Type a Full Name: ')
>         **if donor == 'quit':
>             return**
>         switch_func_dict.get(donor, continue_func)()
>

natasha-aleksandrova: hm shouldn't have to, it's a little hard to read without formatting (use triple back ticks for code formatting on github).

your main loop is a while True so if you end your function it will go back to the main menu on its own.

josh1209:
```
def send_thank_you():
    while True:
        print('')
        donor = input('Type a Full Name: ')
        if donor == 'quit':
            return
        switch_func_dict.get(donor, continue_func)()
```
if the user types in 'quit', it still goes to this function and exits the program entirely (unless I keep the "if" statement). 
```
def quitting():
    sys.exit()
```

natasha-aleksandrova: do you have the latest version of your code you can link me? I can take a look at it - easier when I see the full code.

josh1209: Sure, I just pushed it.
https://github.com/josh1209/Self_Paced-Online/blob/master/students/Josh_HOff/lesson04/mailroom2.py

natasha-aleksandrova: I see now.

your `switch_func_dict` is for your main entry point. 

```
if response == 'quit':
    return
elif response == 'list':
    print_list()
else:
    # do add update/add donor stuff
```

I think you should keep the two flows separate: main entry point and send thank you flow

You could use switch dict in the send thank you flow as well but it is not required.

PR #1063, opened 2018-10-21
mailroom part 2 complete

jayjonesseattle:
hello. I think I made all the suggested changes to mailroom part 2. please take a look at it when you have a moment. thank you


students/terrance_jones/lesson4/mailroom_part2.py
```diff
@@ -107,16 +98,17 @@ def create_report():
...
-        filename = "{}.txt".format(item['full_name'])
+    for key in donor_list.keys():
+        filename = "{}.txt".format(key)
         with open(filename, 'w') as f:
-            donor_letter = letter(item)
+            donor_letter = letter(key)
             f.write(donor_letter)
- 
+    print("Letters have been created for each donor")
+    menu_selection(sub_prompt, sub_dispatch)
```
natasha-aleksandrova: you shouldn't be calling these manually, you should have a `while True` loop at the main entry point that controls the main program flow.

PR #1068, opened 2018-10-22
added the assignment of slicing

AlyssaHong:
Added the assignment of slicing


students/AlyssaHong/lesson03/list_lab.py
```diff
@@ -0,0 +1,94 @@
...
+fruits = ["Apples","Pears", "Oranges","Peaches"]
+print(fruits)
+
+# Ask the user for another fruit and add it to the end of the list.
+response = input("What would you like to add a fruit? ")
+fruits.append(response)
+print(fruits)
+
+# Ask the user for a number
+response = input("What number of fruit would you like to see(the number is 1~5 and you can choose just one)? ")
```
charlesrobison: What if the list is longer than 5?

students/AlyssaHong/lesson03/list_lab.py
```diff
@@ -0,0 +1,94 @@
...
+# Add another fruit -- using "+"
+print(["Mangos"] + fruits)
+
+# Add another fruit -- using "insert()"
+fruits.insert(0, 'Blueberries')
+print(fruits)
+
+# Display all the fruits that begin with “P”
+for x in fruits:
+    if x.startswith('P'):
```
charlesrobison: Good idea to use lower() here, just in case a user inputs lowercase fruits :)

students/AlyssaHong/lesson03/list_lab.py
```diff
@@ -0,0 +1,94 @@
...
+# Make a copy of the list and reverse the letters in each fruit in the copy
+fruits = ["Apples","Pears", "Oranges","Peaches"]
+original_fruits = fruits
+copy_fruits = []
+for fruit in original_fruits:
+    # copy_fruits = ''.join(reversed(fruit))
+    # print(copy_fruits)
+    # copy_fruits.append(copy_fruits)
+    copy_fruits.append(''.join(reversed(fruit)))
+## print(','.join(copy_fruits))
```
charlesrobison: Nice!

students/AlyssaHong/lesson03_1/slicing.py
```diff
@@ -0,0 +1,64 @@
...
+    seq[0], seq[-1] = a_new_sequence
+    print("result #1:", seq)
+    return exchange_first_last
+
+#2 with every other item removed.
+def every_other_item_remove(seq):
+    count_seq = len(seq)
+    del seq[1:int(count_seq-1)]
+    print("result #2:", seq)
+    return every_other_item_remove
```
charlesrobison: This one doesn't quite work - passing in 1,2,3,4,5 returns only 1,5

students/AlyssaHong/lesson03_1/slicing.py
```diff
@@ -0,0 +1,64 @@
...
+
+#3 with the first 4 and the last 4 items removed, and then every other item
+# in between.
+def select_item_and_remove(seq):
+    # print(seq[4:-4])
+    del seq[3]
+    del seq[-4]
+    del seq[4:-3]
+    print("result #3:", seq)
+    return select_item_and_remove
```
charlesrobison: This one too - if you pass in 1,2,3,4,5,6,7,8,9,10 you would expect 1-4 to be removed, as well as 7-10, then return every other between the 2.  So, the return would be 5

students/AlyssaHong/lesson03/list_lab.py
```diff
@@ -0,0 +1,94 @@
...
+    response = input("Do you like " + fruits[fruit_cnt].lower() + "? yes/no > ")
+    if response in yes_choice:
+        fruit_cnt+=1
+        continue
+    elif response in no_choice:
+        fruits.remove(fruits[fruit_cnt])
+        continue
+    else:
+        print("Please input yes/no")
+        continue
```
natasha-aleksandrova: you don't need any of the `continue` statements here

students/AlyssaHong/lesson03_1/slicing.py
```diff
@@ -0,0 +1,64 @@
...
+Lesson3 Assignments > Slicing Lab Exercise
+"""
+#Get the basics of sequence slicing downself.
+
+#1 with the first and last items exchanged.
+def exchange_first_last(seq):
+    a_new_sequence = seq[-1], seq[0]
+    seq[0], seq[-1] = a_new_sequence
+    print("result #1:", seq)
+    return exchange_first_last
```
natasha-aleksandrova: not using slicing here.

Here is what it should look like:
```
def exchange_first_last(seq):
    """with the first and last items exchanged"""
```

natasha-aleksandrova:
@AlyssaHong I would recommend re-working the slicing assignment to ensure you have a good grasp on slicing concepts as well as how functions work.

PR #1070, opened 2018-10-23
Lesson 2 folder added print_grid.py Assignment

njschafi:
Added Lesson 2 assignment print_grid.py


students/njschafi/Lesson2/FizzBuzz.py
```diff
@@ -0,0 +1,13 @@
+## NEIMA SCHAFI, LESSON 2 Assignment - FizzBuzz
+
+def fizzbuzz():
+    for i in range(100):
```
natasha-aleksandrova: you'll want `range(1, 101)` to get 1 through 100 numbers without having to increment manually

PR #1072, opened 2018-10-23
updated flow controls to mailroom_part2.py 

jayjonesseattle:
I think this solved the flow control problem. I will try harder to catch this oversight in my code going further.

### DIFF
--- a/students/terrance_jones/lesson4/mailroom_part2.py
+++ b/students/terrance_jones/lesson4/mailroom_part2.py
@@ -73,7 +73,7 @@ def letter(donor):
 def thankyou():
     donor_name = get_name()
     if donor_name == "list":
-        show_donor_list()
+        show_donor_list
         donor_name =get_name()
 
     if find_donor(donor_name) is True: 
@@ -82,14 +82,14 @@ def thankyou():
         
         dic = {'name': donor_name , 'amount_donated': donation_amount}
         print("Dear {name}, \n Thank you for your {amount_donated:,.2f} donation. \nSincerely, Mailroom.".format(**dic))
-        menu_selection(sub_prompt, sub_dispatch)
+        
     else:
         
         donation_amount = get_amount()
         create_new_donor(donor_name, donation_amount)
         dic = {'name': donor_name , 'amount_donated': donation_amount}
         print("Dear {name}, \n Thank you for your {amount_donated:,.2f} donation. \nSincerely, Mailroom.".format(**dic))
-        menu_selection(sub_prompt, sub_dispatch)
+       
             
 
 
@@ -99,7 +99,7 @@ def create_report():
     print(heading)
     print("-" * len(heading))
     print_rows()
-    menu_selection(sub_prompt, sub_dispatch)
+   
 
 def letters_everyone():
     for key in donor_list.keys():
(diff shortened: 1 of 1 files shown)


students/terrance_jones/lesson5/mailroom_part3.py
```diff
@@ -0,0 +1,169 @@
...
+def show_donor_list():
+    """Create list of donor names only. Display list of all donors"""
+    for key in donor_list.keys():
+        print(key)
+    
+
+def find_donor(x):
+    for key in donor_list.keys():
+        if key == x:
+            return True
```
natasha-aleksandrova: no need for a loop; dictionaries are very efficient with key look ups -> `if key in donor_list` will have the same effect.

students/terrance_jones/lesson5/mailroom_part3.py
```diff
@@ -0,0 +1,169 @@
...
+    name = input("Enter the full name of the donor.")
+    return name
+
+def show_donor_list():
+    """Create list of donor names only. Display list of all donors"""
+    for key in donor_list.keys():
+        print(key)
+    
+
+def find_donor(x):
```
natasha-aleksandrova: a better function name would be `donor_exists` since you're returning a boolean not actually finding and returning the donor.

students/terrance_jones/lesson5/mailroom_part3.py
```diff
@@ -0,0 +1,169 @@
...
+        
+        dic = {'name': donor_name , 'amount_donated': donation_amount}
+        print("Dear {name}, \n Thank you for your {amount_donated:,.2f} donation. \nSincerely, Mailroom.".format(**dic))
+        
+    else:
+        
+        donation_amount = get_amount()
+        create_new_donor(donor_name, donation_amount)
+        dic = {'name': donor_name , 'amount_donated': donation_amount}
+        print("Dear {name}, \n Thank you for your {amount_donated:,.2f} donation. \nSincerely, Mailroom.".format(**dic))
```
natasha-aleksandrova: lines 81-85 and 89-92 are identical except add_to_existing_donor/create_new_donor function calls - you generally want to consolidate duplicate code, how about:
```
donation_amount = get_amount()
add_or_update_donor(donor_name, donor_amount)  # let this function do the appropriate work
dic = {'name': donor_name , 'amount_donated': donation_amount}
print("Dear {name}, \n Thank you for your {amount_donated:,.2f} donation. \nSincerely, Mailroom.".format(**dic))
```

students/terrance_jones/lesson5/mailroom_part3.py
```diff
@@ -0,0 +1,169 @@
...
+   
+
+def menu_selection(prompt, dispatch_dict):
+    """displays the menu. asks for user to select option. if option is not available will throw KeyError and ask the user for new input"""
+    while True:
+        try:
+            response = input(prompt)
+            if dispatch_dict[response]() == "exit menu":
+                return
+        except KeyError:
```
natasha-aleksandrova: nice - it is best to put the exception you're catching only around the code that can throw it (which would wrap around lines 119-120)

PR #1079, opened 2018-10-25
updated mailroom part 3 - please take a look when you have a moment

jayjonesseattle:
Thank you for the awesome suggestion!  I think the new code reflects all of your suggested changes.

### DIFF
--- a/students/terrance_jones/lesson5/mailroom_part3.py
+++ b/students/terrance_jones/lesson5/mailroom_part3.py
@@ -16,11 +16,9 @@ def show_donor_list():
         print(key)
     
 
-def find_donor(x):
-    for key in donor_list.keys():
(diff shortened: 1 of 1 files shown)


students/terrance_jones/lesson5/mailroom_part3.py
```diff
@@ -16,11 +16,9 @@ def show_donor_list():
...
     
 
-def find_donor(x):
-    for key in donor_list.keys():
-        if key == x:
-            return True
-        
+def donor_exists(x):
+    if x in donor_list.keys():
+        return True
```
natasha-aleksandrova: even more simplified: 
```
def donor_exists(x):
    return x in donor_list
```

PR #1082, opened 2018-10-26
Lesson 9 complete

rbillupsii:


students/RoyC/Lesson09/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+        return self.ltr_template.format(**record)
+        
+    def thank_all(self):
+        """
+        Write thank you letters for all donors to separate files
+        """
+        for donor in self._donors:
+            f = open(donor.name.replace(' ',  '_') + ".txt", 'w')
+            last_donation = donor.donations[-1:][0]
+            f.write(self.send_thanx(donor, last_donation))
```
natasha-aleksandrova: it is preferred to use `with` statement when working with files.

PR #1088, opened 2018-10-28
lesson06

au-gold:


students/Mintae_K/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,22 @@
...
+
+donors = {'William Gates, III': [54842.49, 48965.25],
+            'Mark Zuckerberg': [7852.25, 48652.0, 3548.0],
+            'Jeff Bezos': [5486.0, 58794.02, 7412.1],
+            'Paul Allen': [46872.02]}
+
+
+def test_check_name_exist():
+    (yes_exist, s) = check_name_exist('Starbucks Coffee', donors)
+    assert 0 is yes_exist
```
natasha-aleksandrova: `is` comparison is used for objects (like None, True/False)

use `==` operator for comparison of ints/strs etc.

students/Mintae_K/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,22 @@
+from mailroom4 import *
```
natasha-aleksandrova: it is bad practice to use "start" imports

students/Mintae_K/lesson06/test_mailroom.py
natasha-aleksandrova: why aren't you asserting on `s`? and best to use good variable name so that the reader can understand the code, `found_message` is better than `s`

students/Mintae_K/lesson06/mailroom4.py
```diff
@@ -0,0 +1,98 @@
...
+        for x in donors:
+            print(x)
+        print(' ')
+        f_name = input('Please type the Full Name or list> ')
+    (yes_exist, s) = check_name_exist(f_name, donors)
+    myexit = 0
+    while not myexit == 1:
+        amount = input(s.format(f_name))
+        myexit = check_error(amount)
+    append_donation(f_name, amount, yes_exist, donors)
```
natasha-aleksandrova: it is better to catch your ValueError here and get rid of the `my_exit` var and exit when you catch the exception.

students/Mintae_K/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_check_name_exist():
+    (yes_exist, s) = check_name_exist('Starbucks Coffee', donors)
+    assert 0 is yes_exist
+
+
+def test_check_error():
+    x = check_error('not_a_number')
+    assert x is 0
```
natasha-aleksandrova: you don't need this test; see above about this function not being needed.

students/Mintae_K/lesson06/mailroom4.py
```diff
@@ -0,0 +1,98 @@
...
+def send_letters():
+    for name in donors:
+        s = "{}.txt".format(name)
+        with open(s, "w+") as f:
+            f.write("Dear {},".format(name))
+            f.write("\n\n     Thank you for your very kind donation of ${:<10.2f}".format(sum(donors[name])))
+            f.write('\n\n     It will be put to very good use.')
+            f.write('\n\n             Sincerely,')
+            f.write('\n                   -The Team')
+        f.closed
```
natasha-aleksandrova: no need to close when you use `with` statement, it does it for you.

students/Mintae_K/lesson06/mailroom4.py
```diff
@@ -0,0 +1,98 @@
...
+        return 0, s
+
+
+def append_donation(f_name, amount, yes_exist, donors):
+    #global donors
+    if yes_exist == 1:
+        donors[f_name].append(float(amount))
+    else:
+        donors[f_name] = [float(amount)]
+    return donors
```
natasha-aleksandrova: why is this being returned? you're not using the return in your code.

PR #1096, opened 2018-10-30
adding all of lesson 09

hood821:
added all of lesson 09, i got everything to work right through testing in the console, didn't time to finish my test_mailroom, unfortunately. But seems to work OK!

### DIFF
--- a/students/TressaHood/lesson09/mailroom_donors.py
+++ b/students/TressaHood/lesson09/mailroom_donors.py
@@ -1,0 +1,99 @@
+"""This is the file that holds the donation database, and has all the
+classes for Donor, and DonorDonations
+"""
+
+
+class Donor(object):
+    """This is the Donor class where we set up the functions that create a use-able database
+    We initialize it, get the key name (first/last name) and the value (donations).
+    This is used to hold the information about a SINGLE donor, so we can get name, donations
+    and add, as well as the evaluation needed for sorting etc.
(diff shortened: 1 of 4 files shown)


students/TressaHood/lesson09/mailroom_donors.py
```diff
@@ -0,0 +1,99 @@
...
+            l.append(d.get_name_as_key())
+
+        return(l)
+
+    def list_donors(self):
+        """This function lists all the donors in a database
+        :return donor_list: Returns a list of all the names in the donor database
+        """
+        for d in self.donors:
+            print(d.get_name_as_str())
```
natasha-aleksandrova: best to follow common patterns:

* name the function `get_name` (no need to be explicit about "as_str" it's implied that name will be a string)
* or define `__str__` method on the Donor object

students/TressaHood/lesson09/mailroom_donors.py
```diff
@@ -0,0 +1,99 @@
...
+    def __init__(self, db=None):
+        if db is None:
+            self.donors = []
+        else:
+            self.donors = db
+
+    def add_new_donor(self, new_donor):
+        self.donors.append(new_donor)
+
+    def add_to_database(self, donor, donation):
```
natasha-aleksandrova: `update_donor` ?

PR #1098, opened 2018-10-30
cleaned up extra blank lines and added assertion statements

april301975:
I totally did not see the mentioning of assertion!


april301975:
When I submitted the fizzbuzz and series, I swapped the files and the pull
request URL.

PR #1099, opened 2018-10-30
Trigrams - Josh Hoff

josh1209:
Finished the Trigrams program. Made it so longer texts could be fed into the program without endlessly printing out text.


students/Josh_HOff/lesson04/mailroom2.py
```diff
@@ -39,18 +41,18 @@ def create_report():
...
-        gift = len(donors_list[i])
-        average = ( val / gift)
-        print(f'{i:<23} $ {val:>11.2f} {gift:>11} {average:>11.2f}')
+    sorted_donors = sorted(donors_list.items(), key=lambda k: sum(k[1]), reverse=True)
+    for donor_name, donations in sorted_donors:
+#        print(len(donations))
+        gift = len(donations)
+        average = ( sum(donations) / gift)
+        print(f'{donor_name:<23} $ {sum(donations):>11.2f} {gift:>11} {average:>11.2f}')
+
```
natasha-aleksandrova: looks better!

students/Josh_HOff/lesson04/mailroom2.py
```diff
@@ -39,18 +41,18 @@ def create_report():
...
     print('')    
 
 #this function quits the previous menu.    
 def quitting():
-    return False
+    sys.exit()
```
natasha-aleksandrova: nice!

students/Josh_HOff/lesson04/trigrams.py
```diff
@@ -0,0 +1,69 @@
...
+                number = random.choice(range(len(newdict[b])))
+                a += f' {newdict[b][number]}'
+                del newdict[b][number]
+            else:
+                number = 0
+                a += f' {newdict[b][number]}'
+#            if b not in newdict:
+#                break
+#            else:
+#                a += f' {newdict[b][number]}'
```
natasha-aleksandrova: recommendation would be to split up the code into smaller functions; what's nice about that is it can pretty much replace your comments with self-descriptive functions names and makes it easier to read the code.

PR #1104, opened 2018-11-01
Mailroom 1 Pull Request

JonCracolici:
I incorporated your comments about snake casing, breaks and if statements. I did not incorporate your comment about using numbers to as the user input for control flow, but only because I am behind and catching up. I am using that method for the Mailroom 2 program I am writing now.


natasha-aleksandrova:
@JonCracolici there were conflicts here which I resolved.

What I think happened is you either use different computers or you had two locations of the cloned repo.. because somehow you made changes to the code that wasn't pulled-in in your clone, and this is your directory/code so only you would have made changes.

Let's keep an eye on it and make sure to merge with upstream before you start making changes.

PR #1106, opened 2018-11-01
updating lesson 6 assignments

jayjonesseattle:
I added test for both the create report and letter portions of mailroom part 4. I had to make some slight changes to the mailroom file and the test mailroom file but I think It works. As always, I welcome your feedback and thank you for the encouragement.

### DIFF
--- a/students/terrance_jones/lesson6/test_mailroom_part4.py
+++ b/students/terrance_jones/lesson6/test_mailroom_part4.py
@@ -4,8 +4,13 @@ from mailroom_part4 import donor_list
 from mailroom_part4 import donor_exists
 from mailroom_part4 import add_or_update_donor
 from mailroom_part4 import thank_you_letter
+from mailroom_part4 import letter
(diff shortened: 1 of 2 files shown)


students/terrance_jones/lesson6/test_mailroom_part4.py
```diff
@@ -20,3 +25,34 @@ def test_add_or_update_donor():
...
+    Thank you,\n
+    -The Mailroom'''
+
+    assert letter('Esperanza Gomez') == expected
+
+
+    def test_get_report():
+        donor_list = {'Jay Jones': [2,2,2]}
+        expected = "{:<25} ${:^10} {:^10} ${:^10}".format('Jay Jones', 6, 3, 2)
+        assert get_report(donor_list) == expected
```
natasha-aleksandrova: tests look good! i think you meant to have line 50 de-dented?

PR #1112, opened 2018-11-02
Lesson 02 Exercises

patchcarrier:
Submitting my work for all three exercises in Lessson 02.


students/patchcarrier/Lesson02/fizzbuzz_exercise.py
```diff
@@ -0,0 +1,14 @@
+for k in range(1,101):
+    
+    if not((k % 5) or (k % 3)):
+        print('FizzBuzz')
+    
+    elif not(k % 5):
+        print('Buzz')
+    
+    elif not(k % 3):
```
natasha-aleksandrova: no need for parens here, `not` is a statement not a function.

also convention is to check `k % 3 == 0`

PR #1113, opened 2018-11-02
Incorporated Charles' suggestions and made a couple of updates in mailroom2 and mailroom3. 

elnurad:
In mailroom2:
used an os module in the letter_to_all() function.

In mailroom3:
-used an os module in the letter_to_all() function.
-added an option to quit in the main part of the program so that it quits if the user raises an error thus avoiding an infinite loop.


elnurad:
Thank you, Natasha! Fixed it.

PR #1114, opened 2018-11-03
lesson 2, fizz buzz ready for grading

geekwriter2:


students/MichaelM/lesson2/Lesson2_Ex2_FizzBuzz.py
```diff
@@ -0,0 +1,30 @@
...
+# run Lesson2_Ex2_FizzBuzz.py
+# git status
+# git add Lesson2_Ex2_FizzBuzz.py
+# git commit
+# git push
+# goto https://github.com/geekwriter2/Self_Paced-Online/tree/master/students
+# click Pull request > new pull request
+# import name_main
+# run Lesson2_Ex1.py
+# from Natasha: python uses snake_case convention for var names (and functions)> horizontal_line
```
natasha-aleksandrova: nice notes 🙂

students/MichaelM/lesson2/Lesson2_Ex2_FizzBuzz.py
```diff
@@ -0,0 +1,30 @@
...
+
+def buzzles_and_fizzles(top_pot):
+        for i in range(top_pot):
```
natasha-aleksandrova: another thing to keep in mind is indentation, in python it's pretty important and convention is to use 4 spaces (not tabs).

if you use an editor like sublime, you can configure or Pycharm IDE does it out of the box.

students/MichaelM/lesson2/Lesson2_Ex2_FizzBuzz.py
```diff
@@ -0,0 +1,30 @@
...
+            if (i%3) == 0:
+                print("Fizz")
+            elif (i%5) == 0:
+                print("Buzz")
+            elif (i%5) == 0 & (i%3) == 0:
```
natasha-aleksandrova: the convention is to use `and` or `or` statements.

`&` is a bitwise operator and will always evaluate both sides, where `and` will not evaluate `i % 3` if `i % 5` is False.

students/MichaelM/lesson2/Lesson2_Ex2_FizzBuzz.py
```diff
@@ -0,0 +1,30 @@
...
+# git push
+# goto https://github.com/geekwriter2/Self_Paced-Online/tree/master/students
+# click Pull request > new pull request
+# import name_main
+# run Lesson2_Ex1.py
+# from Natasha: python uses snake_case convention for var names (and functions)> horizontal_line
+
+def buzzles_and_fizzles(top_pot):
+        for i in range(top_pot):
+            if (i%3) == 0:
```
natasha-aleksandrova: if i == 15, then this line will evaluate to True and will print "Fizz"; however the requirement is to print FizzBuzz if it divisible by both 3 and 5.

You need to move up your check on line 20 to be the first thing you check for.

PR #1115, opened 2018-11-03
Lesson 2 Assignments

Percolator787:
Submitting lesson 2 assignments.  This includes the following files (all of lesson 2):

- fizz_buzz.py
- grid_print.py
- series.py

Also included in this pull is the README.pst file.  I thought I had done the pull correctly for this before but it seems like I didn't do it correctly.  I am new to git and apparently did something wrong on my initial attempt.  The git video lesson in lesson 2 helped me understand how to do the commits and pulls.

### DIFF
--- a/students/Dennis_Coffey/lesson02/fizz_buzz.py
+++ b/students/Dennis_Coffey/lesson02/fizz_buzz.py
@@ -4,7 +4,7 @@ Created on Wed Oct 31 21:28:33 2018
 
 @author: dennis
 """
-
+"""FizzBuzz assignment for Lesson 2"""
 for i in range(100):
     if i % 5 + i % 3 == 0:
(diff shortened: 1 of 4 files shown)


students/Dennis_Coffey/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Wed Oct 31 21:28:33 2018
+
+@author: dennis
+"""
+"""FizzBuzz assignment for Lesson 2"""
+for i in range(100):
```
natasha-aleksandrova: note that `range` does not include 100, you will want to do `range(1, 101)`

PR #1125, opened 2018-11-05
Lesson03 assignments

njschafi:
Finished the following assignments for lesson03:

- slicing.py
- series.py
- strformat.py


njschafi:
Please disregard series.py assignment. I accidentally put that into the lesson03 folder. Thanks!

PR #1126, opened 2018-11-05
Homework lesson 2 assigments complete

jaredmulholland:
Still getting used to using git (that's why I'm submitting all the assignments at one.

On the Fibonacci part, I'm a little unsure what 'assert' is used for. I get that it can show you an error in your function, but it doesn't really seem that helpful, since I get an error anyway.

### DIFF
--- a/lesson_2/fizzbuzz.py
+++ b/lesson_2/fizzbuzz.py
@@ -1,0 +1,20 @@
+"""
+Prob 2: Fizz Buzz
+
+-Write a program that prints the numbers from 1 to 100 inclusive.
+-But for multiples of three print “Fizz” instead of the number.
+-For the multiples of five print “Buzz” instead of the number.
+-For numbers which are multiples of both three and five print “FizzBuzz” instead. 
+
+"""
+
+
+for x in range(1,101):
+    if x % 3 == 0 and x % 5 == 0:
+        print("FizzBuzz")
+    elif x % 3 == 0 and x % 5 != 0:
+        print("Fizz")
+    elif x % 3 != 0 and x % 5 == 0:
+        print("Buzz")
+    else:
+        print(x)
--- a/lesson_2/series.py
+++ b/lesson_2/series.py
@@ -1,0 +1,40 @@
+"""
+Problem 3: fibonacci, lucas, sum_series
+
+"""
+
+def fibonacci(n):
+    """return the nth value in a fibonacci series (starting with 0 index)"""
+    if n == 0:
+        return(0)
+    elif n == 1:
+        return(1)
+    else:
+        return fibonacci(n-1) + fibonacci(n-2)
+
+
+def lucas(n):
+    """return the nth value in a lucas series (starting with 0 index). Lucas series begins with 2,1,3,..."""
+    if n == 0:
+        return(2)
+    elif n == 1:
+        return(1)
+    else: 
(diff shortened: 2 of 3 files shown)


natasha-aleksandrova:
@jaredmulholland if you get an error from `assert` statement then the code should be fixed to pass the assert 😉

jaredmulholland:
That part I get. I guess I don’t get what it gives you in addition to a
normal error statement.

natasha-aleksandrova:
`assert` is used for logic error checking not syntax or compiled errors.

for example,

```
def add(x, y):
    return x + y


assert add(2, 2) == 5
```
Above is a valid function that will not return an error when executed; our assert statement ensures that function returns what we expect. 
Obviously it should return 4 so the assert statement will fail, `assert` statement is often used with `pytest` unit test library which we'll get into in lesson 6.

jaredmulholland:
Got it. Thanks! 😀

lesson_2/fizzbuzz.py
```diff
@@ -0,0 +1,20 @@
...
+
+for x in range(1,101):
+    if x % 3 == 0 and x % 5 == 0:
+        print("FizzBuzz")
+    elif x % 3 == 0 and x % 5 != 0:
+        print("Fizz")
+    elif x % 3 != 0 and x % 5 == 0:
+        print("Buzz")
+    else:
+        print(x)
```
natasha-aleksandrova: your `if` statement at line 13 already checks for being divisible by both 3 and 5 so `elif` checks on lines 15 and 17 don't need to check both.

meaning:
```
if x % 3 == 0 and x % 5 == 0
elif x % 3 == 0
elif x % 5 == 0
```
are sufficient

lesson_2/series.py
```diff
@@ -0,0 +1,40 @@
...
+        return(x)
+    elif n == 1:
+        return(y)
+    else:
+        return (sum_series((n-1), x, y) + sum_series((n-2), x, y))
+
+
+assert fibonacci(5) == 5
+assert lucas(5) == 5
+assert sum_series(5, x = 0, y = 1) == 5
```
natasha-aleksandrova: looks good!

lesson_2/series.py
```diff
@@ -0,0 +1,40 @@
...
+    if n == 0:
+        return(2)
+    elif n == 1:
+        return(1)
+    else: 
+        return lucas(n-1) + lucas(n-2)
+
+def sum_series(n, x = 0, y = 1):
+    """return sum of n-1 and n-2 numbers in a sequence. This function allows the user to get a result similar to a lucas 
+    or fibonacci sequence but at any starting point"""
```
natasha-aleksandrova: nice use of docstrings

lesson_2/series.py
```diff
@@ -0,0 +1,40 @@
...
+def lucas(n):
+    """return the nth value in a lucas series (starting with 0 index). Lucas series begins with 2,1,3,..."""
+    if n == 0:
+        return(2)
+    elif n == 1:
+        return(1)
+    else: 
+        return lucas(n-1) + lucas(n-2)
+
+def sum_series(n, x = 0, y = 1):
```
natasha-aleksandrova: minor pep8 style violation: no need for spaces around `=` sign -> `(n, x=0, y=1)`

PR #1132, opened 2018-11-06
progress request, two commits 

geekwriter2:


students/MichaelM/lesson2/series.py
```diff
@@ -0,0 +1,58 @@
...
+    # 0, 1 Fib
+    # 2, 1 Luc
+    # 1, 1 undefined
+    term_values = [n_minus2, n_minus1]
+    for i in range(1, n - 1):
+        new_value = term_values[loop_counter] + term_values[loop_counter - 1]
+        term_values.append(new_value)
+        loop_counter += 1
+    result = max(term_values)
+    return result
```
natasha-aleksandrova: note that `sum_series(9)` should return 34

PR #1134, opened 2018-11-07
Adds grid printer solution to lesson02

scottebetts:
I hope I'm doing this right.


students/scottebetts/lesson02/Fizz-Buzz.py
```diff
@@ -0,0 +1,13 @@
...
+        print('Fizzbuzz')
+    elif n % 3 == 0:
+        print('Fizz')
+    elif n % 5 == 0:
+        print('Buzz')
+    else:
+        print(n)
+
+for i in range(1, 100):
+    fizzbuzz(i)
```
charlesrobison: You'd actually want to put this for loop inside your fizzbuzz function.  Something like:
`def fizzbuzz(n):
    for i i range (1,n):
        if i % 3 == 0`
And so on.

students/scottebetts/lesson02/Grid-Printer.py
```diff
@@ -0,0 +1,34 @@
...
+    wall = '|'
+    space = ' '
+    cell = (wall + ((space * cell_hw) * 2 + space + wall) * table_hw)
+    vertice = ((plus + space + ((minus + space) * cell_hw) ) * table_hw + plus)
+
+    for t in range(table_hw):
+        print(vertice)
+        for i in range(cell_hw):
+            print(cell)
+    print(vertice)
```
charlesrobison: Works well!

PR #1145, opened 2018-11-11
added lessons8 and 9

chelseamsmith:


students/ChelseaSmith/lesson8/circle.py
```diff
@@ -0,0 +1,88 @@
...
+        return 2*self.radius
+    @diameter.setter
+    def diameter(self, val):
+        self.radius = val / 2
+
+    @property
+    def area(self):
+        return math.pi * self.radius ** 2
+
+
```
natasha-aleksandrova: your class methods should have one line space between them (where module functions have 2 lines).

students/ChelseaSmith/lesson9/mailroom.py
```diff
@@ -0,0 +1,113 @@
...
+            filename = item.name + ".txt"
+            f = open(filename, "w")
+            f.write(item.write_letter())
+            f.close()
+
+    def save_data(self):
+        f = open(database, "w")
+        for item in record:
+            f.write(f"{item.name}, {', '.join(item.donations)}\n")
+        f.close()
```
natasha-aleksandrova: use `with` statement with files.

students/ChelseaSmith/lesson9/mailroom.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def add_donation(self, name, amount):
+        new = self.check_name(name)
+        if new == True:
```
natasha-aleksandrova: use `is True` when comparing objects.

students/ChelseaSmith/lesson9/mailroom.py
```diff
@@ -0,0 +1,113 @@
...
+if __name__ == "__main__":
+    choice = 5
+    while choice != 0:
+        choice = int(input("Please enter 1 to enter a new donation, 2 to generate a report, 3 to generate letters for all donors, or 0 to quit"))
+        if choice == 1:
+            new_donation()
+        elif choice == 2:
+            donors.report()
+        elif choice == 3:
+            donors.letters_all()
```
natasha-aleksandrova: I get an error running your code:
```
FileNotFoundError: [Errno 2] No such file or directory: 'donors.txt'
```

natasha-aleksandrova: switch dict is still a good pattern here.

students/ChelseaSmith/lesson9/mailroom.py
```diff
@@ -0,0 +1,113 @@
...
+        number = len(self._donations)
+        return number
+
+    def average_gift(self):
+        ave = sum(self._donations)/len(self._donations)
+        return ave
+
+    def write_letter(self):
+        letter = "Dear {},\nThank you for your generous donation of ${:.2f}.\nSincerely,\n Generic Charities".format(self._name, self._donations[-1])
+        return letter
```
natasha-aleksandrova: `Donor` class looks good

students/ChelseaSmith/lesson9/mailroom.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class Manager:
+
+    def __init__(self, database):
+        self.database = database
+        self.record = list()
+        for line in open(database):
+            self.record.append(Donor(line))
```
natasha-aleksandrova: Couldn't get the code to run but curious how this works? The code appears to save a line with name and donations, but this appears to only pass in one arg which would map to the name param?

students/ChelseaSmith/lesson9/mailroom.py
```diff
@@ -0,0 +1,113 @@
...
+        print(f"Donor name" + " " * 7 + " | Total Given | Num Gifts | Average Gift\n")
+        for item in self.record:
+            print(f"{item.name:<18}${item.total_given():12.2f}{item.number_of_gifts():10}   ${item.average_gift():13.2f}\n")
+
+    def check_name(self, name):
+        new = False
+        for item in self.record:
+            if item.name == name:
+                new = True
+        return new
```
natasha-aleksandrova: return as soon as you hit `True` condition:
```
for item in self.record:
    if item.name == name:
        return True
    return False
```

PR #1146, opened 2018-11-11
lesson07

elnurad:
Committing the lesson 07 exercise files. I had a merge conflict and after resolving it I ended with two folders here (lesson07&lesson7) with the same files. The ones in lesson07 folder are more complete since they have all the html output pages so you only need to merge that one. Also, my final html output looks correct and so does the code, but I get "Assert False" errors after I run the pytest and I can't seem to figure out why.


natasha-aleksandrova:
@elnurad the tests are failing because of extraneous new lines.

For example when I looked at `test_ul` test

and added:
```
    print("start")
    print(file_contents)
    print('end')
```
this is what i get:
```
start
<ul>
  some text
</ul>

end
```

if I change assert to `assert file_contents.strip().endswith('</ul>')` then it passes because `strip` will remove the newlines.

I would go over your code to see where that happens.

elnurad:
Thank you, Natasha! Much appreciated!

PR #1149, opened 2018-11-13
MailRoom Part 4 - Unit Testing

ldenney89:
Pull Request for submitting Mailroom Part 4 Assignment. 

I had a little bit of difficulty with this assignment, as most of my functions were not returning anything and so assert statements didn't work. I modified the functions to remedy this, making it more testable. 

Still not 100% confident on the testing process though in terms of real-life examples.

### DIFF
--- a/students/laura_denney/lesson06/test_mailroom_part4.py
+++ b/students/laura_denney/lesson06/test_mailroom_part4.py
@@ -1,0 +1,16 @@
+#-------------------------------------------------#
+# Title: Mail Room Part 4 Unit Testing
+# Dev:   LDenney
+# Date:  November 7, 2018
+# ChangeLog: (Who, When, What)
+#   Laura Denney, 11/7/18, Started MailRoom Part 4
+#-------------------------------------------------#
+
+from mailroom_part4 import *
+
+def test_1():
+    assert len(sort_by_amount()) >= 5
+    assert type(sort_by_amount()) is list
+
+def test_2():
+    assert quit() == 'quit'
--- a/students/laura_denney/lesson06/test_mailroom_part4.py
+++ b/students/laura_denney/lesson06/test_mailroom_part4.py
@@ -4,13 +4,52 @@
 # Date:  November 7, 2018
 # ChangeLog: (Who, When, What)
 #   Laura Denney, 11/7/18, Started MailRoom Part 4
+#   Laura Denney, 11/12/18, Completed work on MailRoom Part 4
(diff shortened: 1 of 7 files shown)


students/laura_denney/lesson06/test_mailroom_part4.py
```diff
@@ -0,0 +1,55 @@
+#-------------------------------------------------#
+# Title: Mail Room Part 4 Unit Testing
+# Dev:   LDenney
+# Date:  November 7, 2018
+# ChangeLog: (Who, When, What)
+#   Laura Denney, 11/7/18, Started MailRoom Part 4
+#   Laura Denney, 11/12/18, Completed work on MailRoom Part 4
+#-------------------------------------------------#
+
+from mailroom_part4 import *
```
natasha-aleksandrova: avoid using start imports (they are evil 🙂 )

students/laura_denney/lesson06/test_mailroom_part4.py
```diff
@@ -0,0 +1,55 @@
...
+
+#tests sort_by_amount
+def test_1():
```
natasha-aleksandrova: recommend naming your tests with more friendly names like `test_sort_by_amount`

students/laura_denney/lesson06/test_mailroom_part4.py
```diff
@@ -0,0 +1,55 @@
...
+    assert len(sort_by_amount()) >= 5
+    assert type(sort_by_amount()) is list
```
natasha-aleksandrova: you should assert what `sort_by_amount` returns, for example:
```
actual = sort_by_amount()
expected = [1, 2, 3]  # this is fake
assert actual == expected
```

students/laura_denney/lesson06/test_mailroom_part4.py
```diff
@@ -0,0 +1,55 @@
...
+def test_4_5():
+    assert validate_donation('four hundred') == 0.0
+
+#tests the function to create a string for report
+def test_7():
+    assert create_string_report() == '\nDonor Name                | Total Given | Num Gifts | Average Gift\n---------------------------------------------------\
+---------------'
+
+def test_7_5():
+    assert create_string_report(sort_by_amount()) > create_string_report()
```
natasha-aleksandrova: this seems like a weird comparison for the test.

I would assert same way you assert in `test_7` but against the new string.

PR #1156, opened 2018-11-14
Lesson 10 functional mailroom

rbillupsii:


students/RoyC/Lesson10/mail_menu.py
```diff
@@ -0,0 +1,129 @@
...
+        donation_records.clear_donations()
+        print("\nDonor records have been cleared\n")
+        
+def calculate_challenge_projection():
+    """
+    Request challenge values and displays the result
+    """
+    print("\nProjecting challenge donations:")
+    min = input_num("Enter minimum donation amount or 'none': ", skip="none")
+    max = input_num("Enter maximum donation amount or 'none': ", skip="none")
```
natasha-aleksandrova: be really careful here.. `min` and `max` are built-in function - and here you over wrote them.

PR #1157, opened 2018-11-14
Lesson 5 work uploaded

JonCracolici:
I will be making the variables all snake case in lesson 6. I'm not sure how I got it in my head that functions are snake case and variables just lower. I will also be refactoring the user input checking to make it more testable. Thanks!

### DIFF
--- a/students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
+++ b/students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
@@ -1,0 +1,180 @@
+#Jon Cracolici
+#Lesson 05 Mailroom 3
+#UW Python Cert
+#import statements
+from collections import OrderedDict
+#database set up
+def initialize_database():
+    """This function intializes the database when called."""
+    william_gates_iii = {'fname': 'William_Gates_III', 'donor': 'William Gates, III', 'totaldonated': 653784.49, 'numdonations': 2, 'avgdonated': 326892.24, 'lastdonation': 30000.00}
+    mark_zuckerberg = {'fname': 'Mark_Zuckerberg', 'donor': 'Mark Zuckerberg', 'totaldonated': 16396.10 , 'numdonations': 3, 'avgdonated': 5465.37, 'lastdonation': 5000.25}
+    jeff_bezos = {'fname': 'Jeff_Bezos', 'donor': 'Jeff Bezos', 'totaldonated': 877.33, 'numdonations': 1, 'avgdonated': 877.33, 'lastdonation': 877.33}
+    paul_allen = {'fname': 'Paul_Allen', 'donor': 'Paul Allen', 'totaldonated': 708.42, 'numdonations': 3, 'avgdonated': 236.14, 'lastdonation': 100.95}
+    
+    dB = {'william_gates_iii': william_gates_iii,'mark_zuckerberg': mark_zuckerberg,'jeff_bezos': jeff_bezos, 'paul_allen': paul_allen}
+    return dB
+#Functions relating to the start menu
+def start_control(dB):
+    """Creates the control flow for the start menu."""
+    start_menu_display()
+    response = start_user_input()
+    try:
+        return startmenulogic[response](dB)
+    except KeyError:
+        print("You have input an invalid instruction. Please enter 1, 2, 3, or 4")
+        return
+def saty_control(dB):
+    """Creates the control flow for the send a thank you menu."""
+    saty_menu_display()
+    response = saty_user_input()
+    try:
+        return satymenulogic[response](dB)
+    except KeyError:
+        print("You have input an invalid instruction. Please enter 1, 2, 3, or 4")
+        return        
+def start_menu_display():
+    """Prints the start menu to the screen."""
+    print('Welcome to our mailroom app!')
+    print('Please select the number of your choice from the following options')
+    print('1) Send a Thank You')
+    print('2) Create a Report')
+    print('3) Send Letters to Everyone')
+    print('4) Quit')
+def saty_menu_display():
+    """Prints the send a thank you menu to the screen."""
+    print('You have chosen to send a thank you note to a donor')
+    print('Please select the number of you choice from the following options')
+    print('1) See a list of current donors')
+    print('2) Send a thank you to a current donor for a new gift')
+    print('3) Send a thank you to a first time donor')
+    print('4) Return to the start menu')
+def start_user_input():
+    """Collects user input. I separate this for easier error handling in future."""
+    task = input('Please type the number of your selection.')
+    return task
+def saty_user_input():
+    """Collects user input. I separate this for easier error handling in future."""
+    task = input('Please type the number of your selection.')
+    return task
+def new_donor(newdonordict, dB):
+    """Creates a new donor in the database and requests donation ammount."""
+    updict = {newdonordict['donornamekey']:{'fname':newdonordict['donornamekey'].title(), 'donor': newdonordict['donor'], 'totaldonated': newdonordict['donation'], 'numdonations': 1, 'avgdonated': newdonordict['donation'], 'lastdonation': newdonordict['donation']}}
+    dB.update(updict)
+    return dB
+def update_donor(newdondict, dB):
+    """Updates an existing donor in the database with a new donation."""
+    dB[newdondict['donornamekey']]['totaldonated'] += newdondict['donation']
+    dB[newdondict['donornamekey']]['numdonations']+=1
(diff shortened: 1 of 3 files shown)


students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
```diff
@@ -0,0 +1,180 @@
...
+def quit_saty(db):
+    print("Bye bye!")
+    return False
+def main():
+    db = initialize_database()
+    while True:
+        x = start_control(db)
+        if x == False:
+            break
+def main_2(db):
```
natasha-aleksandrova: per pep8 make sure to have 2 lines space between module functions; code is a bit difficult to read with no spaces.

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
```diff
@@ -0,0 +1,180 @@
...
+    while True:
+        x = saty_control(db)
+        if x == False:
+            break
+#### This is the beginning of the executed program.
+if __name__=="__main__":
+
+    startmenulogic = {'1': main_2, '2': create_report, '3': mass_mail, '4': quit_program}
+    satymenulogic = {'1': show_current_names, '2': ud_control, '3': nd_control, '4': quit_saty}
+    main()
```
natasha-aleksandrova: recommend simplifying your main entry point, right now there are a lot of layers when it can be more straightforward. Example:
```
def quit_program():
    print("Exiting")
    sys.exit()


def main():
    while True:
        response = input("")
        try:
            switch_dict[response]()
        except KeyError:
            print("Invalid option")

....
switch_dict = {'1': send_thank_you, '2': create_report, '3': mass_mail, '4': quit_program}
main()
```

this is all you really need to do:
* simple while True loop that should only break on "quit" option
* quit is  mapped to sys.exit which simply exits your script
* no special return statement in feature functions to control main flow when it shouldn't really

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
```diff
@@ -0,0 +1,180 @@
...
+            donation = float(newdonation)
+        except ValueError:
+            print("Use numbers for the donation amount. NUMBERS!")
+            return False
+        newdondict = {'donornamekey': donornamekey, 'donor': donor, 'donation': donation}
+        return newdondict
+def print_letters(donordict):
+    """This function takes a string and makes a document with it."""
+    lettername = donordict['fname']+'.txt'
+    outfile = open(lettername, 'w')
```
natasha-aleksandrova: prefer to use `with` statement when working with files

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
```diff
@@ -0,0 +1,180 @@
...
+    outfile.write(note_gen(donordict, dest = 'f'))
+    outfile.close()
+def note_gen(persondict, dest = 's'):
+    """Creates a thank you note. Includes a required positional argument of the intended recipient's info,
+    and a kwarg "dest" that defaults to printing to screen but may be set to generate a file."""
+    message = "Dear {donor}, \nThank you for your generous donation of ${lastdonation:.02f}. Please rest assured that we will use at least \n95% of your contribution to feed the homeless to wolves. We could not do this work without you. \nSincerely, \nThe Billionaires' Club".format(**persondict)
+    if dest == 's':
+        print(message)
+    elif dest == 'f':
+        return message
```
natasha-aleksandrova: a good rule of thumb is for a function to do one thing, in this case  it can be `def get_letter_text` and just return the letter text, the functions that use it can decide what to do with this text, either write text to file or print to console.


This will become even more important in Lesson 9 where separation of business logic (like do I write to file or screen?) vs data model logic (like generate letter text).

Think about separating "what" from "how".

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
```diff
@@ -0,0 +1,180 @@
...
+        print("It appears that you are attempting t update a donor not already in our system.\n Perhaps you would like to select the 'New Donor' option instead?")
+        return False
+    #will add try except to check for donor actually being new in L5
+    update_donor(updonordict, dB)
+    note_gen(dB[updonordict['donornamekey']])
+    return dB#, start_control(dB)
+def donor_input(dB):
+    """This function collects user input about the donation."""
+    #create a set of quit criteria
+    quitset = set(['quit', 'q'])
```
natasha-aleksandrova: no need for `set` here; a tuple is more suitable because it's immutable and good for constant type definitions.

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
```diff
@@ -0,0 +1,180 @@
...
+    #return display #, saty_control(dB)
+
+def create_report(dB):
+    """Creates a summary report for all current donors."""
+    dictlist = OrderedDict(sorted(dB.items(), key=lambda x: x[1]['totaldonated'], reverse=True))
+    print('Donor Name                | Total Given | Num Gifts | Average Gift')
+    print('------------------------------------------------------------------')
+    for i in dictlist:
+        print('{donor:26}${totaldonated:>13.2f}{numdonations:>12} ${avgdonated:>12.2f}'.format(**dB[i]))
+    return
```
natasha-aleksandrova: no need for `return` statement

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
JonCracolici: ok will do

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
JonCracolici: I have heard that it is preferred because it automatically closes the file. but don't people just close the files regardless? Or is there some other reason?

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
JonCracolici: ok. Just tried it and works fine. What is set good for then? being able to keep checking as things get added?

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
JonCracolici: ok

JonCracolici:
Thanks for the tip, I will look into Pycharm. I have just been using Notepad++.

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
natasha-aleksandrova: Exactly, it ensures the file is closed in case of an exception, and the code is cleaner.

Not using with is considered an anti-pattern:
https://docs.quantifiedcode.com/python-anti-patterns/maintainability/not_using_with_to_open_files.html

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
natasha-aleksandrova: Great question; there are many uses for sets but I will give you an example.

Let's say in another use case we are pulling donations from the database and databases records are returned as lists of lists like:
```
donors = [["Bob", 300], ["Mary", 200], ["Bob", 400]]
```
and you need to get a list of donors (unique)

you could:
```
names = []
for donor in donors:
    if name not in names:
        names.append(donor[0])
```
or
```
names = set()
for donor in donors:
    names.add(donor[0])
```

set code is much simpler and you're utilizing a data structure built for this

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
JonCracolici: that website is pretty cool. I will switch to "with"

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
JonCracolici: so sets can be really useful when you are checking things that might have multiple identical entries. Like in that fruit lab exercise.

natasha-aleksandrova: Exactly

students/Cracolici_Jon/lesson05/JonCracolici_Mailroom3.py
natasha-aleksandrova: yes it is! recommend reading more of their posts.

PR #1164, opened 2018-11-17
Lesson 07 Josh Hoff

josh1209:
Finished Lesson 7; verified the result online at validator.w3.org.


students/Josh_HOff/lesson07/html_render.py
```diff
@@ -0,0 +1,197 @@
...
+class Title(OneLineTag):
+
+    tag = 'title'
+    
+
+class SelfClosingTag(Element):
+    def __init__(self, content=None, **kwargs):
+        if self.tag == 'meta charset="UTF-8"':
+            kwargs = {}
+        self.test_dict = copy.deepcopy(kwargs)
```
natasha-aleksandrova: I don't think you this since you're passing in `**kwargs`

PR #1166, opened 2018-11-17
Lesson3_Ex2_Lists.py ready for grading

geekwriter2:


students/MichaelM/lesson3/Lesson3_Ex2_Lists.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def my_fruit_remove_selection(my_fruit, fruit_cart):
+    """
+        removes a specific term from a list
+
+        {Extended description}
+
+        Parameters:
+        my_fruit (string): term to be deleted from the list
```
natasha-aleksandrova: should be named `fruit_to_delete` for clarity

PR #1168, opened 2018-11-19
Lesson04 Assignment Submission

patchcarrier:
Submitting the files mailroom2.py and trigram.py for grading


students/patchcarrier/Lesson04/trigram.py
```diff
@@ -0,0 +1,217 @@
...
+        # output what the last line would have been before the search
+        if line_i[-1][-1] != '.':
+            outfile.write(' '.join(last_line))      
+        # the search was successful, append the additional lines that end with
+        # a period
+        else:
+            
+            for line in lines_after_wordlim:
+                outfile.write(' '.join(line))
+            return
```
natasha-aleksandrova: for readability, above while loop is quite long with a lot of logic inside - recommend breaking code like that into some logical functions that can improve the readability.

PR #1171, opened 2018-11-21
Added lesson 10 program and  text file that goes with lesson 9 code

chelseamsmith:


students/ChelseaSmith/lesson10/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+        print(f"Donor name" + " " * 7 + " | Total Given | Num Gifts | Average Gift\n")
+        for item in self.record:
+            print(f"{item.name:<18}${item.total_given():12.2f}{item.number_of_gifts():10}   ${item.average_gift():13.2f}\n")
+
+    def check_name(self, name):
+        new = False
+        for item in self.record:
+            if item.name == name:
+                new = True
+        return new
```
natasha-aleksandrova: note that you don't need the temp var `new`, you can control this with a return statement to simplify:
```
for item in self.record:
    if item.name == name:
        return True
else:
    return False

PR #1172, opened 2018-11-21
lesson08 submission - Mark Luckeroth

mluckeroth:


students/mark_luckeroth/lesson08/test_circles.py
```diff
@@ -0,0 +1,83 @@
+"""
+test code for circles.py
+
+"""
+
+import pytest
+import math
+from random import shuffle
+
+from circles import *
```
charlesrobison: Simply use import circles - using star not advised.

PR #1176, opened 2018-11-21
Lesson 8 Assignment Josh Hoff

josh1209:
Finished Lesson 8 Assignment with test document to supplement.


students/Josh_HOff/lesson08/circle_class.py
```diff
@@ -0,0 +1,134 @@
...
+import pathlib
+from functools import total_ordering
+
+circles = []
+@total_ordering
+class Circle(object):
+
+    def __init__(self, value):
+        global circles
+        global templist
```
natasha-aleksandrova: what are these globals? should not be here

students/Josh_HOff/lesson08/circle_class.py
```diff
@@ -0,0 +1,134 @@
...
+        return Circle(value/2)
+        
+    @property
+    def radius(self):
+        return self._radius
+    @radius.setter
+    def radius(self, value):
+        self._radius = value
+        self._diameter = value * 2
+        self._area = (math.pi) * (value ** 2)
```
natasha-aleksandrova: since you have setter/getters for diameter and area then only set them there; right now you have them duplicated in 3 places

students/Josh_HOff/lesson08/circle_class.py
```diff
@@ -0,0 +1,134 @@
...
+    
+    @staticmethod
+    def from_diameter(value):
+        return Circle(value/2)
+        
+    @property
+    def radius(self):
+        return self._radius
+    @radius.setter
```
natasha-aleksandrova: make sure to have 1 line space between class methods

students/Josh_HOff/lesson08/circle_class.py
```diff
@@ -0,0 +1,134 @@
...
+    def diameter(self, value):
+        self._diameter = value
+        self._radius = value / 2
+        self._area = (.25 * math.pi) * (value ** 2)
+    @property
+    def area(self):
+        return self._area
+    @area.setter
+    def area(self, value):
+        raise AttributeError
```
natasha-aleksandrova: if you don't define a `setter` then it will error anyway

students/Josh_HOff/lesson08/circle_class.py
```diff
@@ -0,0 +1,134 @@
...
+    def __init__(self, value):
+        global circles
+        global templist
+        self._radius = value
+        self._diameter = value * 2
+        self._area = (math.pi) * (value ** 2)
+    
+    @staticmethod
+    def from_diameter(value):
+        return Circle(value/2)
```
natasha-aleksandrova: this should be a classmethod, not staticmethod with input arg being `cls`

natasha-aleksandrova:
@josh1209 there are conflicts with another student files so I cannot merge it in as is.

Can you please try merging from upstream?
https://help.github.com/articles/syncing-a-fork/

josh1209:
> @josh1209 there are conflicts with another student files so I cannot merge it in as is.
> 
> Can you please try merging from upstream?
> https://help.github.com/articles/syncing-a-fork/

I've merged from upstream; let me know if I did it wrong. I see your comments and I'll adjust my code accordingly tomorrow. Thanks Natasha!

natasha-aleksandrova:
@josh1209 thank you! no more conflicts 🙂

students/Josh_HOff/lesson08/circle_class.py
josh1209: My mistake; I removed them.

students/Josh_HOff/lesson08/circle_class.py
josh1209: @natasha-aleksandrova So I could just delete lines 41 through 43 and it will work essentially the same?

students/Josh_HOff/lesson08/circle_class.py
josh1209: Okay - changed this to class method

students/Josh_HOff/lesson08/circle_class.py
josh1209: @natasha-aleksandrova Could you please explain this in more detail to me? Should I be removing variables or should I be changing my setter methods?

natasha-aleksandrova: so to simplify this is what you could have:
```
def __init__(self, radius):
    self.radius = radius

@property
def diameter(self):
    return self.radius * 2.0

@diameter.setter
def diameter(self, value):
    self.radius = value / 2.0
```
(the example is only for diameter but applies to area as well).

as I mentioned before, right now you have diameter being written in 3 places - in the snippet above the value of diameter is isolated to a single block of code, in the property setter.

students/Josh_HOff/lesson08/circle_class.py
natasha-aleksandrova: correct; if attribute doesn't exist it will throw the same error:
```
>>> class Foo: pass
...
>>> Foo().foo
  File "<stdin>", line 1, in <module>
AttributeError: 'Foo' object has no attribute 'foo'
>>>
```

PR #1177, opened 2018-11-22
completed lesson 2 exercises fizzbuzz and fibonacci series

cocokaku:
added files fizzbuzz.py and series.py to lesson2 folder

### DIFF
--- a/students/cocokaku/lesson2/fizzbuzz.py
+++ b/students/cocokaku/lesson2/fizzbuzz.py
@@ -1,0 +1,5 @@
+for i in range(1, 101):
+    if i%15==0: print("FizzBuzz")
+    elif i%3==0: print("Fizz")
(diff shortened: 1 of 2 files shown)


students/cocokaku/lesson2/fizzbuzz.py
```diff
@@ -0,0 +1,5 @@
+for i in range(1, 101):
+    if i%15==0: print("FizzBuzz")
+    elif i%3==0: print("Fizz")
+    elif i%5==0: print("Buzz")
+    else: print(i)
```
natasha-aleksandrova: although it's a valid syntax, this is against python style guidelines; make sure to use separate lines for if blocks:
```
if i % 15 == 0:
    print("FizzBuzz")
```

PR #1194, opened 2018-11-28
mailroom_part2

stefanlund:
Works and looks better, hopefully good enough for now.

This is the first time I have used the sys.exit(). Can you tell me a little bit more about this method? When to use it …

switch_dict, I had envisioned displaying the menu, asking for user choice input and returning the call to functions as a generic   options - choice - function caller   for any menu provided as a class. The other side is then to have the functions and the corresponding switch_dict. Together it's a mailroom or a bank etc. One sentence answer? Good, bad idea or we'll get to something like that but later ...


natasha-aleksandrova:
https://docs.python.org/2/library/sys.html#sys.exit docs explain the function, basically whenever you want to exit your script (similar to a `return` in a function).

for your switch_dict comment.. there is a YAGNI (https://deviq.com/yagni/) programming principle that you should check out. Basically, keep it simple until you need it. Often times, if I do end up needing a "bank feature" then I will look at what could be the common code between mailroom and the bank program and refactor those common bits into common library. At this point you would be guessing and over-complicating code for possible features that might never happen.

stefanlund:
Thank you for taking the time sifting through my meandering code. The code was adjusted on the fly, more than once. As a result, the logic - intent thread is at points broken. Your code suggestion synthesizes  the current code very well. Going forward I will be mindful of keeping it simpler.

I have one question still lingering and it's regarding exiting the script. Referring to the code submitted before making the final changes you suggested. I have not had the issue of crashing out until now. All code is executed but something happens upon exiting. Somewhere a function is still waiting for a return value but it doesn't make sense. The sys.exit() took care of the issue but I'm still curious as to why the script is not terminating in a normal way. I feel I created a potential bug or at least a weakness in the code that I wasn't able to resolve, without sys.exit() and because I don't understand it I could create this situation again. With more coding experience maybe I'll just grow out of it and look back and think this is not how this kind of problems are solved so this is just not something that is going to happen again.

Looking for a simple answer like, a certain code structure to be avoided and this is what it looks like or use sys.exit() - not a problem?

natasha-aleksandrova:
@stefanlund I'd be happy to take a look at as to why it is crashing. I'm assuming this only happens with a previous version of the code? If yes, can you please provide that code so that I can run it?

stefanlund:
code, the intent was to loop only in the main_menu() and terminate after the quit() function was called. 
https://gist.github.com/stefanlund/0aabe46da22098fd6b7ea833ab94ccee

tried to make the data dictionary available as well, don't think it worked.
https://gist.github.com/stefanlund/2ba69b8404c3f1ef25ee8b96fc64b02b

natasha-aleksandrova:
@stefanlund the code won't run without the pickle file, which should be initialized for the user if it doesn't exist.

you can email me the pickle file or send via MS Teams chat.

stefanlund:
- dictionary as an attachment. Works?

- Having no experience, was sending the script using a gist the way you
expected/wanted/normal way?

natasha-aleksandrova:
1. I think I would need the .pickle file
2. Yes sharing code via gist is the way to go

stefanlund:
.pickle file:

sending it as an attachment to this email.

It's part of what makes the code run. Not sure if you just didn't see the
attachment in the previous email today or if it doesn't work any better
sending it as an attachment than attempting a gist.

Copy and paste.
� ��


Looks similar to the gist.  How should I proceed?

natasha-aleksandrova:
@stefanlund you need email the file to the course account, as a attachment. You're emailing as a response to git comment which will not attach files.

stefanlund:
Got it. Done.

stefanlund:
Just wondering if I finally managed to get you the .pickle file?
Sent to python1@uw.edu.

natasha-aleksandrova:
@stefanlund ok I got the answer for you - let me just start by saying that it took me a *long* time to figure out why it crashes.. And this was the reason why I was recommending code changes: complicated code makes complicated bugs 🙂 

The repro options: a, b, a, d

at second "a" ask_options evals main_menu which does not return your expected `return finished, option`, it returns `None`.. so this line crashes: `finished, option = ask_questions(sub_menu)` because main_menu returns nothing.

stefanlund:
All previous pointers have been very helpful and inspiring. Code looks much
better.

Thank you so very much. Really didn't mean to ask for *that* much time
spent on this. I understand some of it, main_menu() doesn't return anything
so it returns None. Makes sense.

What is not clear to me is why this sequence is taking place at all since
it goes on after the *break* command. It's game over and script has
finished so where is this "extra" frame eval main_menu(), that is not done
through input by the user, come from, and why? Where is my logic wrong?
Even with simpler code I could be tempted to use a similar structure and
keep crashing.

The sys.exit() stops this. Means I just don't know about the issue but I
suspect it's never good to have this situation and could possibly introduce
unwanted results. Don't like not to understand and even more so when I'm
witnessing a crash. A crash that takes place after I think I'm done that is.

natasha-aleksandrova:
Your `break` statement does what it's supposed to do, `break` only stops the loop it is in. You have multiple loops going on here.. something like this:

main_menu() [has while loop, this is where you see your `break` statement work] -> ask_questions() -> send_thank_you() -> ask_questions() -> main_menu() [new while true loop, because we selected option "a"]

natasha-aleksandrova:
In order to avoid weird crashes like that, your code should be clear about it's path and what is happening.. basically if you can follow the execution then you should be able to follow the bugs.

PR #1201, opened 2018-11-30
Mail room assignment

yilmazkennan:


natasha-aleksandrova:
@yilmazkennan please add the missing feature, create a new PR and reference this one in the comment so we have context. Thank you!

PR #1202, opened 2018-11-30
lesson09 submission - Mark Luckeroth

mluckeroth:


students/mark_luckeroth/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,96 @@
...
+        return self.total == other.total
+
+    def __ge__(self, other):
+        return self.total >= other.total
+
+    def __gt__(self, other):
+        return self.total > other.total
+
+    def __ne__(self, other):
+        return self.total != other.total
```
natasha-aleksandrova: nice! you could have also added here the donor thank you text template

students/mark_luckeroth/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,96 @@
...
+            for line in f:
+                load = line.split(';')
+                amount = [float(_) for _ in load[2].split(',')]
+                self.add_donor(load[0],load[1],amount)
+
+    def save_donors(self, filename='donor_list.txt'):
+        with open(filename, 'w') as f:
+            for donor in self.donorlist:
+                f.write("{};{};{}\n".format(donor.first_name, donor.last_name,
+                                            ','.join(str(a) for a in donor.donations)))
```
natasha-aleksandrova: this is a bit outside of this course.. (and something that should be covered in the next course).

Couple good choice here for file format would be:
* write data in pickle format: https://pythonprogramming.net/python-pickle-module-save-objects-serialization/
* store as json string: https://stackabuse.com/reading-and-writing-json-to-a-file-in-python/
* store as CSV: https://realpython.com/python-csv/

If you have time check them out!

students/mark_luckeroth/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @property
+    def names(self):
+        return [(_.last_name, _.first_name) for _ in self.donorlist]
```
natasha-aleksandrova: `_` is used for variables that are not used; since you are using it then you should use proper var name

students/mark_luckeroth/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def find(self, first, last):
+        if (last, first) in self.names:
+            return self.donorlist[self.names.index((last, first))]
+        else:
+            return False
```
natasha-aleksandrova: although certainly allowed in python, it is best to stick to the same type for your function return.

right now it returns a bool or a value - either stick to all bool or return value/None

students/mark_luckeroth/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def update(self, first, last, amount):
+        if (last, first) in self.names:
+            self.find(first, last).add_donation(amount)
```
natasha-aleksandrova: consider re-thinking this block; right now your `find` function and this function both concern with checking if donor exist.

Couple options:
```
def get_donor():
    pass


def update():
    existing_donor = self.get_donor(first, last)
    if existing_donor:
        # update
    else:
        # add new
```

PR #1216, opened 2018-12-04
Lesson9

lauraannf:


natasha-aleksandrova:
Merging this PR - however would like to see objects being actually utilized

lauraannf:
I use Donor on line 50 of mailroom_oo, so then each donor in the donor_list
is a Donor object.
Is this the application you're looking for?

natasha-aleksandrova:
That's exactly it - thank you for pointing that out, not sure how I missed that line.

lauraannf:
Oh good, I was worried that I was doing everything totally wrong!

PR #1217, opened 2018-12-04
Lesson02 completed for John Rogers (sum_series.py)

johnroge:
Had to have some of the recursion explained to me step by step on paper but now I've got it. I thought it was adding at each step, not just the eventual base cases. 

I understand the very last part about default values about 99%; can you recommend any exercises where I would need to do something like the following to enforce optional parameters? Appreciate any help/ advice. 

return sum_series(n -1, n0, n1) + sum_series(n -2, n0, n1)

### DIFF
--- a/students/Net_Michael/lesson3/list_lab.py
+++ b/students/Net_Michael/lesson3/list_lab.py
@@ -1,0 +1,215 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[1]:
+
+
+# series 1
+# Create and display the given list 
+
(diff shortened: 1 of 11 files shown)


natasha-aleksandrova:
@johnroge please take a look at Files Changed tab, it shows some work that is not yours:
https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/1217/files

it is usually due to needing to sync your fork, please follow these steps and resubmit your PR:
https://help.github.com/articles/syncing-a-fork/

PR #1222, opened 2018-12-05
lesson 7 review

jayjonesseattle:
I am submitting lesson 7 for grading. I know that is not totally accurate but honestly I have had an extremely difficult time with this assignment and just want to move on. Not sure if its the time of day that I work on my assignment and my mind just isn't clicking or if the assignment just doesn't seem interesting to me or what it is. The wording is kind of difficult to follow and while the tutorial was helpful, I feel like it was still missing several key pieces and wished it would've completely walked me through each step so I have a better understanding. I think that I have a good understanding of the core concepts in the lesson as I have been able to apply them to code that I have written outside of the assignments to make sure I understood the class concepts but this assignment was super confusing and left me less confident in my skills versus confident. Thank you for your help.

### DIFF
--- a/students/terrance_jones/lesson7/html_render.py
+++ b/students/terrance_jones/lesson7/html_render.py
@@ -8,6 +8,7 @@ A class-based system for rendering html.
 # This is the framework for the base class
 class Element(object):
     tag = "html"
+    indent = " "
 
 
     def __init__(self, content=None, **kwargs):
@@ -21,7 +22,7 @@ class Element(object):
         self.contents.append(new_content)
 
 
-    def render(self, out_file):
+    def render(self, out_file, cur_ind=""):
         #loop through the list of contents:
         open_tag = ["<{}".format(self.tag)]
         for key, value in self.kwargs.items():
@@ -50,7 +51,7 @@ class Head(Element):
     tag = 'head'
 
 class OneLineTag(Element):
-    def render(self, out_file):
+    def render(self, out_file, cur_ind=""):
         #loop through the list of contents:
         out_file.write("<{}>".format(self.tag))
         out_file.write(self.contents[0])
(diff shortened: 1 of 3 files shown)


students/terrance_jones/lesson7/html_render.py
```diff
@@ -0,0 +1,96 @@
...
+        for content in self.contents:
+            try: 
+                content.render(out_file)
+            except AttributeError:
+                out_file.write(str(content))
+                out_file.write("\n")
+        out_file.write("</{}>\n".format(self.tag))
+
+class Html(Element):
+    tag = 'html'
```
natasha-aleksandrova: Step 8:
> Update the Html element class to render the " < !DOCTYPE html > " tag at the head of the page, before the html element.
You can do this by subclassing Element, overriding render(), but then calling the Element render from the new render.

Per above directions, you should define a `render` method in this class and it would look something like:
```
def render(self, file_out, cur_ind=""):
        file_out.write(cur_ind + "<!DOCTYPE html>\n")
        super().render(file_out, cur_ind=cur_ind)
```

students/terrance_jones/lesson7/html_render.py
```diff
@@ -0,0 +1,96 @@
...
+    tag = 'p'
+
+class Head(Element):
+    tag = 'head'
+
+class OneLineTag(Element):
+    def render(self, out_file, cur_ind=""):
+        #loop through the list of contents:
+        out_file.write("<{}>".format(self.tag))
+        out_file.write(self.contents[0])
```
natasha-aleksandrova: should have a loop here?

students/terrance_jones/lesson7/html_render.py
```diff
@@ -0,0 +1,96 @@
...
+class SelfClosingTag(Element):
+
+    def __init__(self, content=None, **kwargs):
+        if content is not None:
+            raise TypeError("SelfClosingTag can not contain any content")
+        super().__init__(content=content, **kwargs)
+
+
+    def render(self, out_file):
+        tag = self._open_tag()[:-1] + " />\n"
```
natasha-aleksandrova: I don't see where `_open_tag`  is defined.

jayjonesseattle:
Hi Natasha. Thank you for the help. When I have a moment I will revisit the
assignment and complete it.

natasha-aleksandrova:
sounds good!

PR #1226, opened 2018-12-07
Mail room ver 2. - Missing feature added

yilmazkennan:
Missing feature was added to the code.  Resubmitting the assignment

### DIFF
--- a/students/KennanY/Lesson3/mailroom.py
+++ b/students/KennanY/Lesson3/mailroom.py
@@ -1,0 +1,69 @@
+donations = [
+            ['Bill Gates', 165234, 42335.37, 21000,4400],
+            ['Jack Ma', 94834.44, 239102.23,9984],
+            ['Jeff Bezos', 1234534,234123],
+            ['Kate Jones', 948374.44, 987362.12],
+            ['Walt Kistler', 1234, 4444, 38476.34],
+            ['Jerry Paros', 110,44,85,8854]]
+
+def createReport():
+    """Print donation history"""
+
+    print('Donor name                        Total given       Number of gifts         Average gift')
+    print('=' *60)
+
+    for donor in donations:
+        total_amount=sum(donor[1:])
+        total_donation=len(donor)-1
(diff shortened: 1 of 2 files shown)


students/KennanY/Lesson3/mailroom.py
```diff
@@ -0,0 +1,85 @@
+donations = [
+            ['Bill Gates', 165234, 42335.37, 21000,4400],
+            ['Jack Ma', 94834.44, 239102.23,9984],
+            ['Jeff Bezos', 1234534,234123],
+            ['Kate Jones', 948374.44, 987362.12],
+            ['Walt Kistler', 1234, 4444, 38476.34],
+            ['Jerry Paros', 110,44,85,8854]]
+
+def createReport():
```
natasha-aleksandrova: functions (and var names) should be snake_case -> create_report

students/KennanY/Lesson3/mailroom.py
```diff
@@ -0,0 +1,85 @@
...
+        response=input("Please select and option (1,2 or 3)")
+        if str(response)=='1':
+            send_ThankYou()
+        elif str(response)=='2':
+            createReport()
+        elif str(response)=='3':
+            print('Finished ...')
+            break
+        else:
+            select=True
```
natasha-aleksandrova: you don't need the `select` var, it is never set to `False`.. use `while True` and your `break` statement will end the loop

natasha-aleksandrova:
@yilmazkennan adjusted your grade, thank you!

yilmazkennan:
Thank you for the feedback.  I am using PyCharm as my IDE.

natasha-aleksandrova:
@yilmazkennan if you're using Pycharm you should see a lot of underline type of warnings, for example:
screen shot 2018-12-10 at 8 06 30 am


those are the violations

yilmazkennan:
I dont see the errors but I used the tab key while using Charm.  I can see the vertical tab lines

Get Outlook for Android<https://aka.ms/ghei36>

PR #1230, opened 2018-12-08
adding lesson09, mailroom oop.

elnurad:


students/eldonab/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class Donor():
+    """create a class that represents a donor."""
+    def __init__(self, name, donations = None):
+        self.name = name
+        self.donations = donations
+
+
+    def add_donation(self, amount):
```
natasha-aleksandrova: note that module functions require two lines space, but Class methods require one line space between them.

elnurad:
Thanks, Natasha!

PR #1236, opened 2018-12-10
adding Lesson 7 assignment plus run file and test file

njschafi:
Submitting lesson 7 assignment, along with corresponding run and test files.


students/njschafi/Lesson07/html_render.py
```diff
@@ -0,0 +1,132 @@
...
+# This is the framework for the base class
+class Element(object):
+    """Main class for object"""
+    tag = 'html'
+    indent = '   '
+
+    def __init__(self, content=None, **kwargs):
+        """Default Object initializer"""
+        if content is not None:
+            self.stuff = [content]
```
natasha-aleksandrova: `self.content` would probably be a better choice 😉

PR #1244, opened 2018-12-12
Lesson 10

lauraannf:
mailroom_fp


students/lauraannf/Lesson10/mailroom_fp.py
```diff
@@ -0,0 +1,62 @@
...
+    def list_donations(self):
+        print(self.name + ' Donations: ' + ', '.join('${}'.format(d)
+              for d in self.donations))
+
+    @property
+    def last_donation(self):
+        return self.donations[-1]
+
+
+class Donor_List():
```
natasha-aleksandrova: Should have no underscores

students/lauraannf/Lesson10/mailroom_fp.py
```diff
@@ -0,0 +1,62 @@
...
+        return self
+
+    def add_donation(self, name, donation):
+        if name in self.donors:
+            self.donors[name].add_donation(donation)
+        else:
+            self.donors[name] = Donor(name, donation)
+
+    def is_donor(self, name):
+        if name in self.donors:
```
natasha-aleksandrova: Can be simplified to `return name in self.donors`

PR #1251, opened 2018-12-15
submit Lesson8 assignment

cocokaku:


students/cocokaku/lesson8/test_circle.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def test_circle_init_no_parameters_error():
+    msg = ""
+    try:
+        Circle()
+    except TypeError as e:
+        msg = e.args[0]
+    finally:
+        assert msg == "__init__() missing 1 required positional argument: 'radius'"
```
natasha-aleksandrova: use `pytest.raises` to verify raised exceptions:
https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

PR #1253, opened 2018-12-16
mailroom v3, try/except and list comprehension mods

geekwriter2:


students/MichaelM/lesson5/mailroom_rev_try_catch.py
```diff
@@ -0,0 +1,258 @@
...
+                   ["John Jacob Astor", 100.0],
+                   ["John Jacob Astor", 1000.0],
+                   ["John Jacob Astor", 10000.0],
+                   ["Alan Rufus, 1st Lord of Richmond", 10.0],
+                   ["Henry Ford", 10.0],
+                   ["Henry Ford", 100.0],
+                   ["Henry Ford", 1000.0],
+                   ["Cornelius Vanderbilt", 10.0],
+                   ["Jakob Fugger", 10.0],
+                   ["Jakob Fugger", 100.0]]
```
natasha-aleksandrova: at this point you should be using a dictionary data structure here which was also called out in the previous submission.

students/MichaelM/lesson5/mailroom_rev_try_catch.py
```diff
@@ -0,0 +1,258 @@
...
+    response_donation_amt = 0.0
+    sum_amt = 0.0
+    tmp_donor_list = list(list_donors())
+    user_response = input(
+        "-type 'list' to view a list of donors\n-type 'new' to  "
+        "add a donor to the list\n- type in an existing donor to add a new donation >")
+    if user_response.lower() == "list":
+        print()
+        # converted to list comprehension
+        [print(''.join([str(elem) for elem in row])) for row in list_donors()]
```
natasha-aleksandrova: this is considered an anti-pattern (print() function + comprehension, print function returns `None`), use comprehensions when you are capturing the result

students/MichaelM/lesson5/mailroom_rev_try_catch.py
```diff
@@ -0,0 +1,258 @@
...
+
+    if user_response.lower() == "new":
+        response_donor_name = input("What is the donor's name? >")
+        response_donation_amt = 0.0
+        while True:
+            try:
+                response_donation_amt = float(input("How much is {}'s donation? >".format(response_donor_name.title())))
+            except ValueError:
+                print("Please enter the donation in dollars")
+                continue
```
natasha-aleksandrova: no need for `continue` statement here

students/MichaelM/lesson5/mailroom_rev_try_catch.py
```diff
@@ -0,0 +1,258 @@
...
+
+    Parameters:
+    {none}
+
+    Returns:
+    distinct_donors (list): distinct donor list
+
+    """
+    # converted to list comprehension
+    donors = [i[0] for i in esteemed_donors]
```
natasha-aleksandrova: if you need a `set` then use set comprehension: `{i[0] for i in donors}`

PR #1257, opened 2018-12-18
Adding list_lab.py for lesson03

mattclau:
Completed list_lab exercise.  I made copies of the list for each series because it didn't make sense to me to use the modified list from the previous series especially when doubling the list would create an untenable list for use when asking about fruit preferences.

### DIFF
--- a/students/mattclau/lesson03/list_lab.py
+++ b/students/mattclau/lesson03/list_lab.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python3
+
+#*** Series 1 ***
+fruits = ['Apples', 'Pears','Oranges','Peaches']
+series1 = fruits[:]
+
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
yes, starting with the original fruit list for each task is the way to go!

students/mattclau/lesson03/list_lab.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+#***Series 3***
+#ask user about fruit preferences
+series3 = fruits[:]
+print(series3)
+answer = ''
+
+for fruit in series3[:]:
+    while answer not in ['Yes', 'No']:
```
natasha-aleksandrova: nice use of `not in`

PR #1258, opened 2018-12-18
Lesson 7 submittal

Percolator787:
Submitting html_render.py and test_html_render.py.  

The assert statements in test_html_render.py will all fail with the current version of html_render.py.  This is because the .htm files used for comparison don't include indentation which was added to html_render at the end.  The test_Element() assert only worked for Step 1.  Step 8 assert can be made to work by getting rid of indent by commenting out row 13 and uncommenting row 14.  Step 2-7 asserts can then be made to work by commenting out rows 78-80 so that the <!DOCTYPE html> tag does not display.

### DIFF
--- a/students/Dennis_Coffey/lesson07/html_render.py
+++ b/students/Dennis_Coffey/lesson07/html_render.py
@@ -1,0 +1,150 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Thu Nov 29 21:13:15 2018
+
+@author: dennis
+"""
+
+"""HTML renderer for Lesson 07"""
+
+class Element():
+    tag = '' # Element type
+    indent = 0 # Indent of tag
+    inline = False # Defines whether content is to be displayed on same line
+    empty = False  # Defines whether tag is an empty tag
+    def __init__(self, content=None, **kwargs):
+        self.content_list = []
+        self.kwargs = kwargs
+        if content:
+            self.content_list.append(content)
+        
+    def append(self, content):
+        self.content_list.append(content)
+        
+    def starttag(self, tag):
+        if tag:
+            tag = ' ' * self.indent + '<' + tag
+            if self.kwargs:
+                attribute = ''
+                for k in self.kwargs:
+                    attribute += ' ' + k + '=\"' + self.kwargs[k] + '\"'
+                tag += attribute
+            tag += '>'
+        return tag
+        
+    def endtag(self, tag):
+        if tag:
+            tag = ' ' * self.indent + '</' + tag + '>'
+        return tag
+        
+    def render(self, output_file, indent=0):
+        # Render opening tag
+        output_file.write(self.starttag(self.tag))
+        if not self.inline and not self.empty:
+            output_file.write('\n')
+        
+        # Render content
+        if self.content_list:
+            for item in self.content_list:
+                if type(item) == str:
+                    output_file.write(item)
+                else:
+                    item.render(output_file, indent + 4)
+                if not self.inline:
+                    output_file.write('\n')
+                
+        # Render closing tag
+        if not self.empty:
+            output_file.write(self.endtag(self.tag))
+        
+class Html(Element):
+    tag = 'html'
+    indent = 0
+    
+#    def render(self, output_file, indent=0):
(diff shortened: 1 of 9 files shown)


students/Dennis_Coffey/lesson07/html_render.py
```diff
@@ -0,0 +1,163 @@
...
+        
+    # Generate start tag, content and closing tag
+    def render(self, output_file, indent=0):
+        # Render opening tag
+        output_file.write(self.starttag(self.tag, indent))
+        if not self.inline and not self.empty:
+            output_file.write('\n')
+        
+        # Render content
+        if self.content_list:
```
natasha-aleksandrova: no need to check this, your for loop below will just attempt to iterate over empty list and nothing will happen

students/Dennis_Coffey/lesson07/html_render.py
```diff
@@ -0,0 +1,163 @@
...
+        
+    # Append content
+    def append(self, content):
+        self.content_list.append(content)
+        
+    # Start tag for element
+    def starttag(self, tag, indent):
+        if tag:
+            tag = self.indent_size * indent + '<' + tag
+            if self.kwargs:
```
natasha-aleksandrova: no need for this check, your for loop will iterate over empty dict and nothing will happen

students/Dennis_Coffey/lesson07/html_render.py
```diff
@@ -0,0 +1,163 @@
...
+        self.kwargs = kwargs
+        if content:
+            raise TypeError('Content is not allowed for Self Closing Tag elements')
+
+    def render(self, output_file, indent=0):
+        # Render opening tag
+        output_file.write(self.indent_size * indent + '<' + self.tag)
+        
+        # Render attribute
+        if self.kwargs:
```
natasha-aleksandrova: same here, no need for `if` check

students/Dennis_Coffey/lesson07/html_render.py
```diff
@@ -0,0 +1,163 @@
...
+            tag += '>'
+        return tag
+        
+    # End tag for element
+    def endtag(self, tag, indent):
+        if tag and not self.inline:
+            tag = self.indent_size * indent + '</' + tag + '>'
+        elif tag:
+            tag = '</' + tag + '>'
+        return tag
```
natasha-aleksandrova: consider simplified function for open and close tags:
```
    def make_tags(self):
        attrs = " ".join(['{}="{}"'.format(key, val) for key, val in self.kwargs.items()])
        if attrs.strip():
            open_tag = "<{} {}>".format(self.tag, attrs.strip())
        else:
            open_tag = "<{}>".format(self.tag)
        close_tag = "</{}>".format(self.tag)

        return open_tag, close_tag
```

students/Dennis_Coffey/lesson07/html_render.py
```diff
@@ -0,0 +1,163 @@
...
+# a tag
+class A(Element):
+    tag = 'a'
+    inline = True
+    def __init__(self, link, content, **kwargs):
+        self.content_list = []
+        self.kwargs = kwargs
+        self.kwargs['href'] = link
+        if content:
+            self.content_list.append(content)
```
natasha-aleksandrova: you should call parent class constructor here:
```
    def __init__(self, link, *args, **kwargs):
        kwargs['href'] = link
        super().__init__(*args, **kwargs)
```

students/Dennis_Coffey/lesson07/html_render.py
```diff
@@ -0,0 +1,163 @@
...
+# header tag
+class H(Element):
+    tag = 'h'
+    inline = True
+    def __init__(self, level, content, **kwargs):
+        self.content_list = []
+        self.kwargs = kwargs
+        self.tag = self.tag + str(level)
+        if content:
+            self.content_list.append(content)
```
natasha-aleksandrova: same here, call parent class constructor
```
def __init__(self, level, *args, **kwargs):
        self.tag = "h" + str(int(level))
        super().__init__(*args, **kwargs)
```

PR #1264, opened 2018-12-19
Trigram completed

yilmazkennan:


students/KennanY/Lesson4/trigram.py
```diff
@@ -0,0 +1,44 @@
...
+    #First two words become the key, next becomes value
+    counter=0
+    key=' '
+    for each in wordlist:
+        counter=counter+1
+        if counter%3 ==0:
+            trigramdata[key] = each
+            key=''
+        else:
+            key=key.strip() + ' ' + each
```
natasha-aleksandrova: it doesn't appear that they key is ever associated to more than 1 word since `append` is never used to append to the key?

also, the assignment asks to create a new story from the trigrams dict.

take a look at my version of trigrams solution to get an idea:
https://github.com/natasha-aleksandrova/Self_Paced-Online/blob/master/solutions/lesson4/trigrams.py

PR #1265, opened 2018-12-19
Lesson06 Assignmet

aguardia:


students/AleGuardia/Lesson06/test.py
```diff
@@ -0,0 +1,141 @@
+# tests for Mailroom_four.py by Alejandro Guardia
+
+from mailroom_four import return_average
+from mailroom_four import return_donors
+from mailroom_four import return_label
+from mailroom_four import return_report
+from mailroom_four import add_donor
+from mailroom_four import add_donation
+from mailroom_four import prompt_donors
+from mailroom_four import print_report
```
natasha-aleksandrova: note that you can do this in single import line:
`from mailroom_four import print_report, add_donor`

or 

```
import mailroom_four


mailroom_four.print_donor  # usage
```

students/AleGuardia/Lesson06/mailroom_four.py
```diff
@@ -0,0 +1,141 @@
...
+                print("Please enter a valid number")
+            else:
+                key_def()
+
+
+def print_report():
+    print()
+    report = return_report(donations)
+    print(report)
+    return report
```
natasha-aleksandrova: your `print_report` function should only do print and not return a value; the unit test should test the `return_report` function onlyl.

note: a better name would be `get_report` (instead of `return_report`)

students/AleGuardia/Lesson06/test.py
```diff
@@ -0,0 +1,141 @@
...
+    assert add_donor('Ale') == {'William Gates, III': [1000000, 585000, 5750000],
+                                'Mark Zuckerberg': [15000, 5000],
+                                'Jeff Bezos': [3000000],
+                                'Paul Allen': [25000, 1000],
+                                'Elon Musk': [30000, 3499],
+                                'Ale': []}
+
+
+def test_donation(monkeypatch):
+    monkeypatch.setattr('builtins.input', lambda x: "20.25")
```
natasha-aleksandrova: nice!

students/AleGuardia/Lesson06/mailroom_four.py
```diff
@@ -0,0 +1,141 @@
...
+        return send_note(name,add_donation(name),thank_you_note)
+    if name not in donations:
+        add_donor(name)
+        return send_note(name,add_donation(name),thank_you_note)
+
+
+def add_donor(name):
+    """Adds a donor name"""
+    donations[name] = []
+    return donations
```
natasha-aleksandrova: similar comment: this function should be void based on the name it implies.

your unit test can access the global dict `donations` (`from mailroom import donations`)

note: in real life scenario this would be a bit different but accessing the module level variable `donations` works here

PR #1270, opened 2018-12-21
Assignment 3 Finished

Iansberg:
Ready to merge

### DIFF
--- a/students/Sahlberg/Lesson3/Assignment3_StringSlicing.py
+++ b/students/Sahlberg/Lesson3/Assignment3_StringSlicing.py
@@ -34,3 +34,40 @@ def exchange_first_last(seq):
 
         else:
             return seqlist.format(seq[-1],*seq[1:-1], seq[0])
+			
+
+
+#print(exchange_first_last("hello"))
+#print(exchange_first_last((1,2,3,4,5)))
+#print(exchange_first_last(('12345')))
+
+
+
+def everyOther(val):
+    build = ''
+    buildElse2 = []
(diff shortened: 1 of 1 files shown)


students/Sahlberg/Lesson3/Assignment3_StringSlicing.py
```diff
@@ -0,0 +1,68 @@
...
+    return print(first + mid + last)
+
+def everyOther(val):
+    """This returns a sequence with every other value removed"""
+    print(val[::2])
+
+def foursOut(val):
+    """This returns a sequence with the first and last 4 values removed and every other value in between"""
+    valRange = val[4:-4]
+    return print(valRange[::2])
```
natasha-aleksandrova: * this return will return `None` because `print` returns `None`
* var names should be snake_case -> `val_range`
* you can do this in one line `return val[4:-4:2]`

students/Sahlberg/Lesson3/Assignment3_StringSlicing.py
```diff
@@ -0,0 +1,68 @@
...
+#Functions
+
+def exchange_first_last(seq):
+    """This returns the first and last item of a sequence exchanged"""
+    first = seq[-1:]
+    mid = seq[1:-1]
+    last = seq[:1]
+    return print(first + mid + last)
+
+def everyOther(val):
```
natasha-aleksandrova: pep8 style violation - use snake_case for function names -> `every_other`

PR #1276, opened 2018-12-22
adding Lesson 8 assignment files

njschafi:
Adding circle class assignment along with unit test file.


students/njschafi/Lesson08/test_circle.py
```diff
@@ -0,0 +1,179 @@
+"""test code for circle.py"""
+# NEIMA SCHAFI, LESSON 8 Assignment - Circle Class
+
+import pytest
+import math
+from circle import *
```
natasha-aleksandrova: avoid using "start" imports, instead -> `from circle import Circle`

PR #1286, opened 2018-12-25
PR for updates to mailroom2.py, Please review mailroom2.py

LouReis:
Hello Natasha,
I updated mailroom2.py based on your comments & recommendations.  Please review and let me know if there's any outstanding items.

Thanks & Regards,
Lou

Merry Christmas

### DIFF
--- a/students/LouReis/Lesson04/mailroom2.py
+++ b/students/LouReis/Lesson04/mailroom2.py
@@ -45,9 +45,19 @@ donations = {"Robin Hood": [50000, 50000, 50000], "Tycoon Reis": [25000000, 2500
 #     total = sum(donations[key])
 #     print(total)
 
+# Below is the main menu function that continues prompting until quit.
+def main_menu(main_prompt,menu_options_dict):
+    while True:
+        response = input(main_prompt)
+        menu_options_dict[response]()
+
+# Below are the 4 menu options that are declared in the dict.
+#
 # The following function takes the 'donations' dictionary and creates a sorted report.
 # Using a complex 'sorted' for loop.
-def calculate_sort(donations):
+def donation_report():
+    print('\nYou Chose Option 1\n\n')
+    print('DONATION SUMMARY REPORT\n\n')
     print('{:25} | {:^13} | {:^13} |   {:>13}'.format('Donor Name', 'Total Given', 'Num Gifts', 'Average Gift'))
     print('---------------------------------------------------------------------------')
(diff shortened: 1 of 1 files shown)


LouReis:
Thank you...

 

It was the first assignment working with dictionaries.  I didn't find the required reading & videos sufficient to develop the more advanced aspects of this assignment.  It takes time getting used to the compact & powerful syntax in some of the functionality.  The first time using the "sorted" function and it was a complex & powerful implementation but I had to spend a signifant amount of time researching & testing it to get it working right.  Navigating dictionaries is also complex when you have no prior use of them.  It will take more time to get fully familiar with basic syntax and I struggle with syntax errors on a regular basis...simplifying code is also going to take some time to get more familiar with, I sometimes have lengthier code because I struggle with the compact & powerful syntax...

 

Lou

 

Sent: Tuesday, December 25, 2018 at 8:20 PM

natasha-aleksandrova:
@LouReis I would recommend doing all of the optional assignments and labs if you think you need a little more practice. For example, the dict lab (under lesson 4) should help:
https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/dict_lab.html

Also, highly recommend referencing official python documentation; in the case of a `dict` it will contain all methods available and examples:
https://docs.python.org/3/library/stdtypes.html#dict

If you ever need help with understanding syntax or resolving syntax errors, you can post quick questions on Discussion board on the course website or in MS Team chat room.

natasha-aleksandrova:
As far as lengthier code... See my suggestions below, that is something I personally do as well:

Write out your code and get it to do what the requirements ask. Test and ensure it works. Write tests to assert correct results (which can be used after all of the refactoring you would do).

Read over your code - does it seem like there is duplicate code? If yes, refactor to remove duplicate code by combining some code blocks or extracting into helper functions.

Read over your code again - does it seem hard to read or complicated? If yes, investigate ways to simplify the code. Researching online is a great start. Someone probably had the same problem as you; SO will have a lot of really good suggestions that can be a learning tool. 

Are you able to easily understand what your functions do and what the variables mean? If no, then probably re-think your functions and rename the variables.

If possible, take a break and come back to your code to read over again - if you do not see any other improvements (via steps above) then go ahead and call it "done". 


----
I know this is lengthy but writing good code is a bit of an art 🙂and takes time to get good at - especially via Code Reviews like we do here. Hope this helps.

PR #1291, opened 2018-12-26
please review lesson 9

jayjonesseattle:
As always, I welcome your feedback. Thank you!

### DIFF
--- a/students/terrance_jones/lesson9/mailroom9.py
+++ b/students/terrance_jones/lesson9/mailroom9.py
@@ -1,0 +1,177 @@
+"""
+Mailroom9.py
+Written by Terrance Jones
+Assignment for lesson 9 UW Selfpaced Online course
+"""
+
+class Donor:
+    """Creates a donor with name and and empty dontation list """
+    def __init__(self, name, donations=None):
+        self._name = name
+        if donations is None:
+            self._donations = []
+        else:
+            self._donations = [donations]
+       
+    @property
+    def name(self):
+        """Returns donor name"""
+        return self._name
+
+    @property
+    def donations(self):
+        """Returns donations list for donor"""
+        return self._donations
+
+    def append_donations(self, value):
+        self.donations.append(value)
+
+    @classmethod
+    def thank_you_letter(self, donor_name, donation_amount):
+        dic = {'name': donor_name , 'amount_donated': donation_amount}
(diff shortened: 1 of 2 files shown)


students/terrance_jones/lesson9/mailroom9.py
```diff
@@ -0,0 +1,177 @@
...
+    @classmethod
+    def thank_you_letter(self, donor_name, donation_amount):
+        dic = {'name': donor_name , 'amount_donated': donation_amount}
+        return("Dear {name}, \n Thank you for your generous donation of ${amount_donated:,.2f}. \nSincerely, Mailroom.".format(**dic))
+
+
+class DonorCollection:
+
+    global donor_list
+    donor_list = {'Carlos Santos':[25,50,100], 'Esperanza Gomez': [10,20,30], 'Paul Jackson':[5,10,15], 'Karl Black':[100,200,300], 'Charles Exx': [15,30,45]}
```
natasha-aleksandrova: no need for global since you're using it as a class variable and can access via `DonorCollection.donor_list`... however, it no longer becomes "OO" as you're not utilizing objects.

students/terrance_jones/lesson9/mailroom9.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @classmethod
+    def donor_exists(self, donor):
+        return donor in donor_list
```
natasha-aleksandrova: in classmethods, you use `cls` var per convention instead of `self` - but these don't need to be classmethods as you don't need access to the class - these can be staticmethods BUT if all your methods are static methods then you're not utilizing OO (as per above comment)

students/terrance_jones/lesson9/mailroom9.py
```diff
@@ -0,0 +1,177 @@
...
+    @property
+    def donations(self):
+        """Returns donations list for donor"""
+        return self._donations
+
+    def append_donations(self, value):
+        self.donations.append(value)
+
+    @classmethod
+    def thank_you_letter(self, donor_name, donation_amount):
```
natasha-aleksandrova: the name and donations are part of your Donor object which should be used here. Something like:
```
# no decorator
def thank_you_letter(self):
    return f"Thank you {self.name} for {self.donations}"
```

natasha-aleksandrova:
merging this for now but this needs a bit more work.

jayjonesseattle:
Thank you. I will review the code and your suggestions and hopefully submit
the edits later tonight or tmrw morning.

PR #1292, opened 2018-12-27
refractored mail room for test_mail_room

april301975:
Refactored the summing each donor's donations, number of donation, and average. Also refactored inserting a donation into the dictionary of donors.

### DIFF
--- a/students/Thai_H/lesson6/mail_room_part4.py
+++ b/students/Thai_H/lesson6/mail_room_part4.py
@@ -1,0 +1,222 @@
+#!/usr/bin/env python3
+import sys
+import os
+from operator import itemgetter
+
+#--------------------------------------------------------------------------
+def send_thank_you_letter(donor_name, donation_amt):
+    #print(donor_name, donation_amt)
+    print(f'\nDear {donor_name}\n'
+         'Thank you for a recent donation to our Foundation in the amount of $' + f'{donation_amt}\n'
+        '\nSincerely'
+        '\nBill & Melinda Gates')
+
+#--------------------------------------------------------------------------
+def add_donation(donor, amt, membership):
+    if membership == 'N':
+        list_of_donors[donor] = []
+    list_of_donors[donor].append(amt)
+
+
+
+#---------------------------------------------------------------------------
+def is_existing_donor(donor_name):
+    return (donor_name in list_of_donors)
+
+
+#-----------------------------------------------------------------------------
+def send_thank_you_add_new_donation():
+    donor_name =input('\n\nPlease enter name of donor:  ')
+
+    try:
+        donation_amt = float(input('\nEnter the amount > '))
+        divide_by_zero = 1 / donation_amt
+    except ZeroDivisionError:
+        print ('Can\'t have zero donation amount')
+        send_thank_you_add_new_donation()
+    else:
+        print ('Amount validated: ', float(donation_amt))
+        #test_create_report()
+        #print('\n')
+
+    #if is_existing_donor(donor_name):
+    if donor_name in list_of_donors:
+        #print('existing donor is true. before add $$\n')
+        #test_create_report()
+
+        # Add amount to existing donor
+        #list_of_donors[donor_name].append(donation_amt)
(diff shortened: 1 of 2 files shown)


students/Thai_H/lesson6/mail_room_part4.py
```diff
@@ -0,0 +1,222 @@
...
+        'Q': quit_program
+    }
+        thank_you_selection = thank_you_menu_selection().upper()
+        thank_you_menu_function.get(thank_you_selection, invalid_thank_you_menu)()
+        #invalid_menu is to handle non-of-above choices
+                                                                #
+
+#----------------------------------------
+def sum_donation(d_amt):
+    return sum(d_amt)
```
natasha-aleksandrova: from the "learning how to unit test" point of view - yes this is a good way to practice that, however, this is only testing built-in `sum` function which we already know works.

You should concentrate on testing your code rather than internals - and no need to wrap built-in function with your own functions.

students/Thai_H/lesson6/test_mail_room_part4.py
```diff
@@ -0,0 +1,54 @@
...
+        assert os.path.isfile('William Boeing.txt')
+
+
+    #-------------------------------------------------------------------------
+    def test_add_donation(self):
+        mr4.init_donor_list()
+        expected = 20000
+        mr4.add_donation('Thai', 20000, 'N')
+        actual = mr4.list_of_donors['Thai'][0]
+        self.assertEqual(expected, actual)
```
natasha-aleksandrova: couple great things to point out:
* really good and clear test function names
* use of "actual" and "expected" which help clarifying the tests intent

students/Thai_H/lesson6/test_mail_room_part4.py
```diff
@@ -0,0 +1,54 @@
...
+        self.assertEqual(actual, expected)
+
+
+    #----------------------------------------------------------------------
+    def test_letter_file_name(self):
+        assert os.path.isfile('Charles Flint.txt')
+        assert os.path.isfile('Paul Allen.txt')
+        assert os.path.isfile('Steve Jobs.txt')
+        assert os.path.isfile('Thomas Edison.txt')
+        assert os.path.isfile('William Boeing.txt')
```
natasha-aleksandrova: this isn't executing any code - the reason this test would pass is because you have ran your code at some point and created those files.

your tests should **always** run your code then assert on something

students/Thai_H/lesson6/mail_room_part4.py
```diff
@@ -0,0 +1,222 @@
...
+#------------------------------------------------------------------------------
+def quit_program():
+    print('Bye!')
+    sys.exit()
+
+
+#-----------------------------------------------------------------------------
+def invalid_menu():
+    print('\nNot a valid option!\n\n')
+    main()
```
natasha-aleksandrova: you're in a while True loop.. main() will resume running on it's own - avoid creating additional nested loops

students/Thai_H/lesson6/mail_room_part4.py
```diff
@@ -0,0 +1,222 @@
...
+
+#------------------------------------------------------------------------------
+def create_report():
+    sorted_list = sort_donor_list(list_of_donors)
+    report_header = ['Donor Name', 'Total Donation Amt', 'Num of Donation',  'Avg Per Donation']
+    print('\n\n{:<25}{:>20}{:>20}{:>20}'.format(*report_header))
+    print('-'*90)
+    for donor in sorted_list:
+        print('{:<26}${:>15,.2f}{:>15}{:>25,.2f}'.format(donor[0], donor[1], donor[2], donor[1] / donor[2]))
+    print('\n\n')
```
natasha-aleksandrova: have this return a string (with report created) or a list (with report data) and unit test this function.

students/Thai_H/lesson6/test_mail_room_part4.py
natasha-aleksandrova: I can only really count this test as a good unit test.

PR #1297, opened 2018-12-29
Lesson 5 complete and Lesson 6 mailroom

stefanlund:
Lesson 6.
Changing the code to make testing possible was a nice help to create a better structure if I may say so myself.
More importantly, what do you think?


natasha-aleksandrova:
@stefanlund for Lesson 6 Unit Testing we're looking for mailroom code to be tested (except the user interaction code).

The main features that should be tested are "send a thank you" "create a report" and "send letters"; and really there should be little untested code.

Your unit tests only cover small "helper" functions which is great but those are usually easy to unit tests as it is; we're looking at a challenge of taking a program and making it unit testable, specifically for those big features.

students/stefan_lund/Lesson_5/Lesson_5_mailroom/mailroom_5_main.py
```diff
@@ -0,0 +1,112 @@
...
+                option_line = f"{letter_key}.)  {action}"
+                print(option_line)
+
+            answer = input("\nEnter letter according to your choice: ")
+
+            # check to see if the "answer" letter is one of the keys in the
+            # switch_option dictionary, good becomes the corresponding dict value
+            # and the loop is broken. If answer is not in dict, default set at
+            # False is maintained as the value of good.
+            good = switch_option.get(answer, False)
```
natasha-aleksandrova: catch KeyError here

students/stefan_lund/Lesson_5/Lesson_5_mailroom/mailroom_5_functions.py
```diff
@@ -0,0 +1,323 @@
...
+    name_of_letters = False
+    while not name_of_letters:
+        name = input("\nEnter name, 'First Last': ")
+        name_of_letters = is_letters_or_space(name)
+
+    # user enter new_amount, check if new_amount is a number, 2 decimals or less
+    digits_in_amount = False
+    while not digits_in_amount:
+        new_amount = input("\nEnter amount donated: ")
+        digits_in_amount = is_digit_or_period(new_amount)
```
natasha-aleksandrova: this is where you should replace the code with exceptions handling, which is really the right approach. Instead of writing complicated checking functions - try the user input:
```
while True:
    try:
        amount = float(input("..."))
    except ValueError:
        print("invalid amount value, try again")
    else:
        break
```

students/stefan_lund/Lesson_5/Lesson_5_mailroom/mailroom_5_functions.py
```diff
@@ -0,0 +1,323 @@
...
+    """
+        prints out a list of names currently in the "data" dictionary
+    """
+    data = mailroom_5_read_write_data.data
+
+    # determine width of fields to print in
+    name_list = data.keys()
+    name_length = []
+    # list comprehension
+    [name_length.append(len(name)) for name in name_list]
```
natasha-aleksandrova: this is incorrect usage of comprehension; see below for correct usage:
```
name_lengths = [len(name) for name in data]
```

students/stefan_lund/Lesson_5/Lesson_5_mailroom/mailroom_5_functions.py
```diff
@@ -0,0 +1,323 @@
...
+    e = p + (s * (amount_len))
+    lne += 3 * e
+    print(lne)
+
+    for name, donations in data.items():
+        number_of = 0
+        donation = 0
+        for date in donations:
+            donation += donations[date]
+            number_of += 1
```
natasha-aleksandrova: use `len` and `sum` functions here

students/stefan_lund/Lesson_5/Lesson_5_mailroom/mailroom_5_functions.py
stefanlund: # name_list = data.keys()
# name_length = []
# (wrong usage of )list comprehension
# [name_length.append(len(name)) for name in name_list]
# name_len = max(name_length) + 3
becomes
name_len = max([len(name) for name in data]) + 3

Thank you for showing me. 
I knew something was off but could not get it right. Much nicer and not only the comprehension but the use of data-dict as well.

PR #1305, opened 2018-12-31
modified file creation code to check for file successful creation

april301975:


students/Thai_H/lesson6/test_mail_room_part4.py
```diff
@@ -31,13 +31,20 @@ def test_avg_donation(self):
...
+
+        mr4.init_donor_list()
+        for donor_name in mr4.list_of_donors.keys():
+            file_name = donor_name + '.txt'
+            total_donation = sum(mr4.list_of_donors[donor_name])
+            thks_letter = open(file_name, 'w')
+            thks_letter.write('Dear {name}, \n'
+                          'Thank you for your generosity to our Foundation in the total amount of ${amt}'.format(name = donor_name,
+                                                                                                                 amt = total_donation))
+            thks_letter.close()
```
natasha-aleksandrova: you need to execute _your mailroom_ code - not write new code for the test, something like:
```
mailroom.create_thank_you_files()
...
assert os.path.isfile(file_name)
```

PR #1315, opened 2019-01-03
adding updates to lesson9

jayjonesseattle:
Hi Natasha,
Still not quite sure if I understand. I feel like I'm getting confused with the data collection class and it holding donor objects. specifically how I access the information for each object through the data collection class. in your example you used this at the end of the code:

donations = DonorCollection(Donor("natasha", [100]))
donations.add_donation("bob", [40])

How do I later access the donor information for "Natasha" if it was created through an instance of a donor collection but that is not a sub class of donor? not sure if i'm wording this right. Could you further expand on your previous comments?

Thanks!

### DIFF
--- a/students/terrance_jones/lesson9/mailroom9.py
+++ b/students/terrance_jones/lesson9/mailroom9.py
@@ -6,75 +6,68 @@ Assignment for lesson 9 UW Selfpaced Online course
 
 class Donor:
     """Creates a donor with name and and empty dontation list """
-    def __init__(self, name, donations=None):
-        self._name = name
-        if donations is None:
-            self._donations = []
-        else:
-            self._donations = [donations]
-       
-    @property
-    def name(self):
-        """Returns donor name"""
-        return self._name
+    def __init__(self, name, donation):
+        self.name = name
+        self.donations = [donation]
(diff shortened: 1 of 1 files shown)


students/terrance_jones/lesson9/mailroom9.py
```diff
@@ -6,75 +6,68 @@
...
-            average = float(round(total / num_gifts, 2))
-            dic = {'name':name, 'total':total, 'num_gifts':num_gifts, 'average':average }
-            new_row = "{name:<25} ${total:^10} {num_gifts:^10} ${average:^10}".format(**dic)
-            report.append(new_row)
-            new_row = ""
-            report.append(new_row)
-        return (report)
 
 
 class cli_main:
```
charlesrobison: Class name should be CapWords: class CliMain for example

students/terrance_jones/lesson9/mailroom9.py
```diff
@@ -6,75 +6,68 @@
...
 
-    @classmethod
```
charlesrobison: This class is missing an __init__ method - is this intended?

PR #1323, opened 2019-01-06
Finished mailroom.py assignment.

Iansberg:
thanks.

### DIFF
--- a/students/Sahlberg/Lesson3/mailroom.py
+++ b/students/Sahlberg/Lesson3/mailroom.py
@@ -12,44 +12,74 @@ donor = {'bob johnson': [150.00],
          'jon jacob':[5000.00, 5.00]
          }
 
-def donate():
+def donate(name, dictionary):
     """Appends a donation amount to a given donor."""
     amount = int(input('How much did they donate?'))
-    return amount, donor[option].append(amount)
+    return amount, dictionary[name].append(amount)
 
-def print_list():
+def print_list(dictionary):
     """Prints our the donor list"""
-    return print(list(donor.keys()))
+    return print(list(dictionary.keys()))
 
 def user_options():
     """Main user options menu choices."""
     user_action = input('Please choose from the following options: Send a Thank You, Create a Report, quit')
     return user_action
 
-def email():
-    return print(f'\n\nThank you {option} for the generous donation of $ {amount}. We appreciate your generosity.\n\nSincerely, \n\nThe Helping R Us Team\n\n')
+def email(name, amount):
+    return print(f'\n\nThank you {name.title()} for the generous donation of $ {amount:.2f}. We appreciate your generosity.\n\nSincerely, \n\nThe Helping R Us Team\n\n')
 
-print(list(donor.keys()))
+def quit():
+    return print('Until next time!')
 
-user_action = user_options()
+def sort_and_stats_dict(dictionary):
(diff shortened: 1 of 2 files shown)


students/Sahlberg/Lesson3/mailroom.py
```diff
@@ -0,0 +1,90 @@
...
+            f'{dictionary[count][0]:<20}$    {dictionary[count][1][0]:<23.2f}{dictionary[count][1][1]:<15}    ${dictionary[count][1][2]:>12.0f}')
+        count += 1
+
+if __name__== '__main__':
+    user_action = user_options()
+    if user_action.lower() == 'quit':
+        quit()
+
+    #Add users and donations, send thank you email when done.
+    while user_action == 'Send a Thank You':
```
charlesrobison: You can do a lot to make this a bit more user friendly, like while user_action is equal to lowercase 's' for example, so the user can input 'Send' 'send' 'S' or 's' for example.

students/Sahlberg/Lesson3/mailroom.py
```diff
@@ -0,0 +1,90 @@
...
+            if option.lower() in donor:
+                amount = int(donate(option,donor)[0])
+                print(donor)
+                email(option,amount)
+
+            else:
+                donor[option] = []
+                amount = int(donate(option,donor)[0])
+                print(donor)
+                email(option,amount)
```
charlesrobison: This portion is not working for me - unable to add a donor with amount.  Is this working for you?  I'm getting a TypeError:  'tuple' object does not support item assignment

students/Sahlberg/Lesson3/mailroom.py
```diff
@@ -0,0 +1,90 @@
...
+    amount = int(input('How much did they donate?'))
+    return amount, dictionary[name].append(amount)
+
+def print_list(dictionary):
+    """Prints our the donor list"""
+    return print(list(dictionary.keys()))
+
+def user_options():
+    """Main user options menu choices."""
+    user_action = input('Please choose from the following options: Send a Thank You, Create a Report, quit')
```
charlesrobison: Recommend not using 'quit' as it's a reserved word.  Can make things a bit complicated.

students/Sahlberg/Lesson3/mailroom.py
```diff
@@ -0,0 +1,90 @@
...
+    while user_action == 'Create a Report':
+        donor_sorted = sort_and_stats_dict(donor)
+        print('Donor Name       |      Donor Total    |  Count of Donations    |  Average of Donations')
+        print('----------------------------------------------------------------------------------------')
+        count = 0
+        for donor in donor_sorted:
+            print(
+                f'{donor_sorted[count][0]:<20}$    {donor_sorted[count][1][0]:<23.2f}{donor_sorted[count][1][1]:<15}    ${donor_sorted[count][1][2]:>12.0f}')
+            count += 1
+        break
```
charlesrobison: Ideally, you'll want to continue the loop until the user selects to quit.

students/Sahlberg/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,114 @@
...
+                email(option,amount)
+
+    while user_action == '2':
+        donor_sorted = sort_and_stats_dict(donor)
+        print_report(donor_sorted)
+        break
+
+    while user_action == '3':
+        email_file(donor)
+        break
```
charlesrobison: This one works quite well - nice job!

PR #1338, opened 2019-01-09
Adding two new files: kata.py and mailroom.py for Lesson 04

dmartin7856:
These two files are the assignment submissions for Lesson 04.

I apologize in advance for seemingly incomplete code.  I wrack my brain over simple tasks and fall behind quickly.  So I'd rather submit something and move along to the next Lesson than try to get something perfect.

### DIFF
--- a/students/dominic_martin/lesson04/mailroom.py
+++ b/students/dominic_martin/lesson04/mailroom.py
@@ -1,0 +1,71 @@
+#!usr/bin/env Python3
+
+donation_list = {'name': ('Ted Bayer', 'Panfila Alvarez', 'JR Reid', 'Simon Laplace', 'Jennifer Meyers'), 'donation': (500, 700, 330, 440, 800)}
+donation_list1 = {'Ted Bayer': 500, 'Panfila Alvarez': 700, 'JR Reid': 330, 'Simon Laplace': 440, 'Jennifer Meyers': 800}
+
+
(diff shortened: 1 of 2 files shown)


students/dominic_martin/lesson04/mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+def thanks():
+	global donation_list
+	b = int(input("Who would you like to send a thanks to?\n(1) - Ted Bayer\n(2) - Panfila Alvarez\n(3) - JR Reid\n(4) - Simon Lapalce\n(5) - Jennifer Meyers\n(6) - Enter a new name\nEnter your choice:"))	
+	nameIdx = (int(b - 1))
+	if nameIdx < 5:
+		donamt(nameIdx)
+	if nameIdx == 5:
+		newDonor(nameIdx)
+
+def newDonor(nameIdx):
```
charlesrobison: Function names should be snake_case per PEP8: def new_donor(name_id_x) for example

students/dominic_martin/lesson04/mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+	global donation_list
+	newName = str(input("Enter a new name:"))
+	print()
+	donation_list['name'][nameIdx] = newName
+	donation_list['donation'][nameIdx] = 0
+	print()
+	donamt(nameIdx)
```
charlesrobison: This isn't adding new donors to the database.

PR #1341, opened 2019-01-10
Assignment 4 scripts

Iansberg:
This is for both sections of assignment 4 (mailroom2 and kata).


Iansberg:
@charlesrobison should I just include the text file I used? I will test and get back to you. How do I submit a new (PR?)?

Iansberg:
@charlesrobison I see the problem I think. I am trying to return a couple hundred words as the assignment says to do. The sherlock_small.txt is only 188 words.  See line 65. Should I do anything or can I still turn in as is? It works for me and says to use "a book-sized file".

Iansberg:
@charlesrobison Sorry about all the @s. I updated my code to be more dynamic. How do I re-submit?

-Ian

charlesrobison:
@Iansberg - just do a new PR and I'll review - thanks much!

PR #1359, opened 2019-01-16
PR for Lesson010 & Lesson09, mailroom5.py & mailroom_fp.py

LouReis:
Hello Natasha,

Welcome back...

Please review the following files for Lesson09 & Lesson010:

Lesson09:
mailroom5.py
test_mailroom5.py

Lesson010:
mailroom_fp.py
test_mailroom_fp.py

Please provide detailed feedback and comments if you see anything I need to rework.  I'm not sure if Lesson09 mailroom5.py is designed just right in regards to the way I did the class.

Thanks & Regards,
Lou

### DIFF
--- a/students/LouReis/Lesson010/mailroom_fp.py
+++ b/students/LouReis/Lesson010/mailroom_fp.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3.7
-# mailroom2.py
+# mailroom_fp.py
 # Coded by LouReis
+# Lesson010
 
 """
 Write a small command-line script called mailroom.py.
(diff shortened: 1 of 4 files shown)


natasha-aleksandrova:
Lou,

As far as Lesson 9 (which is the more important one here... it's important to have a good understanding of OO)
your `Donor` class on it's own looks good but mailroom as a whole and how it utilizes OO needs a bit more work.

for example in line 128:
```
        print ('{:25} ${:>15,.2f} {:>15} ${:>15,.2f}'.format(key, Donor(key).total_donations, Donor(key).num_donations, Donor(key).avg_donation))
```
you are creating 3 instances of `Donor`, this should be a *single* instance as it's the same donor.

Also, ideal solution has two Classes which is also mentioned in the assignment description.

Please review two below resources.

more details guidelines:
https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom-oo.html#the-program
example two classes stubbed out:
https://repl.it/@naaleksandrova/Lesson9Example

PR #1361, opened 2019-01-17
final lesson 03 files uploaded

vmedina1014:


students/V_Medina/lesson_03/mailroom.py
```diff
@@ -4,35 +4,66 @@
...
+        donor_report.append([donor[0], donor_total, amount_count, avg_donation])
+    print('Donor Name      | Total Given |   Num Gifts  | Average Gift')
+    print('------------------------------------------------------------')
+    for donor in donor_report:
+        print('{:<15} ${:>13} {:>14} ${:>13,.2f}'.format(*donor))
+    return None
+
+
+if __name__ == "__main__":
+    Donors = [
```
natasha-aleksandrova: should be lower case `donors`

PR #1365, opened 2019-01-18
Lesson 2 files

alheadstrong:
Hi there - adding the lesson 2 files for review. Note that for the grid printer, I added some optional parameters to use different characters or whole strings for the grid characters. If that was a function I wanted to use more, I would add some exception handling around the optional parameters or a function to convert arguments into strings to be used....but I left it as-is for now. 

Also, a question about documentation - I picked up a habit of comment blocks at the top of scripts with some file information. With git, I wonder if that's still appropriate - what do you think?

And last, I'm still getting the hang of "Pythonic" code, so would appreciate ways to simplify what I've got.

Thanks!

Alex

### DIFF
--- a/students/alheadstrong/lesson_2/grid_printer.py
+++ b/students/alheadstrong/lesson_2/grid_printer.py
@@ -1,8 +1,10 @@
 '''Author: Alex Filson
-Updated: 1.15.19
+Updated: 1.18.19
 Grid Printer Exercise for Lesson 2
 Py210, Online Self-Paced
 '''
+
+#brute force grid
 print('+ '+'- '*4+'+ '+'- '*4+'+')
 print('| '+' '*8+'| '+' '*8+'|')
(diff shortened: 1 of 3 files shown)


students/alheadstrong/lesson_2/grid_printer.py
```diff
@@ -0,0 +1,54 @@
...
+print('| '+' '*8+'| '+' '*8+'|')
+print('| '+' '*8+'| '+' '*8+'|')
+print('| '+' '*8+'| '+' '*8+'|')
+print('| '+' '*8+'| '+' '*8+'|')
+print('+ '+'- '*4+'+ '+'- '*4+'+')
+
+def print_grid(n):
+    '''single parameter grid. Parameter determines cell size by number of
+    dash - space pairs between nodes horizontally and "|" rows vertically.
+    '''
```
natasha-aleksandrova: use triple double quotes for doc strings

PR #1368, opened 2019-01-20
PR for Lesson09 Made some improvements to mailroom_oo

LouReis:
Hello Natasha,

I made some improvements but have a few issues with mailroom_oo.py:

-Attempt to use "find_donor" method but is says Donor doesn't have 'name' attribute (but it does)
-My get_report is not sorted, is there an easy way to do this with class objects?  I cannot use the sorted function as I did with dictionaries.

Thanks,
Lou

### DIFF
--- a/students/LouReis/Lesson09/mailroom_oo.py
+++ b/students/LouReis/Lesson09/mailroom_oo.py
@@ -1,0 +1,259 @@
+#!/usr/bin/env python3.7
+# mailroom_oo.py
+# Coded by LouReis
+# Lesson09
+
+"""
+Refactor the following code to use classes.
+
+Write a small command-line script called mailroom.py.
+It should have a data structure that holds a list of your donors and a history
+of the amounts they have donated. This structure should be populated at first
+with at least five donors, with between 1 and 3 donations each.
+The script should prompt the user (you) to choose from a menu of 3 actions:
+“Send a Thank You”, “Create a Report” or “quit”)
+
+Report should look like the following:
+Donor Name                | Total Given | Num Gifts | Average Gift
+------------------------------------------------------------------
+Joe Donor                  $  653784.49           2  $   326892.24
+John Rich                  $   16396.10           3  $     5465.37
+Sally Neighbor             $     877.33           1  $      877.33
+Mack Jack                  $     708.42           3  $      236.14
+
+"""
+
+# The below snippet will display the instances in memory.
+"""
+ import gc
+ for obj in gc.get_objects():
+     if isinstance(obj, Donor):
+         print (obj.name, obj.donations)
+"""
+
+from functools import total_ordering
+import sys
+import math
+from textwrap import dedent
+
+donations = {}
+
+# Utility so we have data to test with, etc.
+def enter_sample_data():
+    """
+    Returns a list of donor objects to use as sample data to populate the DB
+    """
+    return [Donor("Robin Hood", [50000, 50000, 50000]),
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
there are various examples online on how to sort objects, here is one:
https://www.pythoncentral.io/how-to-sort-a-list-tuple-or-object-with-sorted-in-python/

natasha-aleksandrova:
this is your `find_donor` function:
```
    def find_donor(self, name):
        """
        find a donor in the donor db
        :param: the name of the donor
        :returns: The donor data structure -- None if not in the self.donor_data
        """
        return self.donor_data.get(Donor.name(name))
```

`Donor.name` tries to access class variable `name` which does not exist
`Donor().name` tries to access instance attribute `name` which does exist
`self.donor_data` holds reference to all your donor objects - you should _only_ use this for all of your data access:
```
def find_donor(self, name):
    self.donor_data.get(name)  # this will return Donor object or None
```

students/LouReis/Lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,259 @@
...
+    def list_donors(self):
+        """
+        creates a list of the donors as a string, so they can be printed
+        Not calling print from here makes it more flexible and easier to
+        test
+        """
+        listing = ["Donor list:"]
+        for donor in self.donors:
+            listing.append(donor.name)
+        return "\n".join(listing)
```
natasha-aleksandrova: `self.donors` doesn't exist.
cleaner versions:
`return ["Donor list:"] + list(self.donor_data)`

students/LouReis/Lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def add_donor(self, name):
+        """
+        Add a new donor to the donor db
+        :param: the name of the donor
+        :returns: the new Donor data structure
+        """
+        donor = Donor(name)
+        self.donor_data[donor.name] = donor
+        return donor
```
natasha-aleksandrova: you shouldn't have a return in this function

students/LouReis/Lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,259 @@
...
+        :returns: string with letter
+        note: This doesn't actually write to a file -- that's a separate
+              function. This makes it more flexible and easier to test.
+        """
+        return dedent('''Dear {0:s},
+              Thank you for your very kind donation of ${1:.2f}.
+              It will be put to very good use.
+                             Sincerely,
+                                -The Team
+              '''.format(donor.name)
```
natasha-aleksandrova: this needs to go into `Donor` class.

students/LouReis/Lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,259 @@
...
+    donation = 0
+    if donor in Donor:
+        print("You have entered an existing donor:", donor)
+        try:
+            donation = float(input("Please enter the donation amount '0.00':"))
+        except ValueError:
+            print("\n\n----------You have entered an invalid value, returning to Main Menu----------\n\n")
+            main_menu(main_prompt,menu_options_dict)
+        temp = Donor(donor)
+        temp.add_donation(donation)
```
natasha-aleksandrova: this needs to be in DonorDb class

LouReis:
Thanks for the feedback, I appreciate it...I'll review & update accordingly...

 

Lou

 

Sent: Saturday, January 19, 2019 at 6:46 PM

PR #1370, opened 2019-01-20
PR for Lesson09 mailroom_oo.py Made updates

LouReis:
Hello Natasha,

I've updated the mailroom_oo.py program per your comments, getting closer.  Here's a few difficulties now:

-add_donor method fails and states it's missing 'name' but I am passing name, don't understand
-donations don't seem to be getting added to the DB as the get_report is not reflecting them

Please review & provide feedback

Thanks & Regards,
Lou

### DIFF
--- a/students/LouReis/Lesson09/mailroom_oo.py
+++ b/students/LouReis/Lesson09/mailroom_oo.py
@@ -98,6 +98,24 @@ class Donor():
         amount = float(amount)
         self.donations.append(amount)
 
+
+    # The below gen_letter is another way to generate a letter.
+    def gen_letter(self, donor):
+        """
+        Generate a thank you letter for the donor
+        :param: donor tuple
+        :returns: string with letter
+        note: This doesn't actually write to a file -- that's a separate
+              function. This makes it more flexible and easier to test.
+        """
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
```
  File "student.py", line 255, in <module>
    main_menu(main_prompt,menu_options_dict)
  File "student.py", line 58, in main_menu
    menu_options_dict[response]()
  File "student.py", line 212, in thanks_letter
    DonorDB.donate_letter(donations)
  File "student.py", line 188, in donate_letter
    DonorDB.add_donor(name)
TypeError: add_donor() missing 1 required positional argument: 'name'
```
Lou, I would recommend going back to review OO concepts in general _and_ the solution I sent you - in this case as well you need to execute `add_donor` method on an instance. 
To create an instance of a class -> `DonorDB()`, full statement -> `DonorDB().add_donor(name)`
however that would still be incorrect because you would be create a *new* instance but you need to work with existing instance you already created.

The solution I sent you creates instance named `db` and only works with that instance.

The reason donations aren't getting updated is because you are creating a new Donor object but it is not linked with your DonorDB class. DonorDB instance needs to handle updating donations. Again, please review the solution I shared with you.

PR #1373, opened 2019-01-21
lesson8, circle_class ready for grading

geekwriter2:


students/MichaelM/lesson8/circle_class.py
```diff
@@ -0,0 +1,166 @@
...
+        return a
+
+    @property
+    def circumference(self):
+        c = round(Circle.pi * int(self.diameter),2)
+        return c
+
+    @classmethod
+    def from_diameter(cls, d):
+        cls.cls_diameter = d
```
natasha-aleksandrova: you don't need to set this

PR #1378, opened 2019-01-22
Lesson 2 assignment

ethanenguyen:
Submitted grid print exercise


students/ethan_nguyen/lesson02/gridprint.py
```diff
@@ -0,0 +1,47 @@
+import sys
+import os
```
natasha-aleksandrova: probably don't need these imports?

ethanenguyen: no I don't.  I was just testing :)

PR #1397, opened 2019-01-31
added lesson 10 functions

jaredmulholland:
Not sure why my last two tests of the projection method are not working with on the command line. They functions do work if I run them line by line.

### DIFF
--- a/students/jared_mulholland/lesson_6/mail_room_class_2.py
+++ b/students/jared_mulholland/lesson_6/mail_room_class_2.py
@@ -1,0 +1,324 @@
+"""import required modules"""
+
+import numpy as np
+import os
+import sys
+
+"""
+
+Notes for mail room class testing:
+
+1. Class for individual donor
+    a. contains, first name, last name, donation amount
+    b. method to append to donations
+    c. method to send a thank you
+"""
+
+class Donor(object):
+    """takes full name and creates donor record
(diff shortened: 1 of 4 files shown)


students/jared_mulholland/lesson_6/mail_room_class_2.py
```diff
@@ -0,0 +1,324 @@
...
+        self.donor_list = [donor.fullname]
+        self.donor_dict = {donor.fullname: donor.donation}
+        self.file_path = file_path
+        
+    @property
+    def add_donor(self):
+        return self.donor_list
+    
+    @add_donor.setter
+    def add_donor(self, new_donor):
```
natasha-aleksandrova: these should be functions not properties.

students/jared_mulholland/lesson_6/test_mail_room_class_2.py
```diff
@@ -0,0 +1,156 @@
...
+    assert factored_amount ==  dg_test.projection(factor)
+
+
+def test_factored_donation_min_max():
+
+    factor = 2
+    min_don = 11
+    max_don = 99
+    factored_amount = 280
+    assert factored_amount == dg_test.projection(factor, min_don, max_don)
```
natasha-aleksandrova: dg_test object should be created in the test as it could have weird side-effects

PR #1401, opened 2019-02-01
submit - lesson6 mailroom5.py with edit to search accross any directory 

NetMic:
given that the folder name containing the donations of each donor

### DIFF
--- a/students/Net_Michael/lesson6/mailroom4.py
+++ b/students/Net_Michael/lesson6/mailroom4.py
@@ -11,30 +11,44 @@ from collections import defaultdict
 
 def quit_sel():
     raise SystemExit(1)
+#new_donor_folder is a folder containing the text files of all donors.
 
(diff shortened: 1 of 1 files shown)


students/Net_Michael/lesson6/mailroom4.py
```diff
@@ -11,30 +11,44 @@
...
+#root = os.path.dirname( os.getcwd())
+#root = "\\".join(root.split("\\")[:2])
+# to save time at least provide the first folder in
+# c drive where the donor_list file is located,
+# which is the later option
 
 # given *\*\directory_a\cwd, lists everything in directory_a
 def list_dir_path():
     root = os.path.dirname( os.getcwd())
+    root = "\\".join(root.split("\\")[:2])
```
natasha-aleksandrova: I would use `os` module for file operations.

PR #1402, opened 2019-02-01
Mailroom Lesson 4

mattclau:
In the end, I didn't use the dictionary for creating the letter as I wanted to have a flexible letter that would allow me to create a letter for everyone using a for loop or just for 1 letter when choosing option 1 to send a thank you.  Sending the whole dictionary didn't give me that flexibility.

### DIFF
--- a/students/mattclau/lesson04/mailroom.py
+++ b/students/mattclau/lesson04/mailroom.py
@@ -32,7 +32,7 @@ def get_donation():
     donation = ''
     while not donation.replace('.','',1).isdecimal():
             donation = input('\nEnter donation amount: ')
-            if donation.replace('.','',1).isdecimal():s
+            if donation.replace('.','',1).isdecimal():
                 return donation
             else:
                 print('\n***Error! Use only numbers for the donation amount.***')
--- a/students/mattclau/lesson04/mailroom.py
+++ b/students/mattclau/lesson04/mailroom.py
@@ -1,0 +1,124 @@
+#Mailroom script that creates canned responses for donations and prints a report of all donations
+from operator import itemgetter
+
+#Formatting characters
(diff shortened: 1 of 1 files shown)


students/mattclau/lesson04/mailroom.py
```diff
@@ -0,0 +1,138 @@
...
+        filename = os.path.join(filepath,donor.replace(' ','_') + '.txt')
+        outfile = open(filename, 'w')
+        outfile.write(create_letter(donor))
+        outfile.close()
+
+
+
+def quit():
+    """Function that exists the program when called"""
+    from sys import exit
```
natasha-aleksandrova: import statement should go at the top of the file.

students/mattclau/lesson04/mailroom.py
```diff
@@ -0,0 +1,138 @@
...
+    print('\n')
+
+def send_all():
+    """Sends letters to all donors when called"""
+    filepath = input('\nEnter the path where the files should be created.  Hit enter to create in directory of program: ')
+    for donor in donors:
+        filename = os.path.join(filepath,donor.replace(' ','_') + '.txt')
+        outfile = open(filename, 'w')
+        outfile.write(create_letter(donor))
+        outfile.close()
```
natasha-aleksandrova: use `with` statement when working with files.

mattclau:
> @mattclau when you talk about using a dictionary with letter.. are you referring to using `.format()` with the donor dictionary?

@natasha-aleksandrova yes, the **dict passes in the whole dictionary and I wanted to only pass 1 donor since the thank you letter is for 1 donor only and making a separate function to handle the whole dictionary seemed like a waste unless I wasn't understanding.

PR #1407, opened 2019-02-03
Added assignments for lesson3

ethanenguyen:
For the mailroom assignment, I created a class to represent a donor with its attributes, name, donated amount, number of donation and average per donation.  I then created a dictionary with key as the donor's name and value as a donor object.   I didn't see the tutorial on the mailroom assignment when I started the assignment but I could have created a list of donor with tuple.  Please let me know if my way is over complicated and more difficult to debug.

thanks,

### DIFF
--- a/students/ethan_nguyen/lesson03/mailroom.py
+++ b/students/ethan_nguyen/lesson03/mailroom.py
@@ -1,0 +1,87 @@
+
+import operator
+'''
+Class to store Donor object and its attributes
+'''
+class Donor:
+    def __init__(self, name, amount=0, num=0):
+        self.name = name
+        self.amount_gift = amount
+        self.num_gift = num
+    '''function to calculate average donation'''
+    def getAverge(self):
+        return (self.amount_gift/self.num_gift)
+    '''function to return number of donations'''
+    def getCount(self):
+        return self.num_gift
+    '''function to return historical amount'''
+    def getAmount(self):
+        return self.amount_gift
+    def getName(self):
+        return self.name
+    '''function to add new donation amount to historical'''
+    def addAmount(self, addition):
+        self.amount_gift += addition
+        self.num_gift += 1
+
+if __name__ == "__main__":
+    #create a dict object to store donors
+    DonorDict = {}
+    D1 = Donor("William Gates, III", 100.2, 2)
+    DonorDict['William Gates, III'] = D1
+    D2 = Donor("Mark Zuckerberg", 3000.2, 3)
+    DonorDict['Mark Zuckerberg'] = D2
+    D3 = Donor("Jeff Bezos", 100000.2, 3)
+    DonorDict['Jeff Bezos'] = D3
+    D4 = Donor("Paul Allen", 5000.24, 1)
+    DonorDict['Paul Allen'] = D4
+    D5 = Donor("Donald Trump", 50000.24, 1)
+    DonorDict['Donald Trump'] = D5
+    
+    #sort = sorted(DonorDict.items(), key=operator.itemgetter(1).getAmount(), reverse = True)
+
+    response = input("Please select S for \"Send a Thank You\" C for \"Create a Report\" or Q to quit >")
+    while ("S" or "C" in response.upper()):
+        if 'S' in response.upper():
+            print("Do Thank you")
+            inputName = input("Please enter a name > ")
+            while("list" in inputName):
+                for k,v in DonorDict.items():
+                    print(k)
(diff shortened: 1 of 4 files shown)


natasha-aleksandrova:
@ethanenguyen I would highly recommend following the instructions for the assignments and using data structures learned in the lessons. 

Lesson 3 should be using list/tuple data structures
Lesson 4 should be using a dictionary
Lesson 9 uses classes

in addition, there are certain things we look for in each mailroom revisions as far as grading goes.

it is more challenging to give review on OO implementation in lesson 3 since I don't expect good OO concepts at this point.

students/ethan_nguyen/lesson03/slicing.py
```diff
@@ -0,0 +1,48 @@
...
+'''
+def exchange_first_last(seq):
+    a_new_sequence = seq[-1:]+seq[1:-1]+seq[:1]
+    return a_new_sequence
+
+'''
+function with every other item removed
+'''
+def remove_other(seq):
+    a_new_sequence = seq[0:len(seq):2]
```
natasha-aleksandrova: this can be simplified to `seq[::2]`

students/ethan_nguyen/lesson03/slicing.py
```diff
@@ -0,0 +1,48 @@
...
+    return a_new_sequence
+
+'''
+function with the first 4 and the last 4 items removed, and then every other item in between
+'''
+def remove_first_last_four(seq):
+    a_new_sequence = seq[0:-4][4:]
```
natasha-aleksandrova: this can be simplified to `seq[4:-4]`

students/ethan_nguyen/lesson03/mailroom.py
```diff
@@ -0,0 +1,87 @@
...
+    def getName(self):
+        return self.name
+    '''function to add new donation amount to historical'''
+    def addAmount(self, addition):
+        self.amount_gift += addition
+        self.num_gift += 1
+
+if __name__ == "__main__":
+    #create a dict object to store donors
+    DonorDict = {}
```
natasha-aleksandrova: variable names are snake_case -> `donor_dict`

also, at this point in the lesson your data structure should be list/tuple combo or just a list

students/ethan_nguyen/lesson03/mailroom.py
```diff
@@ -0,0 +1,87 @@
...
+    D3 = Donor("Jeff Bezos", 100000.2, 3)
+    DonorDict['Jeff Bezos'] = D3
+    D4 = Donor("Paul Allen", 5000.24, 1)
+    DonorDict['Paul Allen'] = D4
+    D5 = Donor("Donald Trump", 50000.24, 1)
+    DonorDict['Donald Trump'] = D5
+    
+    #sort = sorted(DonorDict.items(), key=operator.itemgetter(1).getAmount(), reverse = True)
+
+    response = input("Please select S for \"Send a Thank You\" C for \"Create a Report\" or Q to quit >")
```
natasha-aleksandrova: this needs to be inside of your while loop. suggested code layout:
```
while True:
    response = input("...").upper()
    if response == "S":
        send_thank_you()
    elif response == "C":
        create_report():
    elif response == "Q":
        quit_program()
```

students/ethan_nguyen/lesson03/mailroom.py
ethanenguyen: Hi Natasha,

Thank you for your feedback.  I will update mailroom.py to use list/tuple combo.  I was following the instructions on https://startlearning.uw.edu/courses and it doesn't have the more detailed instructions as https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom-part1.html.   It looks like we should follow the exercises posted on the latter link? 

thanks,

natasha-aleksandrova: I was referring to these guidelines on the course website:
```
Guidelines
First, factor your script into separate functions. Each of the above tasks can be accomplished by a series of steps. Write discreet functions that accomplish individual steps and call them.

Second, use loops to control the logical flow of your program. Interactive programs are a classic use-case for the while loop.

Of course, input() will be useful here.

Put the functions you write into the script at the top.

Put your main interaction into an if __name__ == '__main__' block.

Finally, use only functions and the basic Python data types you’ve learned about so far. There is no need to go any farther than that for this assignment.```


yes you can refer to github pages for more detailed information 🙂

PR #1412, opened 2019-02-04
Lesson 4 Pull Request

alheadstrong:
I've updated mailroom as mailroom_part2 with dictionaries, as well as fixing the report function to sort by total given, which was not included in previous version. I am passing the donor database as an unused argument into my "goodbye" function to keep it part of the dictionary - based user menu. Is that acceptable, or is there a better way to do it? 

For the trigram assignment, I used "Heart of Darkness" from Project Gutenberg, but didn't include that in the pull request - please let me know if you'd like to see that as well. 

Thanks,
Alex

### DIFF
--- a/students/alheadstrong/lesson_4/mailroom_part2.py
+++ b/students/alheadstrong/lesson_4/mailroom_part2.py
@@ -1,5 +1,6 @@
 import os
 
+
 def new_thank_you(donor_dict):
     """Add donation amount to new or existing donor and generate a thank you text."""
     while 1:
@@ -21,7 +22,7 @@ def new_thank_you(donor_dict):
                 donor_dict[current_dict['donor_name']] = [current_dict['donation']]
             break
 
-    print('\n\n',create_letter(current_dict))
+    print('\n\n', create_letter(current_dict))
     return donor_dict
 
 
@@ -68,9 +69,10 @@ def report(donor_dict):
 
 
 def letters_to_all(donor_dict):
+    """Writes form letter to txt files based on input dictionary. Optional directory change."""
     user_input = input("Would you like save to current directory,{}? >(y,n)".format(os.getcwd()))
     if user_input == 'n':
-        path = input("Enter full path to desired directory.")
+        path = input("Enter full path to desired directory: ")
         os.chdir(path)
 
     for key in donor_dict.keys():
@@ -83,6 +85,7 @@ def letters_to_all(donor_dict):
 
 
 def create_letter(letter_dict):
+    """Returns form letter string based on input dictionary. Dictionary needs 'donor_name' and 'donation' keys"""
     letter_dict['fleebs_floobed'] = round(letter_dict['donation']/20)
     return ('Dear {donor_name},'
             '\n\n\tThank you for your generous gift of ${donation:,.2f}.'
--- a/students/alheadstrong/lesson_4/mailroom_part2.py
+++ b/students/alheadstrong/lesson_4/mailroom_part2.py
@@ -1,0 +1,85 @@
+def thank_you(donor_dict):
+    """Add donation amount to new or existing donor and generate a thank you text."""
+    while 1:
+        donor_name = input("Enter donor name:")
+        if donor_name == 'list':
+            for d in donor_dict:  # display all donor names on user input 'list'
+                print(d[0])
+                continue
+        else:
+            for d in range(len(donor_dict)):
+                if donor_dict[d][0] == donor_name:
+                    print("Donor name found in database.")
+                    donation = int(input('Enter donation amount:'))
+                    donor_dict[d].append(donation)
+                    break
+            else:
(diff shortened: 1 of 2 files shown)


students/alheadstrong/lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,128 @@
...
+    return 'quit'
+
+
+def main():
+    """Establish initial donor database and initiate top user menu. """
+    ddb = {'Archie Bunker': [20, 100, 75, 98],
+           'Beyonce Knowles': [2000000, 50000000],
+           'Charlie Kauffman': [12345],
+           'David Sedaris': [23000, 1200, 2000],
+           'Edvard Munch': [1, 2, 3]}
```
natasha-aleksandrova: re your question in the PR:
in this particular case I would make this data structure global; if you move this declaration above where you call `main()` then you don't need to pass it around.

students/alheadstrong/lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,128 @@
...
+
+    while ddb != 'quit':
+        user_input = input('\n\nMENU:\n'
+                           '1 - Send a Thank You for a New Donation\n'
+                           '2 - Create a Report\n'
+                           '3 - Send letters to everyone\n'
+                           '4 - Quit\n'
+                           'Please enter 1-4>')
+        user_menu = {'1': new_thank_you, '2': report, '3': letters_to_all, '4': goodbye}
+        user_options = ('1', '2', '3', '4')
```
natasha-aleksandrova: you don't need this line, instead:
`if user_input in user_menu` -> you already have available options in the dictionary

students/alheadstrong/lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,128 @@
...
+        if user_input in user_options:
+            ddb = user_menu.get(user_input)(ddb)
```
natasha-aleksandrova: this is certainly bad practice; your function should only return a value if it's intended to. 

as per above comment, move ddb into module level and you will have access to it in all of your functions.

also, for your quitting function add a sys.exit call:
```
def goodbye():
    sys.exit()  # you will need import sys at the top of the file
```
this way you don't need to special case handle this function.

students/alheadstrong/lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,128 @@
...
+        num_gifts = len(donor_dict[key])
+        average_gift = total_given/num_gifts
+        new_report_line = (key, '$' + str(total_given), str(num_gifts), '$' + str(average_gift))
+
+        for i in range(len(report_table)):
+            if total_given > float(report_table[i][1][1:]):
+                report_table = report_table[:i]+[new_report_line]+report_table[i:]
+                break
+        else:
+            report_table = report_table + [new_report_line]
```
natasha-aleksandrova: you should use a built-in `sorted()` function for sorting.

students/alheadstrong/lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,128 @@
+import os
+
+
+def new_thank_you(donor_dict):
+    """Add donation amount to new or existing donor and generate a thank you text."""
+    while 1:
```
natasha-aleksandrova: `while True` is a more common pattern.

students/alheadstrong/lesson_4/mailroom_part2.py
```diff
@@ -0,0 +1,128 @@
...
+        current_dict = {'donor_name': input("Enter donor name, or type 'list' for current donor list:")}
+        if current_dict['donor_name'] == 'list':
+            for d in donor_dict:  # display all donor names on user input 'list'
+                print(d + ': ', donor_dict[d])
+                continue
```
natasha-aleksandrova: no need for this `continue` statement

students/alheadstrong/lesson_4/trigram.py
```diff
@@ -0,0 +1,99 @@
...
+        else:
+            d[key] = [t_list[i+2]]
+    return d
+
+
+def generate_start_key(d):
+    """Return a two-word key that begins a sentence from a dictionary."""
+    start_words = []
+    for key in d.keys():
+        if key[0][-1:] == ('.' or '?' or '!') and len(key[1:]) > 0:
```
natasha-aleksandrova: are you check here the the value returned by `key[0][-1:]` is either `'.'`  or `'?'` or `'!'` ?
that's not what this is doing, instead you want:
```
if key[0][-1:] in ('.' or '?' or '!')
```

I can explain further if you'd like me to.

natasha-aleksandrova: yes, it seems like it could work but it does something a little different.

in case of `a or b or c` it will test each var at a time, and will stop at first "true" var .. so in your case `'.'` will always evaluate to true so it will only ever compare against `'.'`

PR #1428, opened 2019-02-11
Added assignments for lesson04, mailroom2 and trigram

ethanenguyen:
Please see solutions for assignment 04, mailroom2 and trigram.

### DIFF
--- a/students/ethan_nguyen/lesson04/trigram.py
+++ b/students/ethan_nguyen/lesson04/trigram.py
@@ -1,0 +1,97 @@
+
+import sys
+import random, os
+
+
+def build_trigrams(words):
+    '''
+    build up the trigrams dict from the list of words
+    returns a dict with:
+       keys: word pairs
+       values: list of followers
+    '''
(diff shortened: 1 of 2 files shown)


students/ethan_nguyen/lesson04/trigram.py
```diff
@@ -0,0 +1,97 @@
...
+def build_text(word_pairs):
+
+    '''
+    build up a new story with the trigram dict
+    '''
+
+    story_words = list()
+    #randomely select two staring words
+    start_key = random.choice(list(word_pairs))
+    [story_words.append(item) for item in start_key]
```
natasha-aleksandrova: this should actually be:
```

story_words = [item for item in start_key]
```
a comprehension should include an assignment, if there is no assignment then it should be a regular loop.

In above case you don't really need a comprehension, `start_key` is a tuple and looks like you want to initialize a list with that? this is what you want
```
story_words.extend(list(start_key))
```

PR #1439, opened 2019-02-17
final version for mailroom

HawkeyeUW:
Updated mailroom_pt2 with final version.  Submitted it to Self-learning site.

### DIFF
--- a/students/jvirtue/lesson04/triagram.py
+++ b/students/jvirtue/lesson04/triagram.py
@@ -1,0 +1,90 @@
+#Lesson 4 Assignment 2
+#Triagrams
+#Jason Virtue 02/15/2019
+#UW Self Paced Python Course
+
+import random
+
+source = "sherlock_holmes.txt"
+
+def read_infile(source= source):
+    with open(source, "r") as infile:
+        book = infile.read()
+    return book
+
+def format_book(book):
+    book = book.replace("\n", " ")
+    book = book.replace("\\", " ")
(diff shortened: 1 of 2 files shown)


students/jvirtue/lesson04/triagram.py
```diff
@@ -0,0 +1,90 @@
...
+    book = book.replace("\\", " ")
+    words = book.split(" ")
+    while "" in words:
+        words.remove("")
+    for i in range(len(words)):
+        while " " in words[i]:
+            words[i] = words[i].replace(" ", "")
+        while "\\" in words[i]:
+            words[i] = words[i].replace("\\", "")
+        while "\\\'" in words[i]:
```
natasha-aleksandrova: no need for these `while` loops (lines 22, 24, 26). `replace` will replace all instances that you have without a loop.

students/jvirtue/lesson04/triagram.py
```diff
@@ -0,0 +1,90 @@
...
+            words[i] = words[i].replace("\\\'", "'")
+    return words
+
+def make_dictionary(lst):
+    dic = {}
+    for i in range(len(lst)- 2):
+        two_words = lst[i] + " " + lst[i + 1]
+        value = lst[i+2]
+        if two_words not in dic:
+            dic[two_words] = [value,]
```
natasha-aleksandrova: `,` is only needed for tuples

PR #1440, opened 2019-02-18
Lesson09 for John Rogers.

johnroge:
mailroom.py and donors.py from lesson09


students/john_rogers/lesson09/donors.py
```diff
@@ -0,0 +1,130 @@
...
+    @property
+    def number_of_donations(self):
+        return len(self.donations)
+
+    @property
+    def avg_donation_amount(self):
+        return round(self.total_donations / self.number_of_donations, 2)
+
+    def new_donation(self, amount):
+        self.donations.append(amount)
```
natasha-aleksandrova: :thumbsup:

students/john_rogers/lesson09/donors.py
```diff
@@ -0,0 +1,130 @@
...
+        self.donors.append(donor)
+
+    def donor_names(self):
+        return [d.full_name for d in self.donors]
+
+    def total_amount_raised(self):
+        return sum([sum(d.donations for d in self.donors)])
+
+    def total_number_donations(self):
+        return sum([len(d.donations) for d in self.donors])
```
natasha-aleksandrova: you should use Donor methods here: `[d. number_of_donations for d in self.donors]`

students/john_rogers/lesson09/donors.py
```diff
@@ -0,0 +1,130 @@
...
+            self.donors = donors
+
+    def add_donor(self, donor):
+        self.donors.append(donor)
+
+    def donor_names(self):
+        return [d.full_name for d in self.donors]
+
+    def total_amount_raised(self):
+        return sum([sum(d.donations for d in self.donors)])
```
natasha-aleksandrova: you should use Donor methods here: `[d. total_donations for d in self.donors]`

students/john_rogers/lesson09/donors.py
```diff
@@ -0,0 +1,130 @@
...
+        create a form letter
+        :param amount: donation amount
+        :param name: donor name
+        :return: form letter filled in with donor and amount
+        """
+        letter = (
+            f'Hey {name}, thanks for your donations! '
+            f'As of today, {DonorDataBase.today}, you have donated a total of '
+            f'${amount}.'
+        )
```
natasha-aleksandrova: since this method deals with a single donor it should live in `Donor` object:
```
def letter_text(self):
    return (f'Hey {self.full_name} you have donated {self.total_donations}')
```

students/john_rogers/lesson09/mailroom.py
```diff
@@ -0,0 +1,115 @@
...
+        "5: save a thank you note to disk for each donor\n"
+        ">>> "
+    )
+
+    main_dispatch = {
+        '1': exit_menu,
+        '2': donor_actions,
+        '3': DDB.print_summary,
+        '4': DDB.thank_all,
+        '5': DDB.save_report,
```
natasha-aleksandrova: these methods should be called on the instance (`donor_db`); `{'4': donor_db.thank_all}`

students/john_rogers/lesson09/mailroom.py
```diff
@@ -0,0 +1,115 @@
...
+"""
+mailroom.py: use classes where applicable
+Author: JohnR
+Version: 1.0
+Last updated: 2/17/19
+Notes:
+"""
+
+from donors import DonorDataBase as DDB
+from donors import Donor as D
```
natasha-aleksandrova: avoid aliases unless there is a good reason - _especially_ single letter variables.

students/john_rogers/lesson09/mailroom.py
```diff
@@ -0,0 +1,115 @@
...
+        except KeyError:
+            print('Please enter a number between 1 and 5.')
+
+
+def exit_menu(data):
+    print()
+    save = input('Enter Y to save all data to disk before exiting. ')
+    save = save.lower()
+    if save == 'y':
+        DDB.save_report(data)
```
natasha-aleksandrova: `DonorDatabase.save_report` is a function (object)
`DonorDatabase().save_report` is a method of the class

since you're using OO you should be working with an instance and executing a method of the instance : `donor_db.save_report()`

you can use `DonorDatabase.save_report()` with staticmethods which don't work on the object/instance and are simple functions that relate to the object and that's why one would define them inside of a class.

johnroge:
Lots of really good feedback, thank you Natasha. I've incorporated most of the feedback but may need another day or two to re-work my menu and clean up a few items. Will merge again soon.

Thanks again,
John

natasha-aleksandrova:
@johnroge sounds good 👍

johnroge:
Completed, appreciate any additional feedback.

PR #1444, opened 2019-02-20
Assignment 4/5 Submission

vmedina1014:


vmedina1014:
Assignment 4 submission

students/V_Medina/lesson_04/mailroom_2.py
```diff
@@ -0,0 +1,97 @@
...
+
+if __name__ == "__main__":
+    donors = {
+        'Victor': [100, 20, 30],
+        'John': [12],
+        'Kevin': [91, 32],
+        'Kelly': [5, 21],
+        'Matt': [75, 20],
+        'Josh': [31, 3],
+        'Micah': [120]}
```
natasha-aleksandrova: 👍

students/V_Medina/lesson_04/mailroom_2.py
```diff
@@ -0,0 +1,97 @@
...
+    # initial request
+    response = input('Do you want to: "Send a Thank you", "Create a Report", "Send letters to everyone" or "quit"? ')
+
+    response_func_dict = {'send a thank you': send_thank_you,
+                          'create a report': create_report,
+                          'send letters to everyone': send_letters,
+                          'quit': sys.exit}
```
natasha-aleksandrova: 👍 (nice job on mapping to `sys.exit`

students/V_Medina/lesson_04/mailroom_2.py
```diff
@@ -0,0 +1,97 @@
...
+    while response != 'quit':
```
natasha-aleksandrova: this check is redundant since you're already mapping "quit" to `sys.exit`; it should look more like this:
```
while True:
    response = input("...")
    response_func_dict[response]()
```
that's it!

students/V_Medina/lesson_04/mailroom_2.py
```diff
@@ -0,0 +1,97 @@
...
+    :param donors: dictionary
+        Dictionary of donor names with a value of a list of donation amoounts
+    :return: None
+        function writes out text files. No return is needed
+    """
+    for donor in donors:
+        with open(str(donor) + '.txt', 'w') as file:
+            file.write('Dear {},\n\nThank you for your very kind donation of ${}!\n\nSincerely,\n -The Team'.
+                       format(donor, donors[donor][-1]))
+        file.closed
```
natasha-aleksandrova: this is not needed

students/V_Medina/lesson_04/mailroom_2.py
```diff
@@ -0,0 +1,97 @@
...
+    return None
```
natasha-aleksandrova: no need to return None

students/V_Medina/lesson_04/mailroom_2.py
```diff
@@ -0,0 +1,97 @@
...
+        donor_report[donor] = dict(total_donation=sum(donors[donor]), num_donations=len(donors[donor]),
+                                   avg_donation=sum(donors[donor]) / len(donors[donor]))
+    print(donor_report)
+    print('Donor Name      | Total Given |   Num Gifts  | Average Gift')
+    print('------------------------------------------------------------')
+    for donor in donor_report:
+        print('{:<15} ${:>13} {:>14} ${:>13,.2f}'.format(donor, donor_report[donor]['total_donation'],
+                                                         donor_report[donor]['num_donations'],
+                                                         donor_report[donor]['avg_donation']))
+    return None
```
natasha-aleksandrova: * no need to `return None` 
* I don't see sorting feature here? please add the missing feature.

students/V_Medina/lesson_04/mailroom_2.py
```diff
@@ -0,0 +1,97 @@
...
+        donor = name
+    else:
+        donors[name] = []
+        donor = name
+
+    donation_amount = int(input('Donation Amount? '))
+    donors.get(donor).append(donation_amount)
+    print(donors)
+    print('Thank you {} for your generous donation of {} dollars!'.format(donor, donation_amount))
+    return donors
```
natasha-aleksandrova: this function should not return anything here

students/V_Medina/lesson_04/mailroom_2.py
vmedina1014: what if the function requires a parameter such as the donors dictionary in my case?

vmedina1014:
I think i made the edit. Let me know if i still need any action on my end.

Thanks

students/V_Medina/lesson_04/mailroom_2.py
natasha-aleksandrova: In your case it actually isn't required which is why I omitted it. Your `donors` dictionary is accessible in your functions globally.

There are certainly other cases (other type programs) that could require an argument passed just not in this particular case.

natasha-aleksandrova:
If you look at files changed tab it still shows deleted file.

this is the commit:
https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/1444/commits/1f7efb7edeaf2decd339bdaefd3e8a19f837b888

did you do these steps?:
https://help.github.com/en/articles/syncing-a-fork

if you did then it looks like you might have explicitly deleted the file.

vmedina1014:
Yes i did. So am I all set with that pull request?

natasha-aleksandrova:
The PR still includes a file deletion in another student in folder, you can see that in Files Changed tab.

Normally I would not merge this without fixing your issue however this student is no longer active so I'm ok with that.

PR #1445, opened 2019-02-20
Added files for lesson05

ethanenguyen:
Please see exce3pt_exercise.py and mailroom03.py files for lesson05.   For mailroom03 exercise, I implemented the try catch logic to handle bad user's input for string name and numeric amount.  I can't find anywhere to implement the list comprehension.   I'm using dict to store donor's name and donation and I only print the key and values.

### DIFF
--- a/students/ethan_nguyen/lesson05/mailroom3.py
+++ b/students/ethan_nguyen/lesson05/mailroom3.py
@@ -1,0 +1,113 @@
+
+import sys, os
+
+#create a donor_db as a dict
+donor_db = {"William Gates, III": [653772.32, 12.17],
+            "Jeff Bezos": [877.33],
+            "Paul Allen": [663.23, 43.87, 1.32],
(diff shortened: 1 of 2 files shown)


students/ethan_nguyen/lesson05/mailroom3.py
```diff
@@ -0,0 +1,113 @@
...
+    
+    print("Done!")
+
+if __name__ == "__main__":
+    
+    arg_dict = {'S':send_thank_you, 'C':create_report, 'L':send_letters}
+
+    while True:
+        response = input("Please select S for \"Send a Thank You\" C for \"Create a Report\" L for \"Send Letters\" or Q to quit >").upper()
+        arg_dict.get(response, quit_program)()
```
charlesrobison: Nice job on the added exceptions.  Regarding the comprehensions, you could try and convert the for loops, but it doesn't always make sense to add them.  This works well.

PR #1448, opened 2019-02-21
Adds solutions for Chapter 4 activities

scottebetts:


students/scottebetts/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python
+from time import strftime
+
+donors = [['Tim Berners-Lee', 120000.00, 3, 40000.00],
+          ['Ada Lovelace', 600.00, 2, 300.00],
+          ['Marie Curie', 1000.00, 2, 500.00],
+          ['Nicola Tesla', 20000.00, 4, 5000.00],
+          ['Hans Holbein', 300.00, 3, 100.00]]
```
natasha-aleksandrova: in lesson 4 you should be using a dictionary for the main data structure.

students/scottebetts/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+            break
+
+
+def thanks_menu():
+    menu_selection(thanks_prompt, thanks_dispatch)
+
+
+def quit():
+    print("Quitting this menu now")
+    return "exit menu"
```
natasha-aleksandrova: I would recommend to use `sys.exit()` here instead of return statement then your main loop becomes even simpler

students/scottebetts/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+        file_date = strftime("%Y%m%d")
+        filename_format = "{}_{}.txt"
+        donor_name = donor[0].replace(' ', '_')
+        donor_file = filename_format.format(donor_name, file_date)
+        with open(donor_file, 'w+') as outfile:
+            outfile.write(
+                "Dear {},\n "
+                "Thanks for your recent gift of ${:>.2f}, "
+                "you great big gullible maroon.".format(donor[0], donor[1]))
+            outfile.close()
```
natasha-aleksandrova: no need for `close()` - this is handled by `with` block.

PR #1449, opened 2019-02-22
Lesson 3, Mailroom Assignment Part 1

ganttArt:


students/christopher_gantt/lesson03/mailroom.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+
+donor_db = [("Wassily Kandinsky", [43928.13, 131.34, 1928.0]),
+            ("Jasper Johns", [3134.43, 153.34]),
+            ("Mark Rothko", [135353.33]),
+            ("Richard Serra", [153757.87, 28457.12, 1293451.0]),
+            ("Yves Tanguy", [1534.23, 2542.19])]
```
natasha-aleksandrova: 👍

students/christopher_gantt/lesson03/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+
+prompt = "\n".join(("Type a number from the options below",
+        "1 - Send a Thank You",
+        "2 - Create a Report",
+        "3 - Quit\n"))
```
natasha-aleksandrova: nice use of `join`

students/christopher_gantt/lesson03/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+
+def thank_you():
+    thanks = True
+    while thanks == True:
```
natasha-aleksandrova: use `is` instead of `==`

`is` is used for object comparison (Ex. `is None` `is True`)

however, here you can just simplify to `while thanks:`

ganttArt: Thanks for the feedback Natasha

PR #1451, opened 2019-02-22
lesson03 exercises and assignment

bromleysj:
get_donor_summary feels to me a bit too "hacky"; I read in the directions we were to use only a data structure we've learned thus far so I used a list of lists

### DIFF
--- a/students/scott_bromley/lesson03/mailroom.py
+++ b/students/scott_bromley/lesson03/mailroom.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python3
+import datetime
+
+# prompt user to select from mailroom menu: send a thank you, create a report, or quit
+# send a thank you: allows user to select a name, adds donor if not in donor list, otherwise ask for donation
+# and thanks donor
+# create a report summarizes the donors and their donations in a table
+# quit ends the program
+
+
+def main():
+    donors = [["David Einhorn", "Mark Zuckerberg", "Seth Klarman", "Bill Ackerman"],
+                 [1800.18, 36036.36],
+                 [7272.72, 1818.18, 545454.54],
+                 [180.00, 72.72, 18.00],
+                 [324.32]]
+    menu = ("Send a Thank You", "Create a Report", "quit")
+    mailroom_open = True
+    while mailroom_open is True:
+        response = get_user_input(f"Please enter an option from the following menu: \n {menu} > ").strip()
+        if response.casefold() == "send a thank you":
+            full_name = get_user_input("Please enter the full name of a donor: > ").strip()
+            if full_name in donors[0]:
+                donation = float(get_user_input("Please enter the donation amount: > ").strip('$'))
+                donors[donors[0].index(full_name) + 1].append(donation)
+                send_thank_you(full_name, donation)
+                continue
+            else:
+                add_donor(full_name, donors)
+                continue
+        elif response.casefold() == "create a report":
+            create_report(donors)
+            continue
+        elif response.casefold() == "quit":
+            mailroom_open = False
+        else:
+            print(f"The option you entered '{response}' is invalid, please try again.")
+            continue
+
+
+def send_thank_you(full_name, donation):
+    '''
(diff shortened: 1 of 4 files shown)


students/scott_bromley/lesson03/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+# create a report summarizes the donors and their donations in a table
+# quit ends the program
+
+
+def main():
+    donors = [["David Einhorn", "Mark Zuckerberg", "Seth Klarman", "Bill Ackerman"],
+                 [1800.18, 36036.36],
+                 [7272.72, 1818.18, 545454.54],
+                 [180.00, 72.72, 18.00],
+                 [324.32]]
```
natasha-aleksandrova: alternatively you could use a list of tuples like:
```
donors = [('David E', [100, 200]), ('Mark Z', [200, 300])]
```

students/scott_bromley/lesson03/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+    mailroom_open = True
+    while mailroom_open is True:
+        response = get_user_input(f"Please enter an option from the following menu: \n {menu} > ").strip()
+        if response.casefold() == "send a thank you":
+            full_name = get_user_input("Please enter the full name of a donor: > ").strip()
+            if full_name in donors[0]:
+                donation = float(get_user_input("Please enter the donation amount: > ").strip('$'))
+                donors[donors[0].index(full_name) + 1].append(donation)
+                send_thank_you(full_name, donation)
+                continue
```
natasha-aleksandrova: you don't believe you need any of the continue statements in here

students/scott_bromley/lesson03/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+
+def main():
+    donors = [["David Einhorn", "Mark Zuckerberg", "Seth Klarman", "Bill Ackerman"],
+                 [1800.18, 36036.36],
+                 [7272.72, 1818.18, 545454.54],
+                 [180.00, 72.72, 18.00],
+                 [324.32]]
+    menu = ("Send a Thank You", "Create a Report", "quit")
+    mailroom_open = True
+    while mailroom_open is True:
```
natasha-aleksandrova: can be simplified by using `while True` and using `break` statement on line 35

students/scott_bromley/lesson03/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+    :param donors: list of donors and donations
+    :return: lists of names, totals, number of gifts, avg gift sorted by highest total
+    '''
+    if donors is None:
+        donors = []
+    totals = [float(format(sum(donations), '.2f')) for donations in donors[1:]]
+    num_gifts = [len(gifts) for _, gifts in sorted(zip(totals, donors[1:]), reverse=True)]
+    avg_gift = [float(format(sum(donations) / len(donations), '.2f')) for _, donations in sorted(zip(totals, donors[1:])
+                                                                                                 , reverse=True)]
+    names = [names for _, names in sorted(zip(totals, donors[0]), reverse=True)]
```
natasha-aleksandrova: you have 4 loops above - they can be condensed into a single loop:
```
for donor in donors:
    # compute totals, num_gifts, avg_gift

sorted_donors = # do sorting here
```
return sorted list then loop over and simply print

students/scott_bromley/lesson03/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def add_donor(donor_name, donors=None):
+    '''
+    :param donor_name: full name of donor
+    :param donors: list of donors
+    :return: updated list of donors
+    '''
+    if donors is None:
+        donors = []
```
natasha-aleksandrova: you probably don't need these lines since you're always passing in donors --- even then you don't need to pass donors into your functions if you define the structure in the `if __name__ ==' __main__'` block

PR #1454, opened 2019-02-24
Lesson 6 Pull Request

alheadstrong:
I used pytest exclusively, rather than unittest, and pulled my previous iteration of "mailroom" into many more small testable functions. 

I checked file writing content, but not any of the user input or the sys.exit. Let me know if you'd like to see either implemented and I'll give it a go!

### DIFF
--- a/students/alheadstrong/lesson_6/test_mailroom_part4.py
+++ b/students/alheadstrong/lesson_6/test_mailroom_part4.py
@@ -1,0 +1,91 @@
+from mailroom_part4 import *
+
+
+def test_1():
+    assert is_name_list('not_list') is False
+
+
+def test_2():
+    assert is_name_list('list') is True
+
+
+def test_3():
+    assert is_name_list('') is False
+
+
+def test_4():
+    assert is_name_existing({'donor_name': 'David Sedaris'}) is True
+
+
+def test_5():
+    assert is_name_existing({'donor_name': 'daVid sedaRiS'}) is True
+
+
+def test_6():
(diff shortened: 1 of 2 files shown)


students/alheadstrong/lesson_6/test_mailroom_part4.py
```diff
@@ -0,0 +1,91 @@
+from mailroom_part4 import *
+
+
+def test_1():
+    assert is_name_list('not_list') is False
+
+
+def test_2():
```
natasha-aleksandrova: your test names should have meaningful names that are easy for the reader to know what you're testing.

for example:
```
def is_name_list_not_list
def is_name_list_a_list
def is_name_list_empty_string
```

students/alheadstrong/lesson_6/test_mailroom_part4.py
```diff
@@ -0,0 +1,91 @@
...
+
+def test_8():
+    d = {'donor_name': 'New Donor', 'donation': 456.0}
+    add_donation_to_new(d)
+    assert ddb['New Donor'] == [456.0]
+
+
+def test_9():
+    d = {'donor_name': 'New Donor', 'donation': 456.0}
+    letter = ('Dear New Donor,'
```
natasha-aleksandrova: another good pattern is to use names like actual/expected
```
def test_create_letter():
    expected = (...)
    assert create_letter(d) == expected
```
makes it really clear to the reader what the expected result is

natasha-aleksandrova:
don't forget to submit on the course website

PR #1456, opened 2019-02-25
Added lesson02 assignment series.py

bromleysj:


bromleysj:
I'm having some difficulty completing the lesson04 assignment for the
mailroom.

My problem is this: I have a bunch of lists that contain the data for the
total(sum) of donations, the average, the number of donations, and the
names.  I essentially need to sort the total donations and then each other
list based on this.
I've used a dict where the name of each donor is the key and the donations
are a list.

Here's the meat of my get_donor_summary function:

totals = [float(format(sum(donors.get(donor)), '.2f')) for donor,
donations in donors.items()]
num_gifts = [len(donors.get(donor)) for donor, donations in donors.items()]
avg_gift = [float(format(sum(donors.get(donor)) /
len(donors.get(donor)), '.2f')) for donor, donations in
donors.items()]
names = list(donors.keys())


Any help with what would be a better way to structure this data would be
greatly appreciated and help me get my week4 and 5 assignments submitted
this week.

Thanks,

Scott

PR #1459, opened 2019-02-26
Lesson 4, Mailroom Pt. 2

ganttArt:


students/christopher_gantt/lesson04/mailroom.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
+
+DONOR_DB = {"Wassily Kandinsky": [43928.13, 131.34, 1928.0],
+            "Jasper Johns": [3134.43, 153.34],
+            "Mark Rothko": [135353.33],
+            "Richard Serra": [153757.87, 28457.12, 1293451.0],
+            "Yves Tanguy": [1534.23, 2542.19]}
```
natasha-aleksandrova: :thumbsup:

not that this should be `donor_db` - all caps is used for constants (things that are immutable)

students/christopher_gantt/lesson04/mailroom.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def main():
+    print("Welcome to The Mailroom")
+    while True:
+        response = input(PROMPT)
+        if response == "4": 
+            print("Thank you for visiting The Mailroom")
+            break
+        if int(response)>4:
```
natasha-aleksandrova: to make this more dynamic -> `if response not in SWITCH_FUNC_DICT`

students/christopher_gantt/lesson04/mailroom.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+SWITCH_FUNC_DICT = {"1": thank_you, "2": report, "3": letter_to_all}
+
+
+def main():
+    print("Welcome to The Mailroom")
+    while True:
+        response = input(PROMPT)
+        if response == "4":
```
natasha-aleksandrova: to make this even cleaner you can map optin "4" to a function that call sys.exit:
```
def quit_program():
    print("Bye!")
    sys.exit()
```

students/christopher_gantt/lesson04/mailroom.py
ganttArt: Thank you! I thought it was used for all global variables.

PR #1462, opened 2019-02-28
Lesson 4, Trigrams Assignment

ganttArt:


students/christopher_gantt/lesson04/trigrams.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
+import random
+
+wish_i_may = "I wish I may I wish I might"
+
+f = open('sherlock.txt')
+book = f.read()
```
natasha-aleksandrova: is it preferred to use `with` when working with files

ganttArt: Because there was a bunch of stuff I had to do to the text of the book I wasn’t sure if I would need to keep the book open until the end of the script. Once I assign f.read to a variable can I close it then?

natasha-aleksandrova: exactly 🙂 

you want to close the file once you're done working with it, in this case you've read the content and you're done with that file. In general, keeping files open for long periods of time is not advised.

PR #1466, opened 2019-03-03
Lesson10 submission for JohnRoge

johnroge:
Pretty sure I'm missing a few opportunities and would appreciate any feedback I can get.


natasha-aleksandrova:
@johnroge can you please elaborate on "Pretty sure I'm missing a few opportunities" ?

johnroge:
The code runs fine, but I suspect I'm not taking full advantage of any opportunities for things like recursion, list comprehension, using the various methods. Would love to see these things pointed out.

students/john_rogers/lesson10/donors_fp.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def total_amount_raised(self):
+        return sum([d.total_donations for d in self.donors])
+
+    def total_number_donations(self):
+        return sum([d.number_of_donations for d in self.donors])
+
+    def thank_all(self):
+        for donor in self.donors:
+            letter = Donor.form_letter(donor)
```
natasha-aleksandrova: incorrect usage of class: `form_letter` should be called on an instance.

`donor` is an instance
`Donor` is a class

so should be : `letter = donor.form_letter()`

students/john_rogers/lesson10/donors_fp.py
```diff
@@ -0,0 +1,141 @@
...
+            print(letter)
+            print()
+
+    def save_report(self):
+        print('Saving database to disk...')
+        for donor in self.donors:
+            letter = Donor.form_letter(donor)
```
natasha-aleksandrova: same here, call the method on an instance

students/john_rogers/lesson10/donors_fp.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        sorted_donors = []
+
+        for donor in self.donors:
+            sorted_donors.append([[donor.full_name],
+                                 [donor.total_donations],
+                                 [donor.number_of_donations],
+                                 [donor.avg_donation_amount]])
+
+        sorted_donors.sort(key=lambda x: x[1])
```
natasha-aleksandrova: `sort()` method has a reverse flag that you can set to True

johnroge:
Appreciate the feedback, will get those fixed today.
JR

PR #1469, opened 2019-03-04
Lesson 7 html_render

mattclau:
I really struggled with this one because I've never gotten OO programming before in any of my classes. I  had to lean heavily on the tutorial at first, but by step 4, I was getting the hang of it and didn't need refer back to it.  I can't figure out why the test file keeps telling me that html_render doesn't take 3 arguments.

### DIFF
--- a/students/mattclau/lesson07/html_render.py
+++ b/students/mattclau/lesson07/html_render.py
@@ -1,0 +1,160 @@
+'Main class to hold html elements'
+class Element:
+
+    'class attribute to store html tags'
+    tag = 'html'
+
+    indent = '\t'
+
+    def __init__(self, content=None, **kwargs):
+        'initialize Element with the content to be stored as a list, default list starts with the content passed in'
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
+        if kwargs is not None:
+            self.attribs = kwargs
+        else: self.attribs = {}
+
+
+    def append(self, new_text):
+        'append method that appends new text to the element class'
+        self.content.append(new_text)
+
+    def add_attributes(self):
+
+        attributes = ''
+
+        for key, value in self.attribs.items():
+            attributes += key
+            attributes += '="'
+            attributes += value +'"'
+
+        return attributes
+
+
+    def render(self, out_file, cur_ind = ''):
+        'renders the content of the element to a file with appropriate tags'
+
+        #add attributes to go inside tag if they exist
+        if self.attribs != {}:
+            out_file.write(f'{cur_ind}<{self.tag} {self.add_attributes()}>\n')
+        else:
+            out_file.write(f'{cur_ind}<{self.tag}>\n')
+
+        for item in self.content:
+            try:
+                item.render(out_file, cur_ind+self.indent)
+            except AttributeError:
+                out_file.write(cur_ind+self.indent+item)
+
+            out_file.write('\n')
+
+        out_file.write(f'{cur_ind}</{self.tag}>')
+
+
+class Body(Element):
+    tag = 'body'
+
+class Html(Element):
+    tag = 'html'
+
+    #override render to add doctype tag at top of file then call element render for rest of file
+    def render(self, out_file):
+        out_file.write(f'<!DOCTYPE html>\n')
+
+        Element.render(self, out_file, '')
+
+class P(Element):
+    tag = 'p'
+
+class Head(Element):
+    tag = 'head'
+
+class OneLineTag(Element):
+    'class to render all on 1 line'
+    def render(self, out_file, indent = ''):
+
+        if self.attribs != {}:
+            out_file.write(f'{indent}<{self.tag} {self.add_attributes()}>')
+        else:
+            out_file.write(f'{indent}<{self.tag}>')
+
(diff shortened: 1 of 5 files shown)


mattclau:
Added lesson 8 assignment circle class.  I got through this one much easier now that OO is finally clicking some.

students/mattclau/lesson08/circle.py
```diff
@@ -0,0 +1,65 @@
...
+
+@total_ordering
+class Circle:
+
+    _diameter = 0
+    pi = math.pi
+    _area = 0
+    circles = []
+
+    def __init__(self, radius, diameter=None):
```
natasha-aleksandrova: why have this optional param that's never used?

students/mattclau/lesson08/circle.py
```diff
@@ -0,0 +1,65 @@
+#Class for creating and manipulating circles
+import math
+from functools import total_ordering
+
+@total_ordering
+class Circle:
+
+    _diameter = 0
```
natasha-aleksandrova: why is this a class a variables?

students/mattclau/lesson08/circle.py
```diff
@@ -0,0 +1,65 @@
...
+    pi = math.pi
+    _area = 0
+    circles = []
+
+    def __init__(self, radius, diameter=None):
+        self.radius = radius
+        self.diameter = radius*2
+        self.circles.append(self)
```
natasha-aleksandrova: it is not ideal design. Circle represents a single circle and shouldn't have a reference to many circles.

students/mattclau/lesson07/html_render.py
```diff
@@ -0,0 +1,160 @@
...
+            self.content = []
+        else:
+            self.content = [content]
+        if kwargs is not None:
+            self.attribs = kwargs
+        else: self.attribs = {}
+
+
+    def append(self, new_text):
+        'append method that appends new text to the element class'
```
natasha-aleksandrova: docstrings use triple double quotes

students/mattclau/lesson07/html_render.py
```diff
@@ -0,0 +1,160 @@
...
+        self.content.append(new_text)
+
+    def add_attributes(self):
+
+        attributes = ''
+
+        for key, value in self.attribs.items():
+            attributes += key
+            attributes += '="'
+            attributes += value +'"'
```
natasha-aleksandrova: this would look better on a single line:
`attributes += f'{key}="{value}"'`

students/mattclau/lesson07/html_render.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+    def render(self, out_file, cur_ind = ''):
+        'renders the content of the element to a file with appropriate tags'
+
+        #add attributes to go inside tag if they exist
+        if self.attribs != {}:
+            out_file.write(f'{cur_ind}<{self.tag} {self.add_attributes()}>\n')
+        else:
+            out_file.write(f'{cur_ind}<{self.tag}>\n')
```
natasha-aleksandrova: shouldn't need the if/else since self.add_attributes would return an empty string on empty attrs

students/mattclau/lesson07/html_render.py
```diff
@@ -0,0 +1,160 @@
...
+class Title(OneLineTag):
+    tag = 'title'
+
+class SelfClosingTag(Element):
+
+    #override init to raise error if content passed
+    def __init__(self, content=None, **kwargs):
+        if content is not None:
+            raise TypeError('Self closing tags cannot have content')
+        if kwargs is not None:
```
natasha-aleksandrova: you can call here the super init method instead of duplicating the code.

students/mattclau/lesson07/html_render.py
```diff
@@ -0,0 +1,160 @@
...
+    #override init to allow various headers
+    def __init__(self, level, content=None, **kwargs):
+        'initialize Element with the content to be stored as a list, default list starts with the content passed in'
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
+        if kwargs is not None:
+            self.attribs = kwargs
+        else: self.attribs = {}
```
natasha-aleksandrova: call the parent init method instead of duplicating the code.

mattclau:
Hi Natasha,

I made changes to the circle class, but I don't understand this comment:


+    def __init__(self, radius, diameter=None):


why have this optional param that's never used?

The diameter=None param is used by the alternate constructor to make a circle using the diameter instead of the radius:

@classmethod
        radius = diameter / 2
        diameter = diameter
        return cls(radius, diameter)

Is there some other way to do this?

Thanks,

Matthew Houston

850-702-8788

students/mattclau/lesson08/circle.py
natasha-aleksandrova: to your question about unused diameter param; you're not actually using it in here, you're setting 
`self.diameter = radius * 2`

the alternate constructor should be:
```
    @classmethod
        return cls(diameter / 2)
```

mattclau:
Thnks, that makes sense.


Matthew Houston

850-702-8788

PR #1472, opened 2019-03-05
Assignment 03 Submission for the review. 

mzkhan87:
Dear instructor, 

Please review my assignment 03. Please let me know if there is anything you'd like me to change to improve my skills. 

Thanks so much.

### DIFF
--- a/students/MzKhan/lesson03/mailroom.py
+++ b/students/MzKhan/lesson03/mailroom.py
@@ -1,0 +1,106 @@
+'''
+    Name: Muhammad Khan
+    Date: 02/28/2019
+    Assignment03
+
+'''
+
+#Goal:
+# Automate the mailroom job of an individual working at a local
+# charity organization. The script will prompt the user for the varioius options
+# and perform the desired tasks.
+
+# Global donors List.
+
+donors_list = [ ['Adam', 'Johnson', 6000.20, 2200, 1050,100],
+              ['Matt', 'Marvin', 550,1000,250],
+              ['Ashley', 'Wiggins', 55.66,270,1000],
+              ['Kristina', 'Laughrey', 500, 500,200],
+              ['Kimberley', 'Allen', 12000.99, 5000],
+              ['Doug', 'Boolinger', 260.57, 930.90],
+              ['Sherry', 'Henning', 6000, 2460.20, 900] ]
+
+def prompt_user():
+    print('Please choose one of the following options: \n')
+    print("1: Send a Thank You")
+    print("2: Create a report")
+    print("3: Quit")
+    return int(input("Option:  "))
+
+def thank_you(donors_data = donors_list):
+    name = name_prompt().split(' ')
+    amount = donation_prompt(name)
+    add_donation(name, amount, donors_data)
+    email(name, amount)
+
+def add_donation(name, amount, donors_data = donors_list):
+    donorIndex = find_donor(name,donors_data)
+    if donorIndex < len(donors_data):
+        donors_data[donorIndex].append(amount)
+    else:
+        donors_data.append([name[0],name[1],amount])
+
+def name_prompt():
+    donor_name = input("Please enter the donor's full name OR \n"
+                 +"type 'list' for the existing donors: ")
+    if donor_name == 'list':
+        display_list()
+        donor_name = input("Please enter the donor's full name: ")
+    return donor_name
+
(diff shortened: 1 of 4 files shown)


students/MzKhan/lesson03/mailroom.py
```diff
@@ -0,0 +1,153 @@
...
+    # This method prompts the user for the valid input.
+    # input: string
+    # return: int
+    print('Please choose one of the following options: \n')
+    print("1: Send a Thank You")
+    print("2: Create a report")
+    print("3: Quit")
+    return int(input("Option:  "))
+
+def thank_you(donors_data = donors_list):
```
natasha-aleksandrova: in general using mutable data structure for default value is a bad practice:
https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

students/MzKhan/lesson03/mailroom.py
```diff
@@ -0,0 +1,153 @@
...
+
+def add_donation(name, amount, donors_data = donors_list):
+    # This method adds the donation amount for the given donor.
+    # If the donor is already present in the history, it appends the donation
+    # amount for that user.
+    # If the donor is a new donor, it appends the new donor to the donor's list
+    # parm: list
+    # parm: float
+    # parm: list    (kwarg)
+    donorIndex = find_donor(name,donors_data)
```
natasha-aleksandrova: use snake_case style for var names -> `donor_index`

students/MzKhan/lesson03/mailroom.py
```diff
@@ -0,0 +1,153 @@
...
+    if donorIndex < len(donors_data):
+        donors_data[donorIndex].append(amount)
+    else:
+        donors_data.append([name[0],name[1],amount])
+
+def name_prompt():
+    # This method prompts the user for the donor's name OR
+    # check the existing donors.
+    # parm: none
+    # return: string
```
natasha-aleksandrova: this should be replaced by `"""Function docstring"""`

students/MzKhan/lesson03/mailroom.py
```diff
@@ -0,0 +1,153 @@
...
+    # This method checks if the given donor exists in the donor's list.
+    # If the donor does exist, it returns its index
+    # If the donor doesn't exist, it returns the index to add the new donor.
+    # parm: list
+    # parm: list    (kwarg)
+    # return: int
+    index = len(donors_data)
+    for donor in donors_data:
+        if (donor_name[0] == donor[0] and donor_name[1] == donor[1]):
+            index = donors_data.index(donor)
```
natasha-aleksandrova: if you're looking for something in the list and found you should exit at this point to avoid continuing looping.
so here you can add `return index`

hint: also remember that for loop supports `else` statement that you could utilize here as well

students/MzKhan/lesson03/mailroom.py
```diff
@@ -0,0 +1,153 @@
...
+    # This method returns the total donation amount at index = 2
+    # parm: list
+    # return: list[2]
+    return a_list[2]
+
+if __name__ == "__main__":
+
+    #Main program flow control and the interaction with the user.
+
+    while (True):
```
natasha-aleksandrova: no need for parens here -> `while True:`

students/MzKhan/lesson03/mailroom.py
```diff
@@ -0,0 +1,153 @@
...
+        user_response = prompt_user()
+        if ( user_response == 3):
+            break
+        elif ( user_response == 1):
```
natasha-aleksandrova: parens here are also redundant -> `if user_response == 3`

students/MzKhan/lesson03/mailroom.py
```diff
@@ -0,0 +1,153 @@
...
+    return a_list[2]
+
+if __name__ == "__main__":
+
+    #Main program flow control and the interaction with the user.
+
+    while (True):
+        user_response = prompt_user()
+        if ( user_response == 3):
+            break
```
natasha-aleksandrova: hint: you can map quit option 3 to a function as well (this will be important in next revision).

your quit function can call `sys.exit()`

mzkhan87:
Hi Natasha,

Thanks for the feedback. I will make changes as you suggested in the
mailroom1.py file.

Muhammad.

PR #1474, opened 2019-03-08
Added homework 7

ethanenguyen:
Please see attached html_render.py and pytest in test_html_render.py

### DIFF
--- a/students/ethan_nguyen/lesson07/html_render.py
+++ b/students/ethan_nguyen/lesson07/html_render.py
@@ -1,0 +1,322 @@
+#!/usr/bin/env python3
+
+
+class TextWrapper:
+    """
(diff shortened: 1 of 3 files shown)


students/ethan_nguyen/lesson07/html_render.py
```diff
@@ -0,0 +1,322 @@
...
+        out_file.write("{}<{}>\n".format(cur_ind,"!DOCTYPE html"))
+        super().render(out_file, cur_ind)
+
+class Body(Element):
+    tag = 'body'
+    def __init__(self, content=None):
+        if content is None:
+            self.contents = []
+        else:
+            self.contents = [content]
```
natasha-aleksandrova: I don't think you need this here since Element constructor does the same thing.

PR #1477, opened 2019-03-10
lesson7 assignment

nubber7:


students/g_rama/lesson07/html_render.py
```diff
@@ -0,0 +1,117 @@
...
+        return self.content.append(p_content)
+
+    def render(self, file_out, cur_ind=" "):
+        file_out.write(f"<{self.tag}")
+        for k in self.html_options:
+            file_out.write(f" {k}={self.html_options[k]}")
+        file_out.write(f"> ")
+        file_out.write("\n")
+
+        for _ in range(len(self.content)):
```
natasha-aleksandrova: have you read my previous feedback? `_` is not something you should use for *variable* names.

your loop is also incorrect here, instead should be:
```
for item in self.content:
    item.render(file_out)
```

PR #1478, opened 2019-03-11
Assignment04

mzkhan87:
mailroom2.py
1) I create a folder "Letters" in the lesson04 directory to place all the letters to the donors.
2) I made improvements based on the previous feedback in regard to better coding practices. 

trigram.py
I'm a bit confused on the trigram algorithm. Please check my algorithm. 
I write out the text in multiple lines rather than one long line in the text format. 
I was not able to find if python has a method that writes line by line.

### DIFF
--- a/students/MzKhan/lesson04/mailroom2.py
+++ b/students/MzKhan/lesson04/mailroom2.py
@@ -165,7 +165,8 @@ def quit():
     sys.exit()
 
 def send_letter_everyone(donors_data):
-    ''' This method generates the thank you letters for each donor in the list
+    '''
+    This method generates the thank you letters for each donor in the list
     parm: list
     '''
     date_format = '{:%m-%d-%Y}'.format(datetime.datetime.now())
--- a/students/MzKhan/lesson04/mailroom2.py
+++ b/students/MzKhan/lesson04/mailroom2.py
@@ -1,0 +1,215 @@
+'''
+    Name: Muhammad Khan
+    Date: 03/08/2019
+    Assignment04
+
+'''
+import sys
+import os
+import datetime
+from operator import itemgetter as igetter
+
+#Goal:
+# Automate the mailroom job of an individual working at a local
+# charity organization. The script will prompt the user for the varioius options
+# and perform the desired tasks.
+
+# Global donors List.
+# This list is accessible everywhere in the script.
+
+donors_list = [ ['Adam', 'Johnson', 6000.20, 2200, 1050,100],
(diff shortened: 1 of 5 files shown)


students/MzKhan/lesson04/mailroom2.py
```diff
@@ -0,0 +1,217 @@
...
+    #Main program flow control and the interaction with the user.
+    options = {1:thank_you,2:print_report,3:send_letter_everyone,4:quit}
+    while True:
+        user_response = prompt_user()
+        if user_response == 4:
+            options[user_response]()
+        elif user_response in options:
+            options[user_response](donors_list)
+        else:
+            print("Invalid Response Please Try Again :)")
```
natasha-aleksandrova: you don't need to pass in `donors_list` since its a global variable (module level)

you can simplify this to:
```
while True:
    if user_response in options:
        options[user_response]()
    else:
        print('invalid')
```

students/MzKhan/lesson04/trigram.py
```diff
@@ -0,0 +1,119 @@
...
+
+def read_text(filename):
+    '''
+    The method reads the data from the text file.
+    parm: text file
+    return: text
+    '''
+    with open(filename, 'r') as f:
+        text = f.read()
+        f.close()
```
natasha-aleksandrova: no need for close() call

students/MzKhan/lesson04/trigram.py
```diff
@@ -0,0 +1,119 @@
...
+
+import random
+import re
+
+def read_text(filename):
+    '''
+    The method reads the data from the text file.
+    parm: text file
+    return: text
+    '''
```
natasha-aleksandrova: docstrings use double triple quotes
```
def read_text(filename):
    """The method docstring
    """
    pass
```

students/MzKhan/lesson04/trigram.py
```diff
@@ -0,0 +1,119 @@
...
+    trigram_text = generate_new_text(trigram_dict,number_of_words)
+    write_text(filename,trigram_text)
+    print("Please read the trigram text generated!")
+
+
+if __name__=="__main__":
+
+    filename = "sherlock.txt"
+    #filename = "sherlock_small.txt"
+    trigram_story(filename,1000)
```
natasha-aleksandrova: great work on this assignment!

mzkhan87:
Thanks so much Natasha for your feedback. I resubmitted the assignment04.

Have a great day.

PR #1480, opened 2019-03-12
Solution_Assignment_Lesson09_FlorentinPopescu

FlorentinPopescu:
Thank you for reviewing my solution for the assignment in lesson 8.
Please review my solution for the assignment in lesson 9.
- 18_Object_Oriented_Mailroom_L9_FP.py
- 19_UnitTesting_of_ObjectOrientedMailroom_L9_FP.py
Thank you.

### DIFF
--- a/students/FlorentinPopescu/18_Object_Oriented_Mailroom_L9_FP.py
+++ b/students/FlorentinPopescu/18_Object_Oriented_Mailroom_L9_FP.py
@@ -1,0 +1,260 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Sat Mar 2 14:31:07 2019
+@author: Florentin Popescu
+"""
+
+#===================LESSON_09====================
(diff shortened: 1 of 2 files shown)


students/FlorentinPopescu/18_Object_Oriented_Mailroom_L9_FP.py
```diff
@@ -0,0 +1,260 @@
...
+    def check_name(name):
+        #Test if donor's name contains special characters
+        try:
+            regex = re.compile("[-+=@_!#$%^&*()<>?/\|}{~:]")
+            return bool(regex.search(name))
+        except re.error as r_err:
+            print(f"Error <{r_err}>. Invalid regular expression.")
+   
+    @staticmethod
+    def letter(name, amount):
```
natasha-aleksandrova: this should be called on an instance instead of as staticmethod

PR #1481, opened 2019-03-12
Added files for Lesson 08

mattcasali:
Added files for Lesson 08. Please review. Thank you!

### DIFF
--- a/students/matt_casali/Lesson08/circle.py
+++ b/students/matt_casali/Lesson08/circle.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+
+import math
+
+
+class Circle:
+    # Step 1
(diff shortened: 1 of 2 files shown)


students/matt_casali/Lesson08/circle.py
```diff
@@ -0,0 +1,65 @@
...
+    # Step 4
+    @property
+    def area(self):
+        area = (self._radius ** 2) * math.pi
+        return f"{area:.2f}"
+
+    # Step 5
+    @classmethod
+    def from_diameter(self, diameter):
+        return Circle(diameter / 2)
```
natasha-aleksandrova: with class methods you should use `cls` instead of `self`

then return `cls(diameter / 2)`

PR #1482, opened 2019-03-12
Lesson 6, Unit Testing

ganttArt:


students/christopher_gantt/lesson06/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+            func = switch_dict[response]()
+        except KeyError:
+            print("Not a valid option")
+        else:
+            if func == "exit menu":
+                break
+
+def exit_menu():
+    return "exit menu"
+
```
charlesrobison: Inconsistent spacing between functions - go with 2 blank rows between each function.

students/christopher_gantt/lesson06/mailroom.py
```diff
@@ -0,0 +1,123 @@
...
+
+def print_donors():
+    print("\n".join([key for key in donor_db]))
```
charlesrobison: Nice

PR #1485, opened 2019-03-15
Please see  homework 8

ethanenguyen:
Please see  homework 8


students/ethan_nguyen/lesson08/circle.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @property
+    def area(self):
+        return round(math.pi * (self.radius * self.radius), 4)
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        self = cls(int(diameter/2))
+        self.diameter = int(diameter)
+        return self
```
natasha-aleksandrova: this method should be `return cls(diameter / 2)` - you don't need to manually set the diameter attr

students/ethan_nguyen/lesson08/circle.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def __gt__(self, other):
+        return (self.area > other.area)
+
+    @property
+    def diameter(self):
+        return self.radius * 2
+    @diameter.setter
+    def diameter(self, value):
+        self._diameter = value
```
natasha-aleksandrova: why set `self._diameter` if you're not using it in the diameter getter?

PR #1491, opened 2019-03-18
Assignment05 Submission

mzkhan87:
Dear Reviewer, 

I've done the following. 
mailroom3.py - 
- Used the dict comprehension to create the dictionary.
- Used list comprehension
- Added exceptions - raised exception in the name_prompt() method. I catch that exception in the main method. 
- The code works. 

exception_exercise.py

added the exceptions and I got the good results. 

Please let me know if there is anything I should improve. 

Thanks so much for your feedback :)

### DIFF
--- a/students/MzKhan/lesson05/mailroom3.py
+++ b/students/MzKhan/lesson05/mailroom3.py
@@ -1,0 +1,177 @@
+'''
+Name: Muhammad Khan
+Date: 03/15/2019
+Assignment05
+'''
+
+import sys
+import os
+import datetime
+from operator import itemgetter as igetter
+
+#Goal:
+# Automate the mailroom job of an individual working at a local
+# charity organization. The script will prompt the user for the varioius options
+# and perform the desired tasks.
+
+# Global donors data.
+# Accessible everywhere in the script.
+donors_name = ["Adam Johnson","Matt Marvin","Ashley Wiggins",
+                "Kristina Laughrey","Kimberley Allen", "Doug Boolinger",
+                "Sherry Henning"]
+donations = [ [6000.20, 2200, 1050,100],
+                [550,1000,250],
+                [55.66,270,1000],
+                [500, 500,200],
+                [12000.99, 5000],
+                [260.57, 930.90],
+                [6000, 2460.20, 900] ]
+
+# Create the dictionary using the dict comprehension.
+donors_data = dict(zip(donors_name,donations))
+
+def prompt_user():
+    """
+    prompt the user for the valid input.
+    input: string
+    return: int
+    """
+    print('Please choose one of the following options: \n')
+    print("1: Send a Thank You")
+    print("2: Create a report")
+    print("3: Send letters to everyone")
+    print("4: Quit")
+    return int(input("Option:  "))
+
+def thank_you():
+    """send a thank you note to a donor for his/her donation."""
+    name = name_prompt()
+    amount = donation_prompt(name)
+    add_donation(name, amount)
+    email(name, amount)
+
+def add_donation(name, amount):
+    """
+    parm: string
+    parm: float
+    """
+    donors_data.setdefault(name,[]).append(amount)
+
+def name_prompt():
+    """return a name string"""
+    donor_name = input("Please enter the donor's full name OR \n"
+                 +"type 'list' for the existing donors: ")
+
+    if not "".join(donor_name.split()).isalpha():
+        raise Exception("The name should be all Alphabetic characters: ")
+    if donor_name == 'list':
+        display_list()
+        donor_name = input("Please enter the donor's full name: ")
+        if not "".join(donor_name.split()).isalpha():
+            raise Exception("The name should be all Alphabetic characters: ")
+
+    return donor_name.title()
+
+def donation_prompt(name):
+    """
+    prompt the user for the donation amount for the given donor.
+    parm: string
+    return: float
+    """
+    donation_amount = input("Please enter the donation amount for {}: "
+                                                               .format(name))
+    return float(donation_amount)
+
+
+def display_list():
+    """print the donor's list to the user."""
+    print("\nOur generous donors: \n")
+    for donor in donors_data:
+        print(donor)
+
+def email(donor_name, amount):
+    """
+    generate the email for the given donor for his/her donation.
+    parm: string
+    parm: float
+    """
+    header = "\nDear {}".format(donor_name)
+    body = "\nThank you so much for your generous donation of $ {:.2f}."
+    print(header, body.format(amount))
+
+def print_report():
+    """Print the donors' report in the formatted tabular form."""
+    title = "{:24} | {:12} | {:10} | {:20}"
+    dashes=67*('-');print(dashes)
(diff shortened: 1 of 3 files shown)


students/MzKhan/lesson05/mailroom3.py
```diff
@@ -0,0 +1,177 @@
...
+
+    if not "".join(donor_name.split()).isalpha():
+        raise Exception("The name should be all Alphabetic characters: ")
+    if donor_name == 'list':
+        display_list()
+        donor_name = input("Please enter the donor's full name: ")
+        if not "".join(donor_name.split()).isalpha():
+            raise Exception("The name should be all Alphabetic characters: ")
+
+    return donor_name.title()
```
natasha-aleksandrova: note that above would be nice to have as a while loop: continue asking for name or "list" command and stop when name is entered.

students/MzKhan/lesson05/mailroom3.py
```diff
@@ -0,0 +1,177 @@
...
+
+def donation_prompt(name):
+    """
+    prompt the user for the donation amount for the given donor.
+    parm: string
+    return: float
+    """
+    donation_amount = input("Please enter the donation amount for {}: "
+                                                               .format(name))
+    return float(donation_amount)
```
natasha-aleksandrova: what if they enter as a string? will your program crash?

students/MzKhan/lesson05/mailroom3.py
```diff
@@ -0,0 +1,177 @@
...
+    title = "{:24} | {:12} | {:10} | {:20}"
+    dashes=67*('-');print(dashes)
+    print(title.format('Donor Name','Total Given','Num Gifts','Average Gift'))
+    strf_format = "{:24} ${:12.2f}   {:^10d}  ${:12.2f}"
+    print(dashes)
+    report_list = calculate_total_gift()
+    for donor in sorted(report_list, key = igetter(1), reverse = True):
+        print(strf_format.format(*donor))
+    print(dashes)
+def calculate_total_gift():
```
natasha-aleksandrova: two lines space between module functions.

students/MzKhan/lesson05/mailroom3.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    create a list from the global donors' data.
+    return: list
+    """
+    new_list = [ [donor, sum(donors_data[donor][0:]),len(donors_data[donor]),
+               sum(donors_data[donor][0:])/len(donors_data[donor])]
+               for donor in donors_data ]
+    return new_list
```
natasha-aleksandrova: 👍 for comprehension use

students/MzKhan/lesson05/mailroom3.py
```diff
@@ -0,0 +1,177 @@
...
+
+def quit():
+    """exit out from the user interaction."""
+    print("Thank you. Have a nice day:")
+    sys.exit()
+
+def send_letter_everyone():
+    """generate the thank you letters for each donor in the list"""
+    date_format = '{:%m-%d-%Y}'.format(datetime.datetime.now())
+    for donor in donors_data:
```
natasha-aleksandrova: more pythonic to loop over both keys and values if you need both: 'for name, donations in donors_data.items()`

students/MzKhan/lesson05/mailroom3.py
```diff
@@ -0,0 +1,177 @@
...
+
+    #Main program flow control and the interaction with the user.
+    options = {1:thank_you,2:print_report,3:send_letter_everyone,4:quit}
+    while True:
+        try:
+            user_response = prompt_user()
+            options[user_response]()
+        except KeyError as e:
+            print("Invalid Option Selected. Please try again :) ")
+        except ValueError as e:
```
natasha-aleksandrova: you should group KeyError and Value error together - as both are considered invalid options: 'except (KeyError, ValueError)`

students/MzKhan/lesson05/mailroom3.py
```diff
@@ -0,0 +1,177 @@
...
+            print("VALUE ERROR:",e)
+        except TypeError as e:
+            print("TYPE ERROR:",e)
```
natasha-aleksandrova: what's the reason for TypeError? What should the user see in this case? (think user friendly error messages)

students/MzKhan/lesson05/mailroom3.py
```diff
@@ -0,0 +1,177 @@
...
+        except Exception as ex:
+            print("ERROR: ",ex)
```
natasha-aleksandrova: you can say something like: "System crashed for unknown reason. System Error: %s"

natasha-aleksandrova:
@mzkhan87 please include a Pull Request link in your submissions on the course website. You can copy the link from your browser and paste it into the submission form. 

Your link for this PR: https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/1491

students/MzKhan/lesson05/mailroom3.py
mzkhan87: Nice catch Natasha :)

I didn't use the while loop here because I wanted to keep my functions very simple. I was thinking of implementing all the business logic in the main function. My intention is to catch exceptions, user prompts options all to be handled at one place. 

What are you thoughts in terms of best practices?

Should I catch exceptions locally in the functions or Should I catch exceptions only at one place where all of these functions are implemented?

Should I use while loops in the functions where I'm expecting valid user input Or otherwise?

Thanks.

students/MzKhan/lesson05/mailroom3.py
mzkhan87: This exception will be finally "catch" in the main function. I asked the same question in the last comments. Should I catch exception in this function? Or just catch in the main function?

students/MzKhan/lesson05/mailroom3.py
mzkhan87: I'm sorry I didn't understand what you mean by Two lines space between module functions?

students/MzKhan/lesson05/mailroom3.py
mzkhan87: Awesome :)

students/MzKhan/lesson05/mailroom3.py
mzkhan87: Will do.

students/MzKhan/lesson05/mailroom3.py
natasha-aleksandrova: two lines space:
```
def function1():
    pass
# line 1 space
# line 2 space
def function2():
    pass
```
will look like:
```
def function1():
    pass


def function2():
    pass
```

students/MzKhan/lesson05/mailroom3.py
natasha-aleksandrova: The answer may vary depending on the program and kind of code it is.

In the above case you want catch the exception in here, for reasons:
* exceptions should be around the code that throws it
** you should be clear and understand what code may throw what error
* you may want to let user know with explicit message: "Please enter a number!" - right now you don't get the granularity
* you can correct user mistake instead of just letting your program falling through and the user not knowing what they did wrong.

PR #1492, opened 2019-03-18
Adding files for Lesson 09

mattcasali:
Added file for Lesson 09. Please review. Thanks!

### DIFF
--- a/students/matt_casali/Lesson09/mailroom_9.py
+++ b/students/matt_casali/Lesson09/mailroom_9.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+
+import datetime, sys
+now = datetime.datetime.now()
+
+class Donor:
+    def __init__(self, first_name, last_name, donation_amount):
+        self.first = first_name
+        self.last = last_name
+        self.donation = donation_amount
+
+    @property
+    def full_name(self):
+        return f"{self.first} {self.last}"
+
+    def add_donation(self, amount):
+        return self.donation.append(amount)
+
+    def sum_donation(self):
+        return sum(self.donation)
+
+
+def get_name():
+    return input("Who would you like to thank? If you would like a list of donors, enter 'list'. ")
+
+
+def get_amount():
+    try:
+        return float(input(f"How much did this person donate? "))
+    except ValueError:
+        print("You did not enter a numeric value. Please try again.")
+        sys.exit()
+
+
+class Donor_Actions:
+    def __init__(self, donors):
+        if donors is None:
+            self.donors = []
+        else:
+            self.donors = donors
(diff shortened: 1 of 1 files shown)


students/matt_casali/Lesson09/mailroom_9.py
```diff
@@ -0,0 +1,117 @@
...
+
+def get_amount():
+    try:
+        return float(input(f"How much did this person donate? "))
+    except ValueError:
+        print("You did not enter a numeric value. Please try again.")
+        sys.exit()
+
+
+class Donor_Actions:
```
natasha-aleksandrova: per instructions your data classes should not contain user interactions (calls to `input()`)

your second data class should deal with a collection of Donor objects.

students/matt_casali/Lesson09/mailroom_9.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @property
+    def full_name(self):
+        return f"{self.first} {self.last}"
+
+    def add_donation(self, amount):
+        return self.donation.append(amount)
+
+    def sum_donation(self):
+        return sum(self.donation)
```
natasha-aleksandrova: you can also include methods for average, number of gifts and letter text.

students/matt_casali/Lesson09/mailroom_9.py
mattcasali: Hey Natasha, I made sure to keep the functions involving and calls to input() outside of the two classes I created. Those two functions are called on within the Donor_Actions class. Is there a better way I should approach this problem? Thanks!

natasha-aleksandrova: Even though `input()` isn't called directly - it calls a function that wraps the input function, which ends up being the same.


Your two data classes should be purely dealing with data then your program flow code will interact with these objects.

mattcasali: Okay, I think that makes sense. I've added the calls for input() to the main() function so the calls stay outside of the Donor_Actions class. 

I also went back and added more methods for average donation, the number of donations, the text within the thank you letter, and a few others. 

Would it be possible for to resubmit the exercise?

natasha-aleksandrova: Absolutely - resubmit via a new PR and I will review/adjust the grade accordingly.

mattcasali: Hey Natasha, just submitted another PR with the new code.

https://github.com/UWPCE-PythonCert-ClassRepos/Self_Paced-Online/pull/1493

Thanks for the help with this!

PR #1493, opened 2019-03-20
Updating lesson09 mailroom script

mattcasali:
Submitted a new version of the mailroom script. Please review. Thanks for the help!

### DIFF
--- a/students/matt_casali/Lesson09/mailroom_9_V2.py
+++ b/students/matt_casali/Lesson09/mailroom_9_V2.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python3
+
+import datetime, sys
+now = datetime.datetime.now()
+
+class Donor:
+    def __init__(self, first_name, last_name, donation_amount):
+        self.first = first_name
+        self.last = last_name
+        self.donation = donation_amount
+
+    @property
+    def full_name(self):
+        return f"{self.first} {self.last}"
+
+    def add_donation(self, amount):
+        return self.donation.append(amount)
+
+    def sum_donation(self):
+        return sum(self.donation)
+
+    def average_donation(self):
+        return sum(self.donation)/len(self.donation)
+
+    def number_donations(self):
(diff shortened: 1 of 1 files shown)


students/matt_casali/Lesson09/mailroom_9_V2.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def add_donor(self, donor):
+        self.donors.append(donor)
+
+    def all_donor_names(self):
+        return [donor.full_name for donor in self.donors]
+
+    @staticmethod
+    def letter_text(name, amount):
+        return f"Dear {name},\nThank you very much for your donation of ${amount:,.2f}.\nSincerely,\nMatt Casali"
```
natasha-aleksandrova: why is this method static? the method deals with donor attributes - specifically single donor attributes.. this probably means they should live in the `Donor` class 🙂

students/matt_casali/Lesson09/mailroom_9_V2.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @staticmethod
+    def create_name(name):
+        return name + f" {now.year}{now.month:0>2d}{now.day:0>2d}" + ".txt"
+
+    def write_letter(self, file_name, name, amount):
+        with open(file_name, "w") as f:
+            f.write(self.letter_text(name, amount))
```
natasha-aleksandrova: should be passing a donor object here

students/matt_casali/Lesson09/mailroom_9_V2.py
```diff
@@ -0,0 +1,140 @@
...
+        print("A thank you message has been created.")
+
+    def send_thank_you(self, name, amount):
```
natasha-aleksandrova: remember that this class is for Data handling, not program flow logic. Send thank you is an action from the main program flow.

Instead you should have methods like `add_donation`

PR #1502, opened 2019-03-26
Lesson 8

alheadstrong:
Lesson 8 circle class and tests.

### DIFF
--- a/students/alheadstrong/lesson_8/circle.py
+++ b/students/alheadstrong/lesson_8/circle.py
@@ -1,0 +1,65 @@
+from math import pi
+
+
+class Circle(object):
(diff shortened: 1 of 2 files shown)


students/alheadstrong/lesson_8/circle.py
```diff
@@ -0,0 +1,65 @@
...
+        self.radius = value/2
+
+    @property
+    def area(self):
+        return pi*self.radius**2
+
+    @classmethod
+    def from_diameter(cls, value):
+        self = cls(value/2)
+        return self
```
natasha-aleksandrova: no need to store in a var: `return cls(value / 2)`

PR #1503, opened 2019-03-26
Lesson 7 HTML Render

ganttArt:


students/christopher_gantt/lesson07/html_render.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def __init__(self, content=None, **kwargs):
+        if content is None:
+            self.contents = []
+        else:
+            self.contents = [content]
+        self.attributes = kwargs
+
+
+    def append(self, new_content):
```
natasha-aleksandrova: class methods should only have one line space (module functions have two lines space)

students/christopher_gantt/lesson07/html_render.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+class Element(object):
```
natasha-aleksandrova: no need to extend from object in python3

students/christopher_gantt/lesson07/html_render.py
```diff
@@ -0,0 +1,123 @@
...
+    def _open_tag(self):
+        open_tag = ["<{}".format(self.tag)]
+        for key, value in self.attributes.items():
+            open_tag.append(' {}="{}"'.format(key, value))
+        return "".join(open_tag)
+
+
+    def render(self, out_file, cur_ind=""):
+        if self.tag == "html":
+            out_file.write('<!DOCTYPE html>\n')
```
natasha-aleksandrova: it's best to keep this as part of inheritance structure - `Html` class would have its own `render` method and then call parent's `render()` to do execute the rest of the code.

students/christopher_gantt/lesson07/html_render.py
ganttArt: To clarify, this means that I can write the class as...   class Element():    ...and not add the object?

natasha-aleksandrova: `class Element:` is python3 style

PR #1505, opened 2019-03-29
Reworked to make Donor class more object oriented

nubber7:
Reworked lesson 9 to make Donor class more of an Object

### DIFF
--- a/students/g_rama/lesson09/Donor.py
+++ b/students/g_rama/lesson09/Donor.py
@@ -4,26 +4,29 @@ from CLI import CLI
 
 class Donor(DonorCollection):
 
-    def add_data_print_thanks(self, amount, fullname):
-        if self.amount_validate(amount):
-            super().donors_collection_data.update({fullname: [amount]})
-            self.thank_you_letter(fullname, amount)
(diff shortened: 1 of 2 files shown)


students/g_rama/lesson09/Donor.py
```diff
@@ -4,26 +4,29 @@
 
 class Donor(DonorCollection):
```
natasha-aleksandrova: `Donor` should not inherit from `DonorCollection`

students/g_rama/lesson09/Donor.py
```diff
@@ -4,26 +4,29 @@
...
 
+    def thank_you_letter(self):
+        """Send a thank you letter"""
+        print(f"Thank you {self.fullname} for donating {self.amount} dollars generously.")
+        return f"Thank you {self.fullname} for donating {self.amount} dollars generously."
+
+    def add_data_print_thanks(self):
+        if self.amount_validate():
+            super().donors_collection_data.update({self.fullname: [self.amount]})
+            self.thank_you_letter()
```
natasha-aleksandrova: single Donor object should not be updating collection

PR #1509, opened 2019-03-31
Assignment 07

mzkhan87:
Dear instructor(s), 

Please check my work for Assignment 07. Please let me know if there is anything I can do to improve the code in the html_render.py file. 

Thanks.

### DIFF
--- a/students/MzKhan/lesson07/html_render.py
+++ b/students/MzKhan/lesson07/html_render.py
@@ -1,0 +1,194 @@
+"""
+Name: Muhammad Khan
+Date: 03/25/2019
+Assignment07
+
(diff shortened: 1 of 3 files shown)


students/MzKhan/lesson07/html_render.py
```diff
@@ -0,0 +1,196 @@
...
+Date: 03/25/2019
+Assignment07
+
+Purpose: A class-based system for rendering html.
+"""
+#######
+#Step 1
+#######
+
+class Element(object):
```
natasha-aleksandrova: note that in python3 you don't need to extend from `object` -> `class Element:`

PR #1511, opened 2019-04-03
Lesson9

alheadstrong:
Please ignore duplicate file. GitHub was not showing the added file or commits, so I tried again...and now there are two.

### DIFF
--- a/students/alheadstrong/lesson_9/mailroom.py
+++ b/students/alheadstrong/lesson_9/mailroom.py
@@ -1,0 +1,206 @@
+import os
+import sys
+
+
+class Donor(object):
+    def __init__(self, name, donation_list):
+        self.name = name
+        self.donations = donation_list
+        self._donation_total = sum(donation_list)
+        self._donation_average = self._donation_total / len(self.donations)
+
+    @property
+    def donation_total(self):
+        self._donation_total = sum(self.donations)
+        return self._donation_total
+
+    @property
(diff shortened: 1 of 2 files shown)


students/alheadstrong/lesson_9/mailroom.py
```diff
@@ -0,0 +1,204 @@
...
+    @property
+    def donation_total(self):
+        self._donation_total = sum(self.donations)
+        return self._donation_total
+
+    @property
+    def donation_average(self):
+        self._donation_total = self.donation_total
+        self._donation_average = self._donation_total/len(self.donations)
+        return self._donation_average
```
natasha-aleksandrova: you don't need the extra hidden variables, can be simply:
```
@property
    def donation_total(self):
    return sum(self.donations)
```
same for the average and number of donations

students/alheadstrong/lesson_9/mailroom.py
```diff
@@ -0,0 +1,204 @@
...
+                return True
+        else:
+            return False
+
+    def new_thank_you(self):
+        """Add donation amount to new or existing donor and generate a thank you text."""
+        while True:
+            donor = input("\nEnter donor name, or type 'list' for current donor list:")
+            if not self.is_name_list(donor):
+                break
```
natasha-aleksandrova: you data classes should not have any user interaction code (no `input` calls)

PR #1513, opened 2019-04-04
FizzBuzz

ramr24:
Finished FizzBuzz assignment.

### DIFF
--- a/students/Ramkumar_Rajanbabu/Lesson03/List_Lab.py
+++ b/students/Ramkumar_Rajanbabu/Lesson03/List_Lab.py
@@ -1,0 +1,126 @@
+"""
+Goal: Learn the basic ins and outs of Python lists.
+
+Series 1: DONE
+DONE-Create a list that contains “Apples”, “Pears”, “Oranges” and “Peaches”.
+DONE-Display the list (plain old print() is fine…).
+DONE-Ask the user for another fruit and add it to the end of the list.
+DONE-Display the list.
+DONE-Ask the user for a number and display the number back to the user and the 
+fruit corresponding to that number (on a 1-is-first basis). 
+Remember that Python uses zero-based indexing, so you will need to correct.
+DONE-Add another fruit to the beginning of the list using “+” and display the 
+list.
+DONE-Add another fruit to the beginning of the list using insert() and display 
+the list.
+DONE-Display all the fruits that begin with “P”, using a for loop.
+
+Series 2: DONE
+Using the list created in series 1 above:
+
+DONE-Display the list.
+DONE-Remove the last fruit from the list.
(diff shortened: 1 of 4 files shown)


students/Ramkumar_Rajanbabu/Lesson03/List_Lab.py
```diff
@@ -0,0 +1,126 @@
...
+    
+list2 = ["Pineapple"] #Creating a new list to add item to existing list
+list1 = list2 + list1 #Adding item to the beginning of the list
+print(list1) #Display list
+
+list1.insert(0, "Watermelon") #Using insert to add item to beginning of list
+list1 #Display list
+
+for fruit in list1: #For every fruit in the list
+    if fruit.startswith("P"): #If statement looks for fruits that start with P
```
charlesrobison: You can use lower() here to catch case sensitive input

students/Ramkumar_Rajanbabu/Lesson03/List_Lab.py
```diff
@@ -0,0 +1,126 @@
...
+        
+#Series 2
+list1 = ["Apples", "Pears", "Oranges", "Peaches"] #Create a list
+print(list1) #Display list
+
+list1.pop() #removes last item in list
+print(list1)
+
+answer = str(input("Enter the name of the fruit to delete: "))
+if answer == "Apples":
```
charlesrobison: lower() would be good here too

students/Ramkumar_Rajanbabu/Lesson03/List_Lab.py
```diff
@@ -0,0 +1,126 @@
...
+    list1.remove("Apples")
+elif answer == "Pears":
+    list1.remove("Pears")
+elif answer == "Oranges":
+    list1.remove("Oranges")
+elif answer == "Peaches":
+    list1.remove("Peaches")
+else:
+    print("Please choose a fruit to delete that is within the list.")
+print(list1)
```
charlesrobison: Recommend putting this in a loop so the user can answer if they receive the else response.

students/Ramkumar_Rajanbabu/Lesson03/List_Lab.py
```diff
@@ -0,0 +1,126 @@
...
+    if answer == "no":
+        list1.remove("Oranges")
+        print(list1)
+
+    answer = str(input("Do you like peaches? "))
+    if answer == "no":
+        list1.remove("Peaches")
+        print(list1)
+    else:
+        print("Please answer between yes or no.")
```
charlesrobison: I think a for loop might be better here - once cycling through the list, it's asking again if I like items in the list I said no to the first time through.

PR #1514, opened 2019-04-04
Assignment08

mzkhan87:
Dear Instructor(s), 

I think I am very confident about the concepts covered in assignment 8. I do have one question. 

To make 2*c1 (where c1=Circle(10)) to work, I used the magic method __rmul__(self, num). When this magic function is not available, the multiplication doesn't work for 2*c1. 

I implemented augmented addition and multiplication for the optional part of the assignment. I noticed during testing that my code works fine for the augmented assignments whether I implement the augmented addition, multiplication or not. 

Please provide me some description of implementing the augmented assignments. 

Thanks for your help.

### DIFF
--- a/students/MzKhan/lesson08/circle.py
+++ b/students/MzKhan/lesson08/circle.py
@@ -1,0 +1,123 @@
+"""
+Name: Muhammad Khan
+Date: 04/01/2019
+Assignment08
+
+"""
+import math as m
+
+class Circle:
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
@mzkhan87 the augmented assignment will use `__add__` if `__iadd__` is not defined - which is why you saw it still work without it being defined.

PR #1516, opened 2019-04-07
Lesson3 revisions

GreggHeller1:
I have tried to address most of your comments in my mailroom code - 
- Tried to use tuples and lists instead of dicts
- Eliminated global flag (that wasn't being used and was also needlessly a list)
- changed functions to string constants
- added sorting feature to report

But I do have a few questions - 
1. Some of your suggestions were focused on functions that returned values that were never used - some of these I had included because of suggestions here, as linked from the course material. (https://jeffknupp.com/blog/2018/10/11/write-better-python-functions/) - He suggests that all functions should return something even if its just for testability purposes. Have I gone too far and these functions are just too trivial to return a plain True, or whatever upstream output I ended up passing back down? Or is the fact that I have a bunch of functions that don't need to return anything a sign that I need to refactor my code in some way? I've left most of them in for now since its an easy thing to take out the next round. 

2. I tried to refactor my code so it quits back to the main menu differently - I didn't like that I had to remember to check if I needed to go back to the main menu after every prompt...so I tried to package this into a prompt handler that handled all communication back and forth between the user. I think it was a good exercise that I got it to work, but it turned out being more messy, and much less readable than the first version - and I still had to be careful that no code would try to run after after being prompted with "q." So I'm planning to change this back more like the way I had it before. But with a more complicated program, both methods seem kinda messy. I'm curious, are there situations where this kind of centralized handler would be the way to go? Is there a more elegant way to return through all the nested functions when prompted?

### DIFF
--- a/students/Gregg/lesson3/mailroom.py
+++ b/students/Gregg/lesson3/mailroom.py
@@ -1,146 +1,147 @@
 #!/usr/bin/env python3
 """Use current knowledge to set up an interactive program"""
 
-donations = {
-    "William Gates, III": [3, 5, 7],
-    "Mark Zuckerberg": [4.50, 8, 2],
-    "Jeff Bezos": [7.77],
-    "Paul Allen": [3.6, 4.5],
-    "bob": [.01]
-}
-return_to_primary_prompt = [False]
+#Todo - restructure so there is only one message handler, with dict mapping prompt, input options, functions and next states
+#Todo - input a dictionary to the letter writing function
+# this is a pain because I need to make the dictionary just to take it apart...it
+#Todo add option to save all leters
+#need additional save function
 
-# This doesn't seem like the cleanest way to do this, but I coultn't
-# come up with anything better without turning it into a full blow stat machine
-# that has a centralized handler thingy for every possible transition
-# That can then return to the start state if necessary
-# It seems like it works ok for this limited example
+
+donors = [
+    "William Gates, III",
+    "Mark Zuckerberg",
+    "Jeff Bezos",
+    "Paul Allen",
+    "bob"
+]
+
+donations = [
+    [3, 5, 7],
+    [4.50, 8, 2],
+    [7.77],
+    [3.6, 4.5],
+    [40]
+]
+
+
+def recurring_prompt(quit_string, prompt, default_action, input_tup=(), action_tup=(), pass_text=False, pass_args = None):
+    """Continues prompting with prompt until the quit_string is returned
+
+    Prompt handler is set up to return values based on the functions in
+    action_list, so one of these functions should return the quit_string
+    otherwise there is no way to quit
+    """
+    while(True):
+        result = prompt_handler(prompt, default_action, input_tup, action_tup, pass_text, pass_args)
+        if result == quit_string:
+            break
+
+
+def prompt_handler(prompt, default_action, input_tup=(), action_tup=(), pass_text=False, pass_args = None):
+    """Prompts the user, then runs function/action based on input
+
+    Will quit all nested "handled prompts" with "q" using quitable input
+    """
+    text_in = quitable_input(prompt)
+    if text_in is None:
+        return "Return to main menu"
+    else:
+        if text_in in input_tup:
+            input_idx = input_tup.index(text_in)
+            next_action = action_tup[input_idx]
+        else:
+            next_action = default_action
+        if (pass_args is None) and not(pass_text):
+            result = next_action()
+        else:
+            arg_list = []
+            if pass_text:
+                arg_list.append(text_in)
+            if pass_args:
+                arg_list+=list(pass_args)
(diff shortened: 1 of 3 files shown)


natasha-aleksandrova:
@GreggHeller1 thanks for pointing out that blog post - while most of his points are valid, having all functions _always_ return a value isn't a common pattern including in other languages besides Python.

Consider this common built-in function `sort()`

```
l = [1, 3, 2]
result = l.sort()
print(result)
print(l)
```
if you run this, `sort()` has no return value - this is a perfectly valid function and you will see this everywhere.

Blogger points out that it's difficult to test functions with no return value, that statement has a lot of issues on its own:
* `mock` library exists for this reason to create fake conditions that can alter outcome of your function
* returning True/False to use in a unit test does not fully test your function in most cases - just because a database transactions succeeded does not mean the function inserted correct values.
* it is more common to use exception handling then returning simple True/False - for example if you're making an API call and return True/False, this won't tell you why the API call failed. You may retry on http 500 but give up on http 400 error code.

natasha-aleksandrova:
For your point # 2 - I agree the code is hard to read and seems over-complicated.


In lesson 4 you will learn about switch dict pattern which is much cleaner but for lesson 3 you really need just a simple case of if/else statements.

Something like this:
https://repl.it/@naaleksandrova/Lesson03

this is not fancy and uses simple language features but it is very readable and easy to follow.

students/Gregg/lesson3/mailroom.py
```diff
@@ -151,9 +152,15 @@ def create_donation_report():
...
     row_list.append('_' * 68)
-    for donor in donations:
+    donor_rows_list = []
+    hisorical_donation_amount_list = []
+    for donor in donors:
         donor_stats = get_row(donor)
-        row_list.append(format_row(donor_stats))
+        hisorical_donation_amount_list.append(donor_stats[1])
+        donor_rows_list.append(format_row(donor_stats))
+    hisorical_donation_amount_list, donor_rows_list = zip(*sorted(zip(hisorical_donation_amount_list, donor_rows_list), reverse=True))
```
natasha-aleksandrova: this would be much simpler as a list of lists to sort and then use for report generation

students/Gregg/lesson3/mailroom.py
```diff
@@ -151,9 +152,15 @@ def create_donation_report():
...
+    #donor_rows_list.sort(hisorical_donation_amount_list)
+    row_list += donor_rows_list
     report_string = ('\n').join(row_list)
     print(report_string)
     return report_string
```
natasha-aleksandrova: have a function to generate report data and separate function to print the report:
```
def get_report_data():
    pass

def create_report():
    print(get_report_data())
```

students/Gregg/lesson3/mailroom.py
```diff
@@ -179,27 +187,62 @@ def format_row(row_tupl):
     return "{:<27}${:>12.2f} {:>12}  ${:>12.2f}".format(*row_tupl)
 
 
-def list_donations(donor):
+def list_donations(donor_idx):
```
natasha-aleksandrova: this function doesn't seem to list donations, more like `get_donations_by_index`

students/Gregg/lesson3/mailroom.py
```diff
@@ -1,148 +1,149 @@
...
+    "bob"
+]
+
+donations = [
+    [3, 5, 7],
+    [4.50, 8, 2],
+    [7.77],
+    [3.6, 4.5],
+    [40]
+]
```
natasha-aleksandrova: the above data structure over-complicates your code quite a bit

this is the common data structure:
```
donor_db = [("William Gates, III", [653772.32, 12.17]),
            ("Jeff Bezos", [877.33]),
            ("Paul Allen", [663.23, 43.87, 1.32]),
            ("Mark Zuckerberg", [1663.23, 4300.87, 10432.0]),
            ]
```
as mentioned in here:
https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom_tutorial.html#choosing-a-data-structure


this is how simple your report generation can look:
```
def print_donor_report():
    """
    Generate the report of the donors and amounts donated.
    """
    # First, reduce the raw data into a summary list view
    report_rows = []
    for name, gifts in donor_db:
        total_gifts = sum(gifts)
        num_gifts = len(gifts)
        avg_gift = total_gifts / num_gifts
        report_rows.append((name, total_gifts, num_gifts, avg_gift))

    # sort the report data
    report_rows.sort(key=sort_key)

    # print it out in with a nice format.
    print("{:25s} | {:11s} | {:9s} | {:12s}".format(
          "Donor Name", "Total Given", "Num Gifts", "Average Gift"))
    print("-" * 66)
    for row in report_rows:
        print("{:25s}   {:11.2f}   {:9d}   {:12.2f}".format(*row))

```

PR #1521, opened 2019-04-09
added assignment 9

vmedina1014:


students/V_Medina/assignment_9/mailroom.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def add_donation(self, val):
+        self._donations.append(val)
+
+    @property
+    def avg_donations(self):
+        return self.donation_total / self.num_of_donations
+
+
+class Donor_Records:
```
natasha-aleksandrova: class names use CamelCase style -> `class DonorRecords`

students/V_Medina/assignment_9/mailroom.py
```diff
@@ -0,0 +1,90 @@
...
+    @property
+    def num_of_donations(self):
+        return len(self._donations)
+
+    @property
+    def donation_total(self):
+        donation_total = 0
+        for donation in self._donations:
+            donation_total += donation
+        return donation_total
```
natasha-aleksandrova: should be simply `return sum(self._donations)`

PR #1524, opened 2019-04-10
Fixed docstrings. Fixed recursive sum_seires function by adding in n0…

ramr24:
… and n1 as parameters for the recursive function to call.

### DIFF
--- a/students/Ramkumar_Rajanbabu/Lesson03/Slicing_Lab.py
+++ b/students/Ramkumar_Rajanbabu/Lesson03/Slicing_Lab.py
@@ -1,0 +1,2 @@
+#Slicing Lab
+
--- a/students/Ramkumar_Rajanbabu/Lesson03/Slicing_Lab.py
+++ b/students/Ramkumar_Rajanbabu/Lesson03/Slicing_Lab.py
@@ -1,2 +1,42 @@
 #Slicing Lab
 
+seq = "this is a string"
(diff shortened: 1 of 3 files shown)


students/Ramkumar_Rajanbabu/Lesson03/Slicing_Lab.py
```diff
@@ -0,0 +1,42 @@
...
+
+seq = "this is a string"
+tup = (2, 54, 13, 12, 5, 32)
+
+def exchange_first_last(seq):
+    first = seq[-1] 
+    mid = seq[1:-1]
+    last = seq[0]
+    new_seq = first + mid + last 
+    return new_seq
```
natasha-aleksandrova: this needs to use slicing only - this doesn't work for lists/tuple

ramr24: I'm still working on the slicing lab. Also, I'm still fixing the exchange_first_last since as you mentioned, this doesn't work for lists/tuples. Am I doing the pull request incorrectly? How would I do a pull request after completing an assignment while also starting a new exercise, for example the slicing lab?

natasha-aleksandrova:
PR is a diff between two branches - so any new commits will show under the same PR until it is merged.

In the industry people use branches but it's more of an advanced topic.

If you do end up making another commit and you have an open PR you can always leave a comment to let us know that its not ready. Thanks!

PR #1529, opened 2019-04-13
Assignment09

mzkhan87:
Hi Natasha, 

I really want to thank you for your feedback. As I implement your feedback, my code looks much prettier and it reduces in size. I also learn a lot from your feedback and I improve my coding skills. 

I've made the recommended changes per your feedback. Please review my work and let me know if there is still something you'd like me to improve on. 

My class is ending on April 19 and I have one more assignment to go. I am hoping to finish it all before my course ends. Your promptly feedback are appreciated.


natasha-aleksandrova:
@mzkhan87 I'm really glad to hear that you're able to apply my feedback and improve the code! 

You should be good to go on this lesson, go ahead and proceed to the next one 🙂

PR #1530, opened 2019-04-13
Finished slicing lab

ramr24:
Still working on the list lab and string formatting assignment. Finished slicing lab.


natasha-aleksandrova:
@ramr24 please fix `exchange_first_last` function to only use slicing (there should be no type checking also)

PR #1531, opened 2019-04-14
Fixed exchange_first_last, only used slicing for the function.

ramr24:
Only fixed slicing lab, still working on other 2 exercises.

### DIFF
--- a/students/Ramkumar_Rajanbabu/Lesson03/Slicing_Lab.py
+++ b/students/Ramkumar_Rajanbabu/Lesson03/Slicing_Lab.py
@@ -14,19 +14,10 @@ def exchange_first_last(seq):
         
     """
     
-    if seq == tuple(seq):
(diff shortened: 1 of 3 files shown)


students/Ramkumar_Rajanbabu/Lesson03/Slicing_Lab.py
```diff
@@ -14,19 +14,10 @@ def exchange_first_last(seq):
...
-        last = [seq[-1]]
-    else:
-        first = seq[0] 
-        mid = seq[1:-1]
-        last = seq[-1]
-    new_seq = last + mid + first 
+    first = seq[:1]
+    mid = seq[1:-1]
+    last = seq[-1:]
+    new_seq = last + mid + first
```
natasha-aleksandrova: much better!

PR #1533, opened 2019-04-15
Assignment10

mzkhan87:
Dear Natasha, 

I implemented map, filter and lambda functions in mailroom_fp project. I've added a method in the DonorCollection class called "projected_amount" instead of method name "challenge" in the instructions. I felt projected_amount was a better naming convention. I can correct that if you'd like. 

Please review my assignment and provide your feedback if I should improve the fp part of the assignment. 
Thanks so much for your help throughout the course. I look forward to taking the second python class towards the certificate. 

Muhammad

### DIFF
--- a/students/MzKhan/lesson10/mailroom_fp/cli_main_fp.py
+++ b/students/MzKhan/lesson10/mailroom_fp/cli_main_fp.py
@@ -1,0 +1,155 @@
+'''
+Name: Muhammad Khan
+Date: 04/09/2019
+Assignment09
+'''
+import sys
+from mailroom_fp import Donor
+from mailroom_fp import DonorCollection
+
+def prompt_user():
+    """Return user's selected option as an integer"""
+    print()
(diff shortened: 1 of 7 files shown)


students/MzKhan/lesson10/mailroom_fp/cli_main_fp.py
```diff
@@ -0,0 +1,155 @@
...
+        print("1: Match to total amount donated")
+        print("2: Custom donation")
+        print("3: Leave the menu ")
+        try:
+            user_response = int(input("Option: "))
+            if user_response not in [1,2,3]:
+                print("***INVALID option selected: Try Again!***")
+                continue
+            if user_response == 3:break
+            name = name_prompt(input_msg) ; name = name.title()
```
natasha-aleksandrova: it's a python style violation to use single line multi-statements (both lines 117 and 118)

PR #1534, opened 2019-04-19
Lesson 9, OOP Mailroom

ganttArt:


students/christopher_gantt/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Attribute:
+        *args (class): Donors in class Donor
+    """
+
+    def __init__(self, *args):
+        self.donors = [*args]
+
+    def add_donor(self, new_donor):
+        return self.donors.append(new_donor)
```
natasha-aleksandrova: shouldn't have return here

PR #1538, opened 2019-04-29
Lesson 3 - Used slicing instead of indexing, added missing slicing fu…

Charith-Acharya:
…nction

Lesson 6  - Added else statement, added missing sort function
Lesson 8 - Made changes so user cannot set area
Lesson 9 -Madee corrections  removed if TypeError statement
Added Lesson 7


natasha-aleksandrova:
@Charith-Acharya did you submit your work on the course website? Please do so right away - otherwise you will get zero completion for the course.

PR #1539, opened 2019-04-30
Finished Mailroom Part 1

ramr24:


students/Ramkumar_Rajanbabu/Lesson03/mailroom_p1.py
```diff
@@ -0,0 +1,163 @@
...
+
+        avg = total / gifts
+        avg_gift.append(avg)
+    
+    print()
+    print(title_string)
+    print(title_bar)
+    for index in range(len(donor_name)):
+        f_string = "{:<25s}  ${:>11.2f}  {:>10d}  ${:>12.2f}".format(donor_name[index], total_given[index], num_gifts[index], avg_gift[index])
+        print(f_string)
```
natasha-aleksandrova: this function is missing the sorted feature (please check requirements).


I would also recommend creating a list of your summary like:
```
for donor in donors_db:
    summary.append((name, sum(donations), len(donations), avg)
```
then loop through your summary to create report - will simplify the code and it will make more sense when using with `sorted` function.

PR #1544, opened 2019-05-11
Resubmit mailroom part 1

smckellips:


students/smckellips/lesson03/mailroom.py
```diff
@@ -1,27 +1,28 @@
...
 prompt = "\n".join(("Welcome to the mailroom!",
           "Please choose from below options:",
           "1 - Send a thank you",
           "2 - Create a report",
           "3 - Exit",
           ">>> "))
 
-def thank_you(donor_db):
+def thank_you():
+    global donor_db
```
natasha-aleksandrova: note that you don't need `global` here

PR #1545, opened 2019-05-12
Finished mailroom part 2

ramr24:


students/Ramkumar_Rajanbabu/Lesson04/mailroom_p2.py
```diff
@@ -0,0 +1,141 @@
...
+        summary: list with donor summary
+    Returns:
+        summary[1]: sorts by total_given values
+
+    """
+    return summary[1]
+    
+if __name__ == "__main__":
+    #Guard against running code automatically if this module is imported
+    main()
```
charlesrobison: Nice job on this one!

students/Ramkumar_Rajanbabu/Lesson05/mailroom_p3.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    
+    if name not in donors_db:
+        donors_db[name] = [] #donors_db[key] = [value]
+        
+    #Exception handling
+    try:
+        donation_amount = float(input("Type in the donation amount: $"))
+    except ValueError: #ValueError if not integer/float
+        print("Please enter an integer.")
```
charlesrobison: So we've got a bug here - though it prints out the error, if a user enters a string, what does the program do with that string?  If I then run the report again, I get kicked out.

PR #1551, opened 2019-05-28
Assignment 4 Mailroom part 3 

smckellips:


students/smckellips/lesson05/mailroom3.py
```diff
@@ -0,0 +1,108 @@
...
+        "paul allen": ("Paul Allen", [663.23, 43.87, 1.32]),
+        "mark zuckerberg": ("Mark Zuckerberg", [1663.23, 4300.87, 10432.0])
+    }
+
+
+def thank_you():
+    try:
+        donor = input("Name of the donor? ( 'q' to quit ) ")
+    except Exception as e:
+        print(f"There is an error with your input, {e}.")
```
natasha-aleksandrova: what exception are you anticipating here?

students/smckellips/lesson05/mailroom3.py
```diff
@@ -0,0 +1,108 @@
...
+    if donor_l == 'q':
+        return
+    elif donor_l == "list":
+        print(list_donors())
+    else:
+        try:
+            donation_amt = float(input("Amount of the donation? "))
+        except ValueError:
+            print("Not a float")
+        except Exception as e:
```
natasha-aleksandrova: what exception are you anticipating here?

students/smckellips/lesson05/mailroom3.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def send_letters():
+    for donor in  donor_db.values():
+        letter = write_letter(donor)
+        filename = donor[0].replace(' ','_') + '.txt'
+        try:
+            with open(filename, 'w') as f:
+                f.write(letter)
+        except:
```
natasha-aleksandrova: you should catch a specific exception.

students/smckellips/lesson05/mailroom3.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def menu_selection(prompt, dispatch_dict):
+    while True:
+        try:
+            response = input(prompt)
+        except KeyboardInterrupt:
+            break
+        if response in dispatch_dict.keys():
+            if dispatch_dict[response]() == "exit menu":
```
natasha-aleksandrova: this should be addressed with a KeyError catch.

smckellips:
In general, I was starting with catching the base exception, then going
more specific from there.  I suppose I should spend more time reviewing
likely exceptions to trap.

On the KeyError, is it possible to have a KeyError since I've checked to
make sure the response is a valid key before calling
dispatch_dict[response]()?

natasha-aleksandrova:
For your KeyError question - for this assignment you should be updating your code to handle error cases like bad user input. Even though you're doing a check to see if the key exists to prevent bad user input, for this assignment the expectation is to use exception handling for bad user input. Hope this clarifies.

PR #1552, opened 2019-05-28
Assignment 5: Mailroom Part 4

smckellips:


students/smckellips/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,64 @@
...
+
+def test_list_donors():
+    donor_list = mailroom.list_donors()
+    assert 'Jeff Bezos' in donor_list
+
+
+def test_write_report():
+    report = mailroom.write_report()
+    assert report.startswith("Donor Name                | Total Given | Num Gifts | Average Gift")
+    assert report.__contains__("Paul Allen")
```
natasha-aleksandrova: a better test would be to test the full output

PR #1556, opened 2019-06-25
Finished html render

ramr24:


students/Ramkumar_Rajanbabu/lesson07/html_render.py
```diff
@@ -9,19 +9,132 @@
...
     def append(self, new_content):
+        """Appends new content to content list."""
         self.contents.append(new_content)
-
-    def render(self, out_file):
-        #Loop through the list of contents:
+    
+    #Step 4
+    def _open_tag(self):
+        return "".join([f"<{self.tag}", "".join(f' {key}="{value}"' for key, value in self.attributes.items()), ">"])
```
natasha-aleksandrova: this would be more clear using simple string concat like `f"<{self.tag}" + "".join()`

`join()` is used for multiple values or variable number of values

same comment applies to similar patterns below

PR #1557, opened 2019-07-16
Finished mailroom oop

ramr24:
I will continue to work on this since this mailroom oop project as I know this is not a fully finished code without separating code into cli.py and not having a test.py. I will definitely finish and organize my code to a better stage before starting my next course.

### DIFF
--- a/students/Ramkumar_Rajanbabu/lesson09/donor_models.py
+++ b/students/Ramkumar_Rajanbabu/lesson09/donor_models.py
@@ -1,0 +1,160 @@
+#Main Script
+
+import sys
+
+class Donor():
+    """Single donor information"""
+    
+    def __init__(self, name, donations = None):
+        self._name = name
+        if donations is not None:
+            self._donations = donations
+        else: 
+            self._donations = []
+    
+    def __repr__(self):
+        return f"Donor: {self.name}, Donations: {self.donations}"
+    
+    def add_donations(self, amount):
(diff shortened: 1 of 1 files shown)


students/Ramkumar_Rajanbabu/lesson09/donor_models.py
```diff
@@ -0,0 +1,160 @@
...
+                print(self.list_donor())
+                name = None
+        
+        donation_amount = None
+        while not donation_amount:
+            #Exception handling
+            try:
+                donation_amount = input_donation()
+            except ValueError: #ValueError if not integer/float
+                print("Please enter an integer.")
```
natasha-aleksandrova: there should be no input functions in your data classes (even if input() is not directly called)

students/Ramkumar_Rajanbabu/lesson09/donor_models.py
```diff
@@ -0,0 +1,160 @@
...
+            if user_response == 1:
+                donors_db.send_thanks()
+            elif user_response == 2:
+                donors_db.display_report()
+            elif user_response == 3:
+                donors_db.send_letters_all()
+            elif user_response == 4:
+                exit_program()
+            elif user_response not in range(1,4):
+                print("Please enter a valid number within 1-4!\n")
```
natasha-aleksandrova: switch dict would be more appropriate here

